PR #21, opened 2025-06-28
fix: TableScanNode's ctor and Connector::createDataSource/Sink APIs

mbasmanova:
Summary:
X-link: https://github.com/facebookincubator/velox/pull/13920

TableScanNode, Connector::createDataSource, createIndexSource and createDataSink APIs used to take table and column handles as non-const forcing calling code to use std::const_pointer_cast, which is an anti-pattern.

Differential Revision: D77490731


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #41, opened 2025-07-17
fix: add TableWrite support for Velox optimizer

hdikeman:
Summary:
In this change adding support for processing plans which contain TableWrite nodes in Verax

This is actually a very simple case for two reasons:

1. there can be only one TableWrite node per plan
2. TableWrites are always leaf nodes, which means it's very easy to extract from the plan immediately and hold until it's time to construct the output Velox plan

The details of the table write are stored in an intermediate representation TableWriteIR which just holds some fields for input type, input column names, the table handle, and a couple others. At the time we move from the relational plan back to a Velox plan, this intermediate representation is validated against the optimized plan and recreated

Things would get more complicated if we permitted this table write to be bucketed. It would involve some additional exchanges and holding additional information to construct the output plan. But (1) this is complicated and (2) we don't have an immediate use for it, since the first Axiom use-case is the single node Impulse deployment (non-distributed)

This will need revisiting once we adopt logical plans

Differential Revision: D78389886

### DIFF
--- a/frontend/optimizer/Plan.cpp
+++ b/frontend/optimizer/Plan.cpp
@@ -92,7 +92,7 @@ Optimization::Optimization(
   for (auto& join : root_->joins) {
     join->guessFanout();
   }
-  setDerivedTableOutput(root_, inputPlan_);
+  setDerivedTableOutput(root_, *inputPlanWithoutWrite_);
 }
 
 void Optimization::trace(
--- a/frontend/optimizer/Plan.h
+++ b/frontend/optimizer/Plan.h
@@ -213,6 +213,25 @@ struct PlanSet {
   PlanPtr addPlan(RelationOpPtr plan, PlanState& state);
 };
 
+// The intermediate representation of a table write, containing all the
+// attributes of the write required to reconstruct it on the output.
+struct TableWriteIR {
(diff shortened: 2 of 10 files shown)


mbasmanova:
@MBkkt 

Orri is working on adding support for table write: #181 #119

Curious, what's your use case for write?

MBkkt:
@mbasmanova 

> Orri is working on adding support for table write: ...

I saw, they're looking right now incomplete and even worse to me: I don't understand these interfaces, e.g. why something like createTable presented?
I don't understand how to use it and what overall workflow looks like.
It wasn't like this with velox interfaces or current PR, so I'm a little confused with these Orri PRs

> Curious, what's your use case for write?

We're building sql OLAP database (with search engine), and want to migrate table writes as well as table reads to verax + velox. 

So we planned to implement our own DataSink, DataSource (all related metadata handles will live in our catalog abstraction) and use velox TableWrite/TabeScan nodes to write/read from our storages.

About inserts to the table exists one issue and it's not trivial to resolve.
In Postgres, inserts/update allow to return updated row, this isn't supported by TableWrite/DataSink. We thought about this, and decided it's ok for first version, but later we will need to add this.
Maybe with our custom node or with extensions to current interfaces

PR #43, opened 2025-07-18
add TableNameParser to SchemaResolver

hdikeman:
Summary:
this is a generic implementation of table name parsing for lookups, where table names may have be of the following format:
```
table
schema.table
catalog.schema.table
```

Where defaults are substituted if a table is underspecified. This is actually ripped from the Native Frontend, where it was used for Hive name parsing. It will be removed from that location as part of a follow-up change

I'm also splitting this into its own smaller build target for BUCK so that others can import and use it without pulling in all of Verax as a dependency

Differential Revision: D78534567


yuandagits (approved):

yuandagits:
Don't forget to update the title here to reflect the internal PR title

PR #55, opened 2025-07-23
Remove errorString arg from QueryTestBase::planVelox/Sql methods

mbasmanova:
Differential Revision: D78794731

### DIFF
--- a/axiom/optimizer/LogicalPlanToGraph.cpp
+++ b/axiom/optimizer/LogicalPlanToGraph.cpp
@@ -53,12 +53,25 @@ DerivedTableP Optimization::makeQueryGraphFromLogical() {
   return root_;
 }
 
+namespace {
+
 const std::string* columnName(const lp::Expr& expr) {
   if (expr.isInputReference()) {
     return &expr.asUnchecked<lp::InputReferenceExpr>()->name();
   }
   return nullptr;
 }
+
+template <TypeKind kind>
+std::shared_ptr<const variant> toVariant(BaseVector& constantVector) {
+  using T = typename TypeTraits<kind>::NativeType;
+  if (auto typed = dynamic_cast<ConstantVector<T>*>(&constantVector)) {
+    return std::make_shared<Variant>(typed->valueAt(0));
+  }
+  VELOX_FAIL("Literal not of foldable type");
+}
+
+} // namespace
 
 void Optimization::translateConjuncts(
     const lp::ExprPtr& input,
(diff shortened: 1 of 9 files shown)


axiom/optimizer/LogicalPlanToGraph.cpp
```diff
@@ -53,13 +53,26 @@ DerivedTableP Optimization::makeQueryGraphFromLogical() {
...
+
 const std::string* columnName(const lp::Expr& expr) {
   if (expr.isInputReference()) {
     return &expr.asUnchecked<lp::InputReferenceExpr>()->name();
   }
   return nullptr;
 }
 
+template <TypeKind kind>
+std::shared_ptr<const variant> toVariant(BaseVector& constantVector) {
```
pedroerp, line 66: why not using `vectorToVariant(constantVector, 0)` instead?

axiom/optimizer/ToGraph.cpp
```diff
@@ -16,20 +16,48 @@
...
+
+bool isCall(const core::TypedExprPtr& expr, const std::string& name) {
+  if (auto call = std::dynamic_pointer_cast<const core::CallTypedExpr>(expr)) {
+    return exec::sanitizeName(call->name()) == name;
+  }
+  return false;
+}
+
+template <TypeKind kind>
+const variant* toVariant(BaseVector& constantVector) {
```
pedroerp, line 51: ditto

pedroerp: or maybe having a `ConstantVectorToVariant` utility somewhere

axiom/optimizer/ToVelox.cpp
```diff
@@ -782,31 +955,10 @@ core::PlanNodePtr Optimization::makeFragment(
...
-      }
-      std::vector<std::string> names;
-      std::vector<core::TypedExprPtr> exprs;
-      for (auto i = 0; i < project->exprs().size(); ++i) {
-        names.push_back(project->columns()[i]->toString());
-        exprs.push_back(toTypedExpr(project->exprs()[i]));
-      }
-      return std::make_shared<core::ProjectNode>(
-          nextId(*project), std::move(names), std::move(exprs), input);
+      return makeProject(*op->as<Project>(), fragment, stages);
```
pedroerp, line 958: looks much cleaner :)

pedroerp (approved):
Thank you for the refactor.

axiom/optimizer/LogicalPlanToGraph.cpp
mbasmanova: No particular reason. This is how it was, so I haven't changed it. We can change in a follow-up and, perhaps, also take care of https://github.com/facebookincubator/velox/issues/14057 and add BaseVector::variantAt(index).

PR #65, opened 2025-07-23
Fix zero row support in ValuesNode

pedroerp:
Summary:
When there are zero rows in the ValuesNode, the variant vector will be
empty. That is a legit case.

Differential Revision: D78850942

### DIFF
--- a/axiom/logical_plan/LogicalPlanNode.cpp
+++ b/axiom/logical_plan/LogicalPlanNode.cpp
@@ -50,18 +50,14 @@ ValuesNode::ValuesNode(
     std::vector<Variant> rows)
     : LogicalPlanNode(NodeKind::kValues, id, {}, rowType),
       rows_{std::move(rows)} {
-  if (rows_.empty()) {
-    VELOX_USER_CHECK_EQ(0, rowType->size());
-  }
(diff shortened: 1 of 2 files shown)


axiom/logical_plan/LogicalPlanNode.cpp
```diff
@@ -50,18 +50,15 @@ ValuesNode::ValuesNode(
...
-    VELOX_USER_CHECK_EQ(0, rowType->size());
-  }
-
   UniqueNameChecker::check(rowType->names());
 
   for (const auto& row : rows_) {
+    const auto variantType = row.inferType();
     VELOX_USER_CHECK(
-        rowType->equivalent(*row.inferType()),
+        rowType->equivalent(*variantType),
```
mbasmanova, line 58: We may want to use Variant::isTypeCompatible to allow for empty arrays and maps.

```
  /// Returns true if the type of this Variant is compatible with the specified
  /// type. Similar to inferType()->kindEquals(type), but treats
  /// TypeKind::UNKNOWN equal to any other TypeKind.
  bool isTypeCompatible(const TypePtr& type) const;
```

PR #71, opened 2025-07-24
Extract ParallelProject node and operator into a separate build target

mbasmanova:
Differential Revision: D78900278


xiaoxmeng (approved):
@mbasmanova  thanks!

PR #76, opened 2025-07-25
Add CI for verax

feilong-liu:


.github/workflows/test.yml
```diff
@@ -38,72 +38,69 @@ concurrency:
...
+      # Install dependencies
+      - name: Install Dependencies
+        working-directory: velox
         run: |
-          files=$(git diff --name-only HEAD^1 HEAD)
-          echo "::group::Changed files"
-          echo $files | tr ' ' '\n'
-          echo "::endgroup::"
-          pre-commit run --show-diff-on-failure --color=always --files $files
+          source scripts/setup-ubuntu.sh && install_apt_deps && install_faiss_deps
```
kgpai, line 71: Very curious why you are only installing apt_deps and faiss ? Would be safer to just do install_velox_deps ?

.github/workflows/test.yml
```diff
@@ -38,72 +38,69 @@ concurrency:
...
+      - name: Clear CCache Statistics
+        run: |
+          ccache -sz
+
+      # Build Verax directly (which will also build Velox)
+      - name: Build Verax Debug
+        id: build-verax
+        continue-on-error: true  # Allow this step to fail without stopping the workflow
+        env:
+          MAKEFLAGS: NUM_THREADS=8 MAX_HIGH_MEM_JOBS=4 MAX_LINK_JOBS=3
```
kgpai, line 82: Can you set CPU_TARGET: "avx" here ? Setting it to "default" is wrong and thats why its reverting to sse builds.

.github/workflows/test.yml
feilong-liu: Oh, I was copying from linux-build-base.yml in velox https://github.com/facebookincubator/velox/blob/13de155c716c40b10d679437a6e88d7d2ba26997/.github/workflows/linux-build-base.yml#L165

.github/workflows/test.yml
feilong-liu: I tried set to "avx" but it still failed.

Also in the Makefile it has  `CPU_TARGET ?= "avx"` which I assume will use avx when no cpu target set here?
https://github.com/facebookexperimental/verax/blob/3014ca3fc65a03c65526fe71f1f66c4340dee877/Makefile#L40

kgpai: Looks like this flag is not being propagated and its still trying to do an sse build. As a test can you force the value in the cmakefile : https://github.com/facebookexperimental/verax/blob/main/CMakeLists.txt#L93 (and also print out a log statement to make sure) and then try this CI.

.github/workflows/test.yml
feilong-liu: > Even better will be to reuse one of the prebuilt velox images so these dependencies are pre-installed

Do you have any suggestions on which image to use? I am running on 8-core-ubuntu-22.04.

feilong-liu: @kgpai The build failed due to out of space with "Error: Process completed with exit code 137." after moving to container ghcr.io/facebookincubator/velox-dev:centos9. Did you see this problem before? Is there any solution for it ? Thanks!

xiaoxmeng (approved):

PR #80, opened 2025-07-26
Fix type check in SetNode

mbasmanova:
Summary:
SetNode used to check that all output types of all inputs are the same. However, this is not possible because we require all symbol names in the plan to be unique.

Relax the check to allow different column names, but same types.

Differential Revision: D79019019

### DIFF
--- a/axiom/logical_plan/LogicalPlanNode.cpp
+++ b/axiom/logical_plan/LogicalPlanNode.cpp
@@ -196,6 +196,38 @@ folly::F14FastMap<SetOperation, std::string> setOperationNames() {
 
 VELOX_DEFINE_ENUM_NAME(SetOperation, setOperationNames)
 
+SetNode::SetNode(
+    const std::string& id,
+    const std::vector<LogicalPlanNodePtr>& inputs,
+    SetOperation operation)
+    : LogicalPlanNode(NodeKind::kSet, id, inputs, inputs.at(0)->outputType()),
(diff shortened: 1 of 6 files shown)


axiom/logical_plan/LogicalPlanNode.cpp
```diff
@@ -196,6 +196,38 @@ folly::F14FastMap<SetOperation, std::string> setOperationNames() {
...
+  for (auto i = 1; i < inputs.size(); ++i) {
+    const auto& rowType = inputs.at(i)->outputType();
+
+    // The names are different, but types must be the same.
+    VELOX_USER_CHECK(
+        firstRowType->equivalent(*rowType),
+        "Output schemas of all inputs to a Set operation must match");
+
+    // Individual column types must match exactly.
+    for (auto j = 0; j < firstRowType->size(); ++j) {
```
xiaoxmeng, line 219: The row type check above is not sufficient and cover the child type check? Thanks!

xiaoxmeng (approved):
@mbasmanova LGTM % a nit question. Thanks!

axiom/logical_plan/LogicalPlanNode.cpp
mbasmanova: That's right. firstRowType->equivalent(*rowType) doesn't ensure that field names in nested structs match.

PR #81, opened 2025-07-26
Add subquery support to logical plan

feilong-liu:
The user can call `subqueryBuilder()` function to get a PlanBuilder which can be used to build subqueries.

The PR provides three util functions, `withInSubquery`, `withExistsSubquery`, `withScalarSubquery` correspondingly for In subquery, Exists subquery and Scalar subquery. This is because for exists subquery and in subquery, the usage of subquery is tied with its usage environment. For example, for scalar subqueries, we can first map the subquery to a variable, e.g. `subquery_value`, then use it in SQL expression, like `value > subquery_value`. But for exists query, if we map the subquery to a variable, e.g. `subquery_value`, the expression `exists(subquery_value)` will not compile as it's not a valid syntax. So we need to directly construct the EXISTS special expression within the util functions.

Now let me show an example of how to construct simple subqueries.
In the following example queries, the table schema is:
L(key:integer, v:integer)
R(key:integer, w:integer)
* SQL query: SELECT * FROM L WHERE key IN (SELECT key FROM R)
PlanBuilder code snippet:
```
  PlanBuilder planBuilder;
  auto plan = planBuilder.values(leftType, leftData)
                  .as("l")
                  .withInSubquery(
                      "in_subquery_value",
                      planBuilder.subqueryBuilder()
                          .values(rightType, rightData)
                          .as("r")
                          .project({"key"})
                          .build(),
                      "key")
                  .filter("in_subquery_value")
                  .build();
```
Plan printer output:
```
- Filter: in_subquery_value -> ROW<key:INTEGER,v:INTEGER,in_subquery_value:BOOLEAN>
  - Project: -> ROW<key:INTEGER,v:INTEGER,in_subquery_value:BOOLEAN>
      key := key
      v := v
      in_subquery_value := IN(key, 
Subquery
(- Project: -> ROW<key:INTEGER>
    key := key_0
  - Project: -> ROW<key_0:INTEGER>
      key_0 := key_0
    - Values: 2 rows -> ROW<key_0:INTEGER,w:INTEGER>
)
)
    - Values: 4 rows -> ROW<key:INTEGER,v:INTEGER>
```
* SQL query: SELECT * FROM L WHERE EXISTS (SELECT key FORM r WHERE key=L.key)
PlanBuilder code snippet:
```
  PlanBuilder planBuilder;
  auto plan = planBuilder.values(leftType, leftData)
                  .as("l")
                  .withExistsSubquery(
                      "exist_subquery_value",
                      planBuilder.subqueryBuilder()
                          .values(rightType, rightData)
                          .as("r")
                          .filter("key = l.key")
                          .project({"key"})
                          .build())
                  .filter("exist_subquery_value")
                  .build();
```
Plan printer output:
```
- Filter: exist_subquery_value -> ROW<key:INTEGER,v:INTEGER,exist_subquery_value:BOOLEAN>
  - Project: -> ROW<key:INTEGER,v:INTEGER,exist_subquery_value:BOOLEAN>
      key := key
      v := v
      exist_subquery_value := EXISTS(
Subquery
(- Project: -> ROW<key:INTEGER>
    key := key_0
  - Project: -> ROW<key_0:INTEGER>
      key_0 := key_0
    - Filter: eq(key_0, key) -> ROW<key_0:INTEGER,w:INTEGER>
      - Values: 2 rows -> ROW<key_0:INTEGER,w:INTEGER>
)
)
    - Values: 4 rows -> ROW<key:INTEGER,v:INTEGER>
```
* SQL QUERY: SELECT * FROM L WHERE v > (SELECT MIN(w) FROM R)
PlanBuilder code snippet:
```
  auto plan = planBuilder.values(leftType, leftData)
                  .as("l")
                  .withScalarSubquery(
                      "scalar_subquery_value",
                      planBuilder.subqueryBuilder()
                          .values(rightType, rightData)
                          .as("r")
                          .aggregate({}, {"min(w) as min_w"})
                          .build())
                  .filter("v > scalar_subquery_value")
                  .build();
```
Plan printer output:
```
- Filter: gt(v, scalar_subquery_value) -> ROW<key:INTEGER,v:INTEGER,scalar_subquery_value:INTEGER>
  - Project: -> ROW<key:INTEGER,v:INTEGER,scalar_subquery_value:INTEGER>
      key := key
      v := v
      scalar_subquery_value := 
Subquery
(- Aggregate( -> ROW<min_w:INTEGER>
    min_w := min(w)
  - Values: 2 rows -> ROW<key_0:INTEGER,w:INTEGER>
)

    - Values: 4 rows -> ROW<key:INTEGER,v:INTEGER>
```

--------------
Summary:
Add subquery support to logical plan node.
For subqueries:
* subqueries can reference columns in outer query
* columns in subquery can have the same name as columns in outer query
* Resolution of columns will try to match the current query first, and try outer queries from inner to outer until it finds a match.

To achieve this, we make the following changes to the PlanBuilder.
* Instead of having one single `NameMapping outputMapping_`, now we have a list of `NampeMapping outputMapping_`. Here the NameMapping in the front represents the outer most query, while the NameMapping in the back represents the current query the PlanBuilder is working on.
  * When we are building a query, the plan builder tries to resolve the input by iterating the list of NampeMapping from back to front, and return the first match it finds
* Another change is to make the NampeMapping not only remember the unique ID assigned to a input expression , but also the type of the input expression. This is because currently we rely on the current node to resolve the type of an input, but with subquery, the current node may not be the node which holds this input expression. Directly remember both name and type make it easy to get the type.
* Since we will use the same name allocator for the whole query plan build, all expressions have unique name so we do not need to attach scope to the cross referenced input in the output plan.

Differential Revision: D78953518

### DIFF
--- a/axiom/logical_plan/Expr.h
+++ b/axiom/logical_plan/Expr.h
@@ -338,7 +338,9 @@ enum class SpecialForm {
 
   kStar = 9,
 
-  // TODO Add IN and EXISTS.
+  kIn = 10,
+
+  kExists = 11,
 };
 
 VELOX_DECLARE_ENUM_NAME(SpecialForm)
@@ -622,14 +624,25 @@ using LambdaExprPtr = std::shared_ptr<const LambdaExpr>;
 class LogicalPlanNode;
 using LogicalPlanNodePtr = std::shared_ptr<const LogicalPlanNode>;
 
-/// Scalar subquery that returns exactly one row and one column. Can be used
-/// anywhere a scalar function call can be used.
+/// Enum representing the type of subquery.
+enum class SubqueryType {
+  SCALAR, // Scalar subquery that returns exactly one row and one column
+  IN, // IN subquery used in expressions like "x IN (subquery)"
+  EXISTS // EXISTS subquery used to check if subquery returns any rows
+};
+
+/// Subquery expression that can be used in various contexts depending on its
+/// type.
 class SubqueryExpr : public Expr {
  public:
-  explicit SubqueryExpr(const LogicalPlanNodePtr& subquery);
+  explicit SubqueryExpr(const LogicalPlanNodePtr subquery, SubqueryType type);
 
   const LogicalPlanNodePtr& subquery() const {
     return subquery_;
+  }
+
+  SubqueryType subqueryType() const {
+    return subqueryType_;
   }
 
   void accept(const ExprVisitor& visitor, ExprVisitorContext& context)
@@ -637,6 +650,7 @@ class SubqueryExpr : public Expr {
 
  private:
   const LogicalPlanNodePtr subquery_;
+  const SubqueryType subqueryType_;
 };
 
 using SubqueryExprPtr = std::shared_ptr<const SubqueryExpr>;
--- a/axiom/logical_plan/ExprPrinter.cpp
+++ b/axiom/logical_plan/ExprPrinter.cpp
@@ -16,6 +16,14 @@
 
 #include "axiom/logical_plan/ExprPrinter.h"
 
+// Forward declaration to avoid circular dependency
+namespace facebook::velox::logical_plan {
+class PlanPrinter {
+ public:
+  static std::string toText(const LogicalPlanNode& root);
+};
+} // namespace facebook::velox::logical_plan
+
 namespace facebook::velox::logical_plan {
 
 namespace {
@@ -98,10 +106,16 @@ class ToTextVisitor : public ExprVisitor {
     expr.body()->accept(*this, context);
   }
 
-  void visit(const SubqueryExpr& /* expr */, ExprVisitorContext& /* context */)
+  void visit(const SubqueryExpr& expr, ExprVisitorContext& context)
       const override {
-    // TODO Implement.
-    VELOX_NYI();
+    auto& out = toOut(context);
+    out << std::endl;
+    out << "Subquery";
+    out << std::endl;
+    out << "(";
+    out << PlanPrinter::toText(*(expr.subquery()));
+    out << ")";
+    out << std::endl;
   }
 
  private:
--- a/axiom/logical_plan/PlanBuilder.h
+++ b/axiom/logical_plan/PlanBuilder.h
@@ -17,12 +17,11 @@
 
 #include "axiom/logical_plan/LogicalPlanNode.h"
 #include "axiom/logical_plan/NameAllocator.h"
+#include "axiom/logical_plan/NameMappings.h"
 #include "velox/parse/ExpressionsParser.h"
 #include "velox/parse/PlanNodeIdGenerator.h"
 
 namespace facebook::velox::logical_plan {
-
-class NameMappings;
 
 class PlanBuilder {
  public:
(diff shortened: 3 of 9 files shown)


axiom/logical_plan/tests/PlanPrinterTest.cpp
```diff
@@ -357,5 +357,156 @@ TEST_F(PlanPrinterTest, lambda) {
           testing::Eq("")));
 }
 
+TEST_F(PlanPrinterTest, InSubquery) {
```
mbasmanova, line 360: test methods start with lowercase: InSubquery -> inSubquery

axiom/logical_plan/PlanBuilder.h
```diff
@@ -91,8 +104,17 @@ class PlanBuilder {
 
   PlanBuilder& as(const std::string& alias);
 
+  ExprPtr buildSubquery(
```
mbasmanova, line 107: This API doesn't fit the overall style of the builder. What does it do? Why do we need it?

axiom/logical_plan/Expr.h
```diff
@@ -622,21 +624,33 @@ using LambdaExprPtr = std::shared_ptr<const LambdaExpr>;
...
+  IN, // IN subquery used in expressions like "x IN (subquery)"
+  EXISTS // EXISTS subquery used to check if subquery returns any rows
+};
+
+/// Subquery expression that can be used in various contexts depending on its
+/// type.
 class SubqueryExpr : public Expr {
  public:
-  explicit SubqueryExpr(const LogicalPlanNodePtr& subquery);
+  explicit SubqueryExpr(const LogicalPlanNodePtr subquery, SubqueryType type);
```
mbasmanova, line 638: Why do we need to specify SubqueryType explicitly? Shouldn't this be inferred from the context instead?

axiom/logical_plan/ExprPrinter.cpp
```diff
@@ -16,6 +16,14 @@
 
 #include "axiom/logical_plan/ExprPrinter.h"
 
+// Forward declaration to avoid circular dependency
```
mbasmanova, line 19: This shouldn't be necessary.

axiom/logical_plan/PlanBuilder.h
feilong-liu: It's because we do not have full query parse support. When parsing expression, we are using the parser util in velox which uses the duckdb parser.

The workaround I am thinking here is to let users extract the subquery expressions out, create an `SubqueryExpr` for it with this `buildSubquery` util function and project it to an output variable which can be used in subsequent expressions.

This function is to be used with the `PlanBuilder& with(const std::unordered_map<ExprPtr, std::string>& subqueryOutput);` defined above. Like:

```
auto in_subquery_ptr = buildSubquery(...);
with({in_subquery_ptr, "in_subquery_value"});
```
and use `in_subquery_value` where the subquery is referenced.

axiom/logical_plan/PlanBuilder.h
```diff
@@ -65,6 +74,10 @@ class PlanBuilder {
   /// Similar to 'project', but appends 'projections' to the existing columns.
   PlanBuilder& with(const std::vector<std::string>& projections);
 
+  // Create a projection mapping subquery output to name corresponding to hints
```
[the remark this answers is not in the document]
feilong-liu: This function is to be used with `buildSubquery` function.
Will update the comment.

axiom/logical_plan/tests/PlanPrinterTest.cpp
```diff
@@ -131,7 +131,7 @@ TEST_F(PlanPrinterTest, aggregate) {
           .values(rowType, data)
           .aggregate(
               {"a"}, {"sum(b) as total", "avg(b) as mean", "min(b + 1::int)"})
-          .with({"total + 1", "mean * 0.3"})
+          .with(std::vector<std::string>{"total + 1", "mean * 0.3"})
```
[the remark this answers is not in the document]
feilong-liu: Yeah, this tests was failing after I added the 
`PlanBuilder& with(const std::unordered_map<ExprPtr, std::string>& subqueryOutput);`
function, i was thinking that this new function has `ExprPtr` as key not `std::string` which should not be a problem, but somehow it confused the compiler which function we are trying to run here.

PR #83, opened 2025-07-28
Improve error message 'Unsupported PlanNode 8'

mbasmanova:
Summary:
Use node kind name in the error message.

Before: Unsupported PlanNode 8
After: Unsupported PlanNode SET

Differential Revision: D79094324


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #102, opened 2025-08-01
Extract DerivedTable struct into its own .h and .cpp

mbasmanova:
Reviewed By: oerling

Differential Revision: D79443275


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #118, opened 2025-08-06
Implement values node

MBkkt:
Unfortunately I don't understand two things:
1) why ctor is different for `logical_plan::ValuesNode` compared to `core::ValuesNode`.
2) how unique names should be handled in planbuilder


axiom/logical_plan/LogicalPlanNode.h
```diff
@@ -111,24 +112,38 @@ class LogicalPlanNode {
 /// A table whose content is embedded in the plan.
 class ValuesNode : public LogicalPlanNode {
  public:
+  // TODO Implement such ctor, unfortunately transform from rows to values isn't
```
mbasmanova, line 115: > transform from rows to values isn't trivial

Are you looking for fbcode/velox/vector/VariantToVector.h ?

Related PR: https://github.com/facebookincubator/velox/pull/14341

MBkkt: Yep, I tried to use it, but it does a little different thing.

It converts variant value to constant vector.
But here if I understand right, I need to transform vector of rows to single row_vector (aka vector of columns)

MBkkt: So as far as I understand this variantToVector api cannot be used here because core::ValuesNode needs RowVector, but variantToVector produce ConstantVector

mbasmanova: You are right.

You can workaround by calling variantToVector in a loop and copying result to RowVector, but, perhaps, we do need a version of ValuesNode that takes RowVector. How large are these vectors in your application? Wondering if maybe we could add something like `InMemoryTableNode(std::vector<RowVectorPtr>)`.

axiom/optimizer/QueryGraph.cpp
```diff
@@ -50,6 +50,9 @@ std::string Column::toString() const {
   if (!-opt->cnamesInExpr()) {
     return name_;
   }
+  if (relation_ && relation_->type() == PlanType::kValuesTable) {
+    return name_;
+  }
```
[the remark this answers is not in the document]
mbasmanova: All tables are assigned unique correlation names, which allows multiple tables to have same column names. Columns are referred by cname.name.

axiom/logical_plan/LogicalPlanNode.h
```diff
@@ -115,20 +116,24 @@ class ValuesNode : public LogicalPlanNode {
   /// must be non-empty and unique.
   /// @param rows A list of rows. Each row is a list of values, one per column.
   /// The number, order and types of columns must match 'rowType'.
-  ValuesNode(
-      const std::string& id,
-      const RowTypePtr& rowType,
-      std::vector<Variant> rows);
+  ValuesNode(std::string id, RowTypePtr rowType, std::vector<Variant> rows);
+
+  ValuesNode(std::string id, std::vector<RowVectorPtr> values);
```
mbasmanova, line 121: Let's extract extending ValuesNode to allow RowVector inputs into a separate PR.

Let's add comments for the new ctor.

axiom/logical_plan/LogicalPlanNode.h
```diff
@@ -115,20 +116,24 @@ class ValuesNode : public LogicalPlanNode {
...
+  const auto& data() const {
+    return data_;
   }
 
   void accept(const PlanNodeVisitor& visitor, PlanNodeVisitorContext& context)
       const override;
 
  private:
-  const std::vector<Variant> rows_;
+  uint64_t cardinality_ = 0;
```
mbasmanova, line 135: This variable can be const.

axiom/logical_plan/LogicalPlanNode.cpp
```diff
@@ -81,6 +82,24 @@ ValuesNode::ValuesNode(
...
+          NodeKind::kValues,
+          std::move(id),
+          {},
+          values.empty() ? ROW({}) : values.front()->rowType()},
+      data_{std::move(values)} {
+  UniqueNameChecker::check(outputType()->names());
+  for (const auto& value : std::get<std::vector<RowVectorPtr>>(data_)) {
+    VELOX_USER_CHECK_NOT_NULL(value);
+    VELOX_USER_CHECK(
+        *outputType() == *value->type(), "All values should have same type");
```
mbasmanova, line 98: perhaps, include types in the error message to allow for easier troubleshooting

axiom/logical_plan/NameAllocator.cpp
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 #include "axiom/logical_plan/NameAllocator.h"
-#include <velox/common/base/Exceptions.h>
+#include "velox/common/base/Exceptions.h"
```
mbasmanova, line 17: I think we want to keep using angle brackets for Velox includes.

axiom/logical_plan/PlanBuilder.cpp
```diff
@@ -49,6 +49,21 @@ PlanBuilder& PlanBuilder::values(
...
 }
 
+PlanBuilder& PlanBuilder::values(std::vector<RowVectorPtr> values) {
+  VELOX_USER_CHECK_NULL(node_, "Values node must be the leaf node");
+
+  outputMapping_ = std::make_shared<NameMappings>();
+
+  auto rowType = values.empty() ? ROW({}) : values.front()->rowType();
+  for (const auto& name : rowType->names()) {
+    outputMapping_->add(name, name);
```
mbasmanova, line 59: Let's reuse the logic from the other 'values' method to ensure we assign unique names.

axiom/optimizer/Cost.cpp
```diff
@@ -118,10 +117,20 @@ void TableScan::setCost(const PlanState& input) {
...
     return;
-  } else {
-    cost_.fanout =
-        index->distribution().cardinality * baseTable->filterSelectivity;
   }
+  cost_.fanout =
+      index->distribution().cardinality * baseTable->filterSelectivity;
+  auto size = byteSize(columns_);
+  auto numColumns = columns_.size();
+  auto rowCost = numColumns * Costs::kColumnRowCost +
```
mbasmanova, line 125: mark these variables 'const'

This logic seems duplicated in Values::setCost. Would be nice to refactor to avoid copy-paste.

axiom/optimizer/DerivedTable.cpp
```diff
@@ -768,10 +770,8 @@ void DerivedTable::distributeConjuncts() {
             changedDts.push_back(childDt);
           }
         }
-        conjuncts.erase(conjuncts.begin() + i);
-        --numCanonicalConjuncts;
-        --i;
-        continue;
+      } else if (tables[0]->type() == PlanType::kValuesTable) {
```
mbasmanova, line 773: perhaps, move this 'else if' before kDerivedTable for readability

```
      if (tables[0] == this) {
        continue; // the conjunct depends on containing dt, like grouping or
                  // existence flags. Leave in place.
      } else if (tables[0]->type() == PlanType::kValuesTable) {
         continue; // ValuesTable does not have filter push-down.
      } else ...
```

axiom/optimizer/Plan.h
```diff
@@ -791,7 +791,9 @@ class Optimization {
   // Converts a table scan into a BaseTable wen building a DerivedTable.
   PlanObjectP makeBaseTable(const core::TableScanNode* tableScan);
 
-  PlanObjectP makeBaseTable(const logical_plan::TableScanNode* tableScan);
+  PlanObjectP makeBaseTable(const logical_plan::TableScanNode& tableScan);
```
mbasmanova, line 794: This refactoring is welcome. Let's extract in a separate PR though. Thanks.

axiom/optimizer/QueryGraph.h
```diff
@@ -719,6 +720,34 @@ struct BaseTable : public PlanObject {
...
 using BaseTableCP = const BaseTable*;
 
+struct ValuesTable : public PlanObject {
+  explicit ValuesTable(const logical_plan::ValuesNode& values)
+      : PlanObject{PlanType::kValuesTable}, values{values} {}
+
+  // Correlation name, distinguishes between uses of the same values node.
+  Name cname{nullptr};
+
+  const logical_plan::ValuesNode& values;
```
mbasmanova, line 730: Perhaps, avoid dependency on the ValuesNode. Add vectors of variants or RowVectors explicitly.

axiom/optimizer/ToVelox.cpp
```diff
@@ -1025,6 +1025,27 @@ velox::core::PlanNodePtr Optimization::makeUnionAll(
       localSources);
 }
 
+core::PlanNodePtr Optimization::makeValues(const Values& values) {
+  // TODO we need to remove unnecessary columns from the output type vectors
```
mbasmanova, line 1029: Let's address this TODO and add a test case to verify.

axiom/optimizer/ToVelox.cpp
```diff
@@ -1038,32 +1059,28 @@ core::PlanNodePtr Optimization::makeFragment(
     ExecutableFragment& fragment,
     std::vector<ExecutableFragment>& stages) {
   switch (op->relType()) {
-    case RelType::kProject: {
+    case RelType::kProject:
```
mbasmanova, line 1062: This refactoring is welcome. Let's extract it in a separate PR though.

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -786,6 +787,66 @@ TEST_F(PlanTest, except) {
...
 }
 
+TEST_F(PlanTest, values) {
+  auto nationType =
+      ROW({"n_nationkey", "n_regionkey", "n_name", "n_comment"},
+          {BIGINT(), BIGINT(), VARCHAR(), VARCHAR()});
+
+  const std::vector<std::string>& names = nationType->names();
+
+  auto row_vector = makeRowVector({
```
mbasmanova, line 797: naming: row_vector -> rowVector or data

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -786,6 +787,66 @@ TEST_F(PlanTest, except) {
...
+          "nation2",
+          "nation3",
+      }),
+      makeFlatVector<StringView>({
+          "comment1",
+          "comment2",
+          "comment3",
+      }),
+  });
+  row_vector->setType(nationType);
```
mbasmanova, line 819: This is an anti-pattern. Instead,

```
makeRowVector(names, {...});
```

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -786,6 +787,66 @@ TEST_F(PlanTest, except) {
...
+                .values({row_vector})
+                .filter("n_nationkey > 16")
+                .project({"n_nationkey", "n_regionkey"});
+  auto t3 = lp::PlanBuilder(ctx)
+                .values({row_vector})
+                .filter("n_nationkey <= 5")
+                .project({"n_nationkey", "n_regionkey"});
+
+  auto logicalPlan = lp::PlanBuilder(ctx)
+                         .setOperation(lp::SetOperation::kExcept, {t1, t2, t3})
```
mbasmanova, line 836: Any particular reason to test kExcept? 

Let's add tests with joins and column pruning.

mbasmanova (changes requested):
@MBkkt Overall looks great. Thank you for adding this functionality. Some comments.

axiom/logical_plan/LogicalPlanNode.cpp
```diff
@@ -81,6 +82,24 @@ ValuesNode::ValuesNode(
...
 }
 
+ValuesNode::ValuesNode(
+    std::string id,
+    std::vector<RowVectorPtr> values)
+    : LogicalPlanNode{
+          NodeKind::kValues,
+          std::move(id),
+          {},
+          values.empty() ? ROW({}) : values.front()->rowType()},
```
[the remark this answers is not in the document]
MBkkt: I just copied from velox plan node, in general I don't see use cases where it can be useful.

So let me replace it with assert

axiom/logical_plan/NameAllocator.cpp
MBkkt: I just noticed it's different in different places, mostly it's include "

axiom/optimizer/ToVelox.cpp
MBkkt: 1) I don't know how to check this from tests, is there any example?
2) Values operator is really simple, it doesn't affected by count of output columns. Small performance benefits can be for the operator after Values, but I doubt it will be large.

So maybe it's ok to do it in the next PRs?

axiom/optimizer/tests/PlanTest.cpp
MBkkt: No, I just copied from last test in the file.

Ok, I will add test with joins.
Column pruning already exist in this test, it doesn't read all columns, only 2 from 4.

MBkkt: btw expect do joins internally

axiom/optimizer/QueryGraph.cpp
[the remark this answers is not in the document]
MBkkt: Unfortunately I don't see how it can works for RowVector that provided by user.

If I will write in a such way current tests will fail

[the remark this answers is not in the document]
MBkkt: And in general as user of verax we really want to have control over naming.
For an example we generate names in different way compared to PlanBuilder

I mean it's probably good that user of PlanBuilder doesn't need to think about names, but user who build plan by themself IMO should be able to provide their own names.

> In Optimization::makeValuesTable, we would do something like

I already done this, for other places

axiom/optimizer/Plan.h
MBkkt: https://github.com/facebookexperimental/verax/pull/134

axiom/logical_plan/LogicalPlanNode.cpp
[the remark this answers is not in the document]
MBkkt: Done

axiom/logical_plan/LogicalPlanNode.cpp
MBkkt: Done

axiom/logical_plan/NameAllocator.cpp
MBkkt: I removed two changes that was related to this and that was unrelated to this PR

axiom/optimizer/tests/PlanTest.cpp
MBkkt: Done

axiom/logical_plan/PlanBuilder.cpp
```diff
@@ -49,6 +49,45 @@ PlanBuilder& PlanBuilder::values(
...
+    VELOX_USER_CHECK(
+        value->rowType()->equivalent(*rowType),
+        "All values must have the equilent type: {} vs. {}",
+        value->rowType()->toString(),
+        rowType->toString());
+    auto newValue = std::make_shared<RowVector>(
+        value->pool(),
+        rowType,
+        value->nulls(),
+        static_cast<size_t>(value->size()),
```
mbasmanova, line 80: Oh... looks like there is a bug in Vector API where it uses size_t instead of vector_size_t for length / size. Any chance you could make a Velox PR to fix this?

axiom/optimizer/ToVelox.cpp
mbasmanova: CC: @hotcodemacha

MBkkt:
@mbasmanova I added additional test and find that my code doesn't work when `numWorkers != 1`

Maybe you have idea where I should look for this bug?
Probably something is different for Values compared to TableScan and some condition is missed 

Example of wrong plan (numWorkers == 2, numDrivers == 1):

```
experiment: Fragment 0: stage2 numWorkers=2:
-- PartitionedOutput[6][BROADCAST Presto] -> "dt2.n_regionkey_9":BIGINT, "dt2.n_nationkey_8":BIGINT
  -- Project[5][expressions: (dt2.n_regionkey_9:BIGINT, "n_regionkey_9"), (dt2.n_nationkey_8:BIGINT, "n_nationkey_8")] -> "dt2.n_regionkey_9":BIGINT, "dt2.n_nationkey_8":BIGINT
    -- Filter[4][expression: lt("n_regionkey_9",21)] -> n_nationkey_8:BIGINT, n_regionkey_9:BIGINT, n_name_10:VARCHAR, n_comment_11:VARCHAR
      -- Values[3][3 rows in 1 vectors] -> n_nationkey_8:BIGINT, n_regionkey_9:BIGINT, n_name_10:VARCHAR, n_comment_11:VARCHAR

Fragment 1: stage1 numWorkers=2:
-- PartitionedOutput[10][SINGLE Presto] -> "dt1.x1":BIGINT, "dt1.x2":BIGINT, "dt1.y3":VARCHAR
  -- Project[9][expressions: (dt1.x1:BIGINT, "dt2.n_nationkey_8"), (dt1.x2:BIGINT, "dt2.n_regionkey_9"), (dt1.y3:VARCHAR, "dt3.n_comment_15")] -> "dt1.x1":BIGINT, "dt1.x2":BIGINT, "dt1.y3":VARCHAR
    -- HashJoin[8][INNER dt3.n_regionkey_13=dt2.n_regionkey_9] -> "dt2.n_regionkey_9":BIGINT, "dt2.n_nationkey_8":BIGINT, "dt3.n_comment_15":VARCHAR
      -- Project[2][expressions: (dt3.n_regionkey_13:BIGINT, "n_regionkey_13"), (dt3.n_comment_15:VARCHAR, "n_comment_15")] -> "dt3.n_regionkey_13":BIGINT, "dt3.n_comment_15":VARCHAR
        -- Filter[1][expression: gt("n_regionkey_13",11)] -> n_nationkey_12:BIGINT, n_regionkey_13:BIGINT, n_name_14:VARCHAR, n_comment_15:VARCHAR
          -- Values[0][3 rows in 1 vectors] -> n_nationkey_12:BIGINT, n_regionkey_13:BIGINT, n_name_14:VARCHAR, n_comment_15:VARCHAR
      -- Exchange[7][Presto] -> "dt2.n_regionkey_9":BIGINT, "dt2.n_nationkey_8":BIGINT

Inputs:  7 <- stage2 
Fragment 2:  numWorkers=1:
-- Exchange[11][Presto] -> "dt1.x1":BIGINT, "dt1.x2":BIGINT, "dt1.y3":VARCHAR

Inputs:  11 <- stage1 

/home/mironov/serenedb/verax/velox/velox/exec/tests/utils/QueryAssertions.cpp:1098: Failure
Failed
Expected 1, got 4
3 extra rows, 0 missing rows
3 of extra rows:
        2 | 20 | "comment2"
        2 | 20 | "comment2"
        2 | 20 | "comment2"

0 of missing rows:
```

mbasmanova:
> Probably something is different for Values compared to TableScan 

Values is not paralelizable, hence, must run single-threaded in a single-node fragment.

MBkkt:
> Values is not paralelizable, hence, must run single-threaded in a single-node fragment.

@mbasmanova https://github.com/facebookexperimental/verax/pull/118/commits/376d5a73f54ec3534ae20e803b8f1445c7fb6e2f

Is it right?

MBkkt:
> Is it right?

No, it doesn't work in more complicated cases :(

MBkkt:
@mbasmanova I think I fixed it, can you look?

axiom/optimizer/ToVelox.cpp
```diff
@@ -1119,6 +1153,8 @@ core::PlanNodePtr Optimization::makeFragment(
       return makeFragment(op->input(), fragment, stages);
     case RelType::kUnionAll:
       return makeUnionAll(*op->as<UnionAll>(), fragment, stages);
+    case RelType::kValues:
+      return makeValues(*op->as<Values>());
```
mbasmanova, line 1157: I think we should pass 'fragment' to makeValues and unconditionally set fragment.width = 1. 

We don't need to add isParallel to DistributionType.

axiom/optimizer/QueryGraph.cpp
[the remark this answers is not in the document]
mbasmanova: I'm not sure I understand.

> we really want to have control over naming.

These are names internal to the optimizer. The users can control the output column names for the query results, but I don't see why it is reasonable for users to control internal names. Am I missing something?

MBkkt:
@mbasmanova Done

axiom/optimizer/QueryGraph.cpp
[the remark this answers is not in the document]
MBkkt: Ok, I made it this like you suggested.

In general I think I just completely dislike all this overhead introduced by string names and the way how this handled.
I really believe it can be done a lot of simpler maybe with a little less fancy output.
Something like everything is unique integers and hashmap on user side that maps unique integer to non unique complicated names (catalog + schema + name + whatever needed).

```
experiment: Fragment 0:  numWorkers=1:
-- Project[15][expressions: (dt1.x1:BIGINT, "dt2.n_nationkey_640"), (dt1.x2:BIGINT, "dt2.n_regionkey_641"), (dt1.y1:BIGINT, "dt4.n_nationkey_647"), (dt1.z1:BIGINT, "dt6.n_nationkey_654"), (dt1.w3:VARCHAR, "dt8.n_comment_664")] -> "dt1.x1":BIGINT, "dt1.x2":BIGINT, "dt1.y1":BIGINT, "dt1.z1":BIGINT, "dt1.w3":VARCHAR
  -- HashJoin[14][INNER dt4.n_regionkey_648=dt2.n_regionkey_641] -> "dt2.n_regionkey_641":BIGINT, "dt2.n_nationkey_640":BIGINT, "dt4.n_nationkey_647":BIGINT, "dt6.n_nationkey_654":BIGINT, "dt8.n_comment_664":VARCHAR
    -- HashJoin[10][INNER dt6.n_regionkey_655=dt4.n_regionkey_648] -> "dt4.n_regionkey_648":BIGINT, "dt4.n_nationkey_647":BIGINT, "dt6.n_regionkey_655":BIGINT, "dt6.n_nationkey_654":BIGINT, "dt8.n_regionkey_662":BIGINT, "dt8.n_comment_664":VARCHAR
      -- HashJoin[6][INNER dt8.n_regionkey_662=dt6.n_regionkey_655] -> "dt6.n_regionkey_655":BIGINT, "dt6.n_nationkey_654":BIGINT, "dt8.n_regionkey_662":BIGINT, "dt8.n_comment_664":VARCHAR
        -- Project[2][expressions: (dt8.n_regionkey_662:BIGINT, "vt9.n_regionkey_662"), (dt8.n_comment_664:VARCHAR, "vt9.n_comment_664")] -> "dt8.n_regionkey_662":BIGINT, "dt8.n_comment_664":VARCHAR
          -- Filter[1][expression: gt("vt9.n_regionkey_662",1)] -> "vt9.n_regionkey_662":BIGINT, "vt9.n_comment_664":VARCHAR
            -- Values[0][3 rows in 1 vectors] -> "vt9.n_regionkey_662":BIGINT, "vt9.n_comment_664":VARCHAR
        -- Project[5][expressions: (dt6.n_regionkey_655:BIGINT, "vt7.n_regionkey_655"), (dt6.n_nationkey_654:BIGINT, "vt7.n_nationkey_654")] -> "dt6.n_regionkey_655":BIGINT, "dt6.n_nationkey_654":BIGINT
          -- Filter[4][expression: gt(3,"vt7.n_regionkey_655")] -> "vt7.n_nationkey_654":BIGINT, "vt7.n_regionkey_655":BIGINT
            -- Values[3][3 rows in 1 vectors] -> "vt7.n_nationkey_654":BIGINT, "vt7.n_regionkey_655":BIGINT
      -- Project[9][expressions: (dt4.n_regionkey_648:BIGINT, "vt5.n_regionkey_648"), (dt4.n_nationkey_647:BIGINT, "vt5.n_nationkey_647")] -> "dt4.n_regionkey_648":BIGINT, "dt4.n_nationkey_647":BIGINT
        -- Filter[8][expression: gt(3,"vt5.n_regionkey_648")] -> "vt5.n_nationkey_647":BIGINT, "vt5.n_regionkey_648":BIGINT
          -- Values[7][3 rows in 1 vectors] -> "vt5.n_nationkey_647":BIGINT, "vt5.n_regionkey_648":BIGINT
    -- Project[13][expressions: (dt2.n_regionkey_641:BIGINT, "vt3.n_regionkey_641"), (dt2.n_nationkey_640:BIGINT, "vt3.n_nationkey_640")] -> "dt2.n_regionkey_641":BIGINT, "dt2.n_nationkey_640":BIGINT
      -- Filter[12][expression: lt("vt3.n_regionkey_641",3)] -> "vt3.n_nationkey_640":BIGINT, "vt3.n_regionkey_641":BIGINT
        -- Values[11][3 rows in 1 vectors] -> "vt3.n_nationkey_640":BIGINT, "vt3.n_regionkey_641":BIGINT
```

axiom/optimizer/Cost.cpp
```diff
@@ -96,9 +96,21 @@ float orderPrefixDistance(
   return selection;
 }
 
+namespace {
+
+void updateCost(float cardinality, ColumnVector& columns, Cost& cost) {
```
mbasmanova, line 101: Can columns be const &?

axiom/optimizer/LogicalPlanToGraph.cpp
```diff
@@ -1346,9 +1375,7 @@ PlanObjectP Optimization::makeQueryGraph(
     uint64_t allowedInDt) {
   switch (node.kind()) {
     case lp::NodeKind::kValues:
-      VELOX_NYI(
-          "Unsupported PlanNode {}",
-          logical_plan::NodeKindName::toName(node.kind()));
+      return makeValuesTable(*node.asUnchecked<lp::ValuesNode>());
```
mbasmanova, line 1378: nice :-)

axiom/optimizer/QueryGraph.h
mbasmanova: > why do you consider such dependency as something bad?

Primarily because currently QG doesn't depend on the API (Logical Plan).

BTW, how will we ensure that values reference will stay valid? Are we relying on optimizer running single-threaded??

axiom/optimizer/RelationOp.cpp
```diff
@@ -213,6 +213,28 @@ std::string TableScan::toString(bool /*recursive*/, bool detail) const {
...
+  key_ = sanitizeHistoryKey(out.str());
+  return key_;
+}
+
+std::string Values::toString(bool /*recursive*/, bool detail) const {
+  std::stringstream out;
+  out << valuesTable.values.id() << " " << valuesTable.cname;
+  if (detail) {
+    printCost(detail, out);
+    if (!input()) {
```
mbasmanova, line 231: Can Values node have an input? Is this check always true?

axiom/optimizer/ToVelox.cpp
```diff
@@ -1086,6 +1087,64 @@ velox::core::PlanNodePtr Optimization::makeUnionAll(
...
+  fragment.width = 1;
+  const auto& newColumns = values.columns();
+  const auto newType = makeOutputType(newColumns);
+  VELOX_DCHECK_EQ(newColumns.size(), newType->size());
+
+  const auto& data = values.valuesTable.values.data();
+  std::vector<RowVectorPtr> newValues;
+  if ([[maybe_unused]] auto* row = std::get_if<std::vector<Variant>>(&data)) {
+    [[maybe_unused]] auto& newValue = newValues.emplace_back();
+    VELOX_NYI("Translate rows from vector<Variant> to RowVector");
```
mbasmanova, line 1102: Let's create an issue to follow-up.

axiom/optimizer/ToVelox.cpp
```diff
@@ -1086,6 +1087,64 @@ velox::core::PlanNodePtr Optimization::makeUnionAll(
...
+  } else {
+    const auto& oldValues = std::get<std::vector<RowVectorPtr>>(data);
+    newValues.reserve(oldValues.size());
+
+    VELOX_DCHECK(!oldValues.empty());
+    const auto oldType = oldValues.front()->rowType();
+
+    std::vector<uint32_t> oldColumnIdxs;
+    oldColumnIdxs.reserve(newColumns.size());
+    // TODO: This place and a lot of other that use getChildIdx/etc results to
```
mbasmanova, line 1112: That's a good point. There was an attempt to do that, but it had to be reverted: 

- original PR: https://github.com/facebookincubator/velox/pull/5728
- backout: https://github.com/facebookincubator/velox/pull/8245

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -786,6 +787,355 @@ TEST_F(PlanTest, except) {
...
+    auto plan = exec::test::PlanBuilder(idGenerator, pool_.get());
+    if (leafType == 0) {
+      plan.tableScan("nation", nationType);
+    } else {
+      plan.values({rowVector});
+    }
+    return plan;
+  };
+
+  struct Combination {
```
mbasmanova, line 848: It is a bit hard to tell what these are. Would it be possible to add a comment or come up with a way to make the test a bit easier to read?

Do we verify that we do not create unnecessary remote exchanges? I believe you have logic to ensure that, but it is not clear to me if this is being verified in a test.

mbasmanova (commented):
@MBkkt Thank you for iterating. This is looking great % the question of whether we need a separate isParallel boolean or can use isGather.

axiom/optimizer/ToVelox.cpp
MBkkt: Do you have benchmark for this? I think we can do a little hack, compared to old attempt, to make it noticably less memory usage:
1. using string_view in the map (it will cause FastMap will be ValueMap instead of VectorMal btw)
2. using string_view in the search
3. use F14ValueMap instead of fast (fast was F14VectorMap which is slower)
4. make creation a bit faster

So additional memory usage will be something like (24 * count of columns (n) + ~1 bit per  column) / fill factor for hash map instead of 32(libc++)/48(libstdc++) * n * (from 1 to vector growth factor) + (8*count of columns + ~1bit per column) / fill factor


Also do you consider abseil usage instead? As far as I know they spent a lot of effort on optimizing hash tables for small (1-100 sizes) tables compared to F14, at the same time they have comparable performance for medium-large tables.
Also their fill factor 0.875

axiom/optimizer/Schema.h
```diff
@@ -146,6 +146,7 @@ struct DistributionType {
   int32_t numPartitions{1};
   LocusCP locus{nullptr};
   bool isGather{false};
+  bool isParallel{true};
```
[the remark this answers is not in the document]
MBkkt: isGather works different way, it's doesn't prevent shuffles from what I saw

axiom/optimizer/ToVelox.cpp
MBkkt: > Perhaps, we can keep it that way in this PR and optimize in a follow-up.

Tests what I wrote doesn't check do we have shuffle or not.
But they're crashed if I remove `isParallel`.

Maybe if you have some thoughts how this `isParallel` can be avoided you can try it after merge?
So follow-up will be avoiding it/replacing with some different ideas.

axiom/optimizer/ToVelox.cpp
MBkkt: https://github.com/facebookexperimental/verax/issues/149

axiom/optimizer/tests/PlanTest.cpp
mbasmanova: > if they're generated something crashed on execution

I looked into this crash. Turns out it is a bug in how ToVelox sets numBroadcastDestinations in Optimization::makeRepartition. It ends up producing a plan where stage 3 broadcasts to stage 2 with numBroadcastDestinations == 4, but stage 2 has only one worker.

A fix would be to change

```
  if (distribution.isBroadcast) {
    source.numBroadcastDestinations = options_.numWorkers;
  }
```

to 

```
  if (distribution.isBroadcast) {
    source.numBroadcastDestinations = fragment.width;
  }
```

and, I think we also need to fix similar place further down

```
(keys.empty()) ? 1 : fragment.width,
```

Hope this makes sense.

axiom/optimizer/ToVelox.cpp
MBkkt: Made PR, it's a little better than I described https://github.com/facebookincubator/velox/pull/14403

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -831,6 +832,367 @@ TEST_F(PlanTest, except) {
...
+    auto plan = exec::test::PlanBuilder(idGenerator, pool_.get());
+    if (leafType == 0) {
+      plan.tableScan("nation", nationType);
+    } else {
+      plan.values({rowVector});
+    }
+    return plan;
+  };
+
+  // In this test, we verify that the optimizer can handle
```
mbasmanova, line 893: This comment is very helpful. 

I think we can shorten test code as well to make it easier to read and maintain.

Here is one way we can do that: https://gist.github.com/mbasmanova/f104389b3e0c83bc33b2566a294fb50c

axiom/optimizer/tests/PlanTest.cpp
mbasmanova: I think we can get rid of numBroadcastDestinations altogether and have Runner figure it out on its own. This information is redundant.

axiom/optimizer/ToVelox.cpp
mbasmanova: Now that we have figured out the crash, we have an option of not introducing isParallel in this PR and following up to figure out how to avoid adding unnecessary shuffles. This is preferable since introducing isParallel doesn't seem right and will confuse other developers.

axiom/optimizer/QueryGraph.h
mbasmanova: Got it. Thank you for clarifying.

axiom/optimizer/tests/PlanTest.cpp
MBkkt: Oh I see now, `fragment.width` isn't `source.width` so it makes sense to me now

MBkkt: I applied suggested fix, removed isParallel, and also your applied refactoring to tests

axiom/optimizer/Cost.cpp
```diff
@@ -96,9 +96,21 @@ float orderPrefixDistance(
   return selection;
 }
 
+namespace {
+
+void updateCost(float cardinality, const ColumnVector& columns, Cost& cost) {
```
mbasmanova, line 101: nit: updateCost -> updateLeafCost

It is confusing that for leaf nodes fanout and unitCost have a different meaning than their names suggest. For the leaf, fanout is cardinality and unitCost is total cost. For non-leaf, fanout is cardinality change (output card / input card) and unitCost is a cost per-row.

Might be helpful to add a comment here to help future readers.

axiom/optimizer/LogicalPlanToGraph.cpp
```diff
@@ -1040,6 +1042,33 @@ PlanObjectP Optimization::makeBaseTable(const lp::TableScanNode& tableScan) {
...
+  auto* valuesTable = make<ValuesTable>(values);
+  valuesTable->cname = newCName("vt");
+  logicalPlanLeaves_[&values] = valuesTable;
+
+  auto channels = usedChannels(&values);
+  const auto& type = values.outputType();
+  const auto& names = values.outputType()->names();
+  const auto cardinality = valuesTable->cardinality();
+  for (uint32_t i = 0; i < type->size(); ++i) {
+    if (std::find(channels.begin(), channels.end(), i) == channels.end()) {
```
mbasmanova, line 1055: Any reason not to loop over channels?

axiom/optimizer/Plan.cpp
```diff
@@ -943,8 +946,9 @@ void Optimization::joinByIndex(
     const JoinCandidate& candidate,
     PlanState& state,
     std::vector<NextJoin>& toTry) {
-  if (candidate.tables.at(0)->type() != PlanType::kTable ||
-      candidate.tables.size() > 1 || !candidate.existences.empty()) {
+  if (candidate.tables.size() != 1 ||
```
mbasmanova, line 949: the original code seems clearer: > 1 vs. != 1

axiom/optimizer/Plan.cpp
```diff
@@ -1079,6 +1087,7 @@ void Optimization::joinByHash(
   auto memoKey = MemoKey{
       candidate.tables[0], buildColumns, buildTables, candidate.existences};
   PlanObjectSet empty;
+  const bool canBePartitioned = !isSingle_;
```
mbasmanova, line 1090: no longer needed

axiom/optimizer/Plan.cpp
```diff
@@ -1655,18 +1664,34 @@ void Optimization::makeJoins(RelationOpPtr plan, PlanState& state) {
...
-              columns);
-
-          state.columns.unionObjects(columns);
+              std::move(columns));
           state.addCost(*scan);
           makeJoins(scan, state);
         }
+      } else if (from->type() == PlanType::kValuesTable) {
+        const auto* valuesTable = from->as<ValuesTable>();
+        auto downstream = state.downstreamColumns();
```
mbasmanova, line 1680: nit: perhaps, this variable is not needed

axiom/optimizer/PlanObject.h
```diff
@@ -37,7 +37,8 @@ enum class PlanType {
   kOrderBy,
   kLimit,
   kField,
-  kLambda
+  kLambda,
+  kValuesTable,
```
mbasmanova, line 41: let's move it next to kTable

axiom/optimizer/ToVelox.cpp
```diff
@@ -1138,6 +1138,64 @@ velox::core::PlanNodePtr Optimization::makeUnionAll(
...
+    // O(n * m) complexity, so it doesn't work well for large projections
```
mbasmanova, line 1164: Let's remove this comment. It applies to many places and this place is not particularly special.

mbasmanova (approved):
@MBkkt Thank you for iterating. Looks great.

mbasmanova:
@MBkkt Seeing lint errors

Screenshot 2025-08-09 at 7 54 45 PM

axiom/optimizer/LogicalPlanToGraph.cpp
MBkkt: It's copied from table scan, I agree that it looks strange and can produce O(n*m). Lets fix it :)

axiom/optimizer/QueryGraph.cpp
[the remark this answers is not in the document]
mbasmanova: > I just completely dislike all this overhead introduced by string names and the way how this handled.

Are you referring to std::string use for names in Velox plans? In Verax, these are interned strings (char*) which doesn't have much overhead. Am I missing something?

axiom/optimizer/Plan.cpp
MBkkt: Original code will throw exception in case of `size() == 0` (maybe impossible but still).
And I don't understand why is it good, with this condition it's exception free and we're also avoid additional check inside `.at(0)`

axiom/optimizer/Plan.cpp
MBkkt: What do you mean? It's used and it's written same way as for `kTable` above

axiom/optimizer/Cost.cpp
MBkkt: Added

axiom/optimizer/LogicalPlanToGraph.cpp
MBkkt: Fixed

axiom/optimizer/PlanObject.h
MBkkt: Moved

axiom/optimizer/ToVelox.cpp
MBkkt: Removed

mbasmanova:
> Btw I found place where input possible for TableScan (but not for Values)

Oh.. makes sense.

axiom/optimizer/Plan.cpp
MBkkt: Not sure I understand you right, but I made this change https://github.com/facebookexperimental/verax/pull/118/commits/931833aa1b1cf287475dea9c6de797e2a04d911d

axiom/optimizer/QueryGraph.cpp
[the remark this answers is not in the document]
MBkkt: > which doesn't have much overhead.

Field itself? Yes their sizeof == 8.

But it's also arena allocation, memcpy to fill data in this arena.

But if we're looking on what we need to do to construct these names it's different allocations + a lot memcpy (e.g. these "vt.name_0" names).

Also c-string works quite bad when we're doing something with them, because we need to compute length every time

axiom/optimizer/Plan.cpp
mbasmanova: Thanks. Looks good.

MBkkt:
@mbasmanova I see that some CI is red, should I fix something or is it ok?

PR #119, opened 2025-08-06
Connector Interface for Table Writing

oerling:
Introduces API for coordinating distributed queries with DML operations.

the initiating and concluding operations are performed on a single node. These are before the first and after completion of the last table write operation. Inserts, deletes and updates are all modeled as table writes.

The connector specifies what serves as a row identifier for creating a delete or update record.

There is a notion of session that can encapsulate any transaction context. It is passed to the relevant API calls and connectors that support transactions can embed the context information into the created handles.

Other DDL operations will be added in follow ups.


axiom/optimizer/connectors/ConnectorMetadata.h
```diff
@@ -570,6 +601,79 @@ class ConnectorMetadata {
       const std::string& queryId) {
     VELOX_UNSUPPORTED();
   }
+
+  /// Creates an insert table handle for use with Velox TableWriter. '
```
pedroerp, line 605: nit: extra ' at the end

oerling:
Updated the comments.

 

Getting the shuffle information is connector metadata and not insert table handle because the latter does not know. It could be there is a shuffle on bucket and hive partition columns or something else. Also the handle is a Velox runtime object that does not know about the plan, just the schema of the table. You could just as well write a bucketed table without a shuffle with a single worker. Andwe want the changes to be limited to the front end. Mus hav an inimplementation to know for sure.

PR #121, opened 2025-08-06
Extend HiveScanMatcher to match remaining filter

mbasmanova:
Summary: Use the new capabilities to tighten plan verification in PlanTest.cpp and LogicalSubfieldTest.cpp

Differential Revision: D79699399


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #128, opened 2025-08-07
Add In and Exists expression

feilong-liu:
Summary: Add In and Exists subquery in logical plan node

Differential Revision: D79703347

### DIFF
--- a/axiom/logical_plan/PlanBuilder.cpp
+++ b/axiom/logical_plan/PlanBuilder.cpp
@@ -502,6 +502,16 @@ ExprPtr tryResolveSpecialForm(
         rowType.childAt(zeroBasedIndex), SpecialForm::kDereference, newInputs);
   }
 
+  if (name == "in") {
+    return std::make_shared<SpecialFormExpr>(
+        BOOLEAN(), SpecialForm::kIn, resolvedInputs);
+  }
+
+  if (name == "exists") {
+    return std::make_shared<SpecialFormExpr>(
+        BOOLEAN(), SpecialForm::kExists, resolvedInputs);
+  }
+
   return nullptr;
 }
 } // namespace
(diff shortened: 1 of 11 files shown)


axiom/logical_plan/PlanBuilder.cpp
```diff
@@ -695,6 +705,33 @@ ExprPtr ExprResolver::tryFoldCall(
   return nullptr;
 }
 
+ExprPtr ExprResolver::tryFoldSpeicalForm(
```
mbasmanova, line 708: typo: tryFoldSpeicalForm -> tryFoldSpecialForm

axiom/logical_plan/PlanBuilder.cpp
```diff
@@ -768,6 +805,9 @@ ExprPtr ExprResolver::resolveScalarTypes(
       }
     }
 
+    if (auto folded = tryFoldSpeicalForm(name, inputs)) {
```
mbasmanova, line 808: Shouldn't we first call tryResolveSpecialForm, then try-fold?

axiom/optimizer/ToVelox.cpp
```diff
@@ -265,6 +265,28 @@ template <typename T>
 core::TypedExprPtr makeKey(const TypePtr& type, T v) {
   return std::make_shared<core::ConstantTypedExpr>(type, variant(v));
 }
+
+core::TypedExprPtr createArrayForInList(
+    const Call* call,
```
mbasmanova, line 270: pass as const &

axiom/optimizer/LogicalPlanToGraph.cpp
```diff
@@ -549,6 +549,10 @@ const char* specialFormCallName(const lp::SpecialFormExpr* form) {
       return "if";
     case lp::SpecialForm::kSwitch:
       return "switch";
+    case lp::SpecialForm::kIn:
+      return "in";
+    case lp::SpecialForm::kExists:
+      return "exists";
```
mbasmanova, line 555: There is no such function in Velox. Let's remove this case for now since it doesn't work.

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -281,6 +281,63 @@ TEST_F(PlanTest, rejectedFilters) {
...
+                           .tableScan(kTestConnectorId, "numbers", {"a", "b"})
+                           .filter("4 in (1, 2, 3)")
+                           .map({"a + 2"})
+                           .build();
+
+    auto plan = toSingleNodePlan(logicalPlan);
+
+    auto matcher = core::PlanMatcherBuilder()
+                       .tableScan()
+                       .filter("false")
```
mbasmanova, line 315: It seems incorrect that the optimizer produces a filter node that never passes. Is this really working like that? If so, let's create an issue to fix? Same for filter("true").

mbasmanova (commented):
Looks good % some nits. CI is red. I assume you'll need to advance Velox to make it green.

feilong-liu:
> Looks good % some nits. CI is red. I assume you'll need to advance Velox to make it green.

Yes, will do after https://github.com/facebookincubator/velox/pull/14396/ merged

mbasmanova (approved):

PR #131, opened 2025-08-07
test: fill out TestConnector fixture and add automation

hdikeman:
Summary:
Adding the rest of the functionality required to optimize and execute a plan using a TestConnector as the sink or source for query execution. Data can be appended to a table either directly via TestConnector API or through the DataSink interface as part of table write (though the latter is not currently supported by Verax)

Additional pieces are also being added to support query execution, such as implementations of interfaces required for split generation. Test automation for the TestConnector is also included, in the file TestConnectorTest.cpp

Differential Revision: D79839654

### DIFF
--- a/axiom/optimizer/connectors/tests/CMakeLists.txt
+++ b/axiom/optimizer/connectors/tests/CMakeLists.txt
@@ -16,3 +16,22 @@ velox_add_library(velox_test_connector TestConnector.cpp)
 
 velox_link_libraries(velox_test_connector velox_connector_metadata
                      velox_common_base velox_memory velox_connector)
+
+add_executable(
+  velox_test_connector_test
+  TestConnectorTest.cpp)
+
+add_test(
+    velox_test_connector_test
+    velox_test_connector_test)
+
+target_link_libraries(
+  velox_test_connector_test
+  velox_test_connector
+  velox_common_base
+  velox_memory
+  velox_connector
+  velox_exec
+  velox_vector_test_lib
+  gtest
+  gtest_main)
--- a/axiom/optimizer/connectors/tests/TestConnector.cpp
+++ b/axiom/optimizer/connectors/tests/TestConnector.cpp
@@ -10,24 +10,234 @@
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * See the License for the specific language governing
+ * permissions and limitations under the License.
  */
-
 #include "axiom/optimizer/connectors/tests/TestConnector.h"
 
 namespace facebook::velox::connector {
 
+TestTable::TestTable(
+    const std::string& name,
+    const RowTypePtr& schema,
+    TestConnector* connector)
+    : Table(name), connector_(connector) {
+  Table::type_ = schema;
+  std::vector<const Column*> columnVector;
+
+  exportedColumns_.reserve(schema->size());
+  columnVector.reserve(schema->size());
+  for (auto i = 0; i < schema->size(); ++i) {
+    const auto& columnName = schema->nameOf(i);
+    const auto& columnType = schema->childAt(i);
+    VELOX_CHECK(
+        !columnName.empty(), "column {} in table {} has empty name", i, name);
+    exportedColumns_.emplace_back(
+        std::make_unique<Column>(columnName, columnType));
+    columnVector.emplace_back(exportedColumns_.back().get());
+    auto [_, ok] = columns_.emplace(columnName, exportedColumns_.back().get());
+    VELOX_CHECK(ok, "duplicate column name '{}' in table {}", columnName, name);
+  }
+
+  auto layout =
+      std::make_unique<TestTableLayout>(name_, this, connector_, columnVector);
+  layouts_.push_back(layout.get());
+  exportedLayouts_.push_back(std::move(layout));
+  pool_ = memory::memoryManager()->addLeafPool(name + "_table");
+}
+
+std::vector<SplitSource::SplitAndGroup> TestSplitSource::getSplits(uint64_t) {
+  std::vector<SplitAndGroup> result;
+  if (currentPartition_ >= partitions_.size()) {
+    result.push_back({nullptr, kUngroupedGroupId});
+  } else {
+    result.push_back(
+        {std::make_shared<ConnectorSplit>(connectorId_), kUngroupedGroupId});
+  }
+  currentPartition_++;
+  return result;
+}
+
+std::vector<PartitionHandlePtr> TestSplitManager::listPartitions(
+    const ConnectorTableHandlePtr&) {
+  return {std::make_shared<PartitionHandle>()};
+}
+
+std::shared_ptr<SplitSource> TestSplitManager::getSplitSource(
+    const ConnectorTableHandlePtr& tableHandle,
+    std::vector<PartitionHandlePtr> partitions,
+    SplitOptions) {
+  return std::make_shared<TestSplitSource>(
+      tableHandle->connectorId(), std::move(partitions));
+}
+
+Table* FOLLY_NULLABLE
+TestConnectorMetadata::findTableInternal(const std::string& name) {
+  auto it = tables_.find(name);
+  return it != tables_.end() ? it->second.get() : nullptr;
+}
+
+const Table* FOLLY_NULLABLE
+TestConnectorMetadata::findTable(const std::string& name) {
+  return findTableInternal(name);
+}
+
+ColumnHandlePtr TestConnectorMetadata::createColumnHandle(
+    const TableLayout& layout,
+    const std::string& columnName,
+    std::vector<common::Subfield>,
+    std::optional<TypePtr> castToType,
+    SubfieldMapping) {
+  auto column = layout.findColumn(columnName);
+  VELOX_CHECK_NOT_NULL(
+      column, "Column {} not found in table {}", columnName, layout.name());
+  return std::make_shared<TestColumnHandle>(
+      columnName, castToType.value_or(column->type()));
+}
+
 ConnectorTableHandlePtr TestConnectorMetadata::createTableHandle(
-    const TableLayout& /* layout */,
-    std::vector<ColumnHandlePtr> /* columnHandles */,
+    const TableLayout& layout,
+    std::vector<ColumnHandlePtr> columnHandles,
     core::ExpressionEvaluator& /* evaluator */,
     std::vector<core::TypedExprPtr> filters,
     std::vector<core::TypedExprPtr>& rejectedFilters,
     RowTypePtr /* dataColumns */,
     std::optional<LookupKeys>) {
   rejectedFilters = std::move(filters);
-  return std::make_shared<TestTableHandle>(connector_->connectorId());
+  return std::make_shared<TestTableHandle>(layout, std::move(columnHandles));
+}
+
+TestTable* TestConnectorMetadata::createTable(
+    const std::string& name,
+    const RowTypePtr& schema) {
+  auto table = std::make_unique<TestTable>(name, schema, connector_);
+  auto [it, ok] = tables_.emplace(name, std::move(table));
+  VELOX_CHECK(ok, "table {} already exists", name);
+  return it->second.get();
+}
+
+void TestConnectorMetadata::appendData(
+    const std::string& name,
+    const RowVectorPtr& data) {
+  auto it = tables_.find(name);
+  VELOX_CHECK(it != tables_.end(), "no table {} exists", name);
+  it->second->addData(data);
+}
+
+TestDataSource::TestDataSource(
+    const RowTypePtr& outputType,
+    const ColumnHandleMap& handles,
+    const Table* table,
+    memory::MemoryPool* pool)
+    : outputType_(outputType), pool_(pool) {
+  auto maybeTable = dynamic_cast<const TestTable*>(table);
+  VELOX_CHECK(maybeTable, "table {} not a TestTable", table->name());
+  data_ = maybeTable->data();
+
+  auto tableType = table->rowType();
+  outputMappings_.reserve(outputType_->size());
+  for (const auto& name : outputType->names()) {
+    VELOX_CHECK(
+        handles.contains(name),
+        "no handle for output column {} for table {}",
+        name,
+        table->name());
+    auto handle = handles.find(name)->second;
+
+    const auto idx = tableType->getChildIdxIfExists(handle->name());
+    VELOX_CHECK(
+        idx.has_value(),
+        "column '{}' not found in table '{}'.",
+        handle->name(),
+        table->name());
+    outputMappings_.emplace_back(idx.value());
+  }
+}
+
+void TestDataSource::addSplit(std::shared_ptr<ConnectorSplit> split) {
+  split_ = std::move(split);
+}
+
+std::optional<RowVectorPtr> TestDataSource::next(
+    uint64_t,
+    velox::ContinueFuture&) {
+  VELOX_CHECK(split_, "no split added to DataSource");
+  if (data_.size() <= idx_) {
+    return nullptr;
+  }
+  auto vector = data_[idx_++];
+
+  completedRows_ += vector->size();
+  completedBytes_ += vector->retainedSize();
+
+  std::vector<VectorPtr> children;
+  children.reserve(outputMappings_.size());
+  for (const auto idx : outputMappings_) {
+    children.emplace_back(vector->childAt(idx));
+  }
+
+  return std::make_shared<RowVector>(
+      pool_, outputType_, BufferPtr(), vector->size(), std::move(children));
+}
+
+void TestDataSource::addDynamicFilter(
+    column_index_t,
+    const std::shared_ptr<common::Filter>&) {
+  VELOX_NYI("TestDataSource does not support dynamic filters");
+}
+
+std::unique_ptr<DataSource> TestConnector::createDataSource(
+    const RowTypePtr& outputType,
+    const ConnectorTableHandlePtr& tableHandle,
+    const ColumnHandleMap& columnHandles,
+    ConnectorQueryCtx* connectorQueryCtx) {
+  auto table = metadata_->findTable(tableHandle->name());
+  VELOX_CHECK(
+      table,
+      "cannot create data source for nonexistent table {}",
+      tableHandle->name());
+  return std::make_unique<TestDataSource>(
+      outputType, columnHandles, table, connectorQueryCtx->memoryPool());
+}
+
+std::unique_ptr<DataSink> TestConnector::createDataSink(
+    RowTypePtr,
+    ConnectorInsertTableHandlePtr tableHandle,
+    ConnectorQueryCtx*,
+    CommitStrategy) {
+  VELOX_CHECK(tableHandle, "table handle must be non-null");
+  auto table = metadata_->findTableInternal(tableHandle->toString());
+  VELOX_CHECK(
+      table,
+      "cannot create data sink for nonexistent table {}",
+      tableHandle->toString());
+  return std::make_unique<TestDataSink>(table);
+}
+
+TestTable* TestConnector::createTable(
+    const std::string& name,
+    const RowTypePtr& schema) {
+  return metadata_->createTable(name, schema);
+}
+
+void TestConnector::appendData(
+    const std::string& name,
+    const RowVectorPtr& data) {
+  metadata_->appendData(name, data);
+}
+
+std::shared_ptr<Connector> TestConnectorFactory::newConnector(
+    const std::string& id,
+    std::shared_ptr<const config::ConfigBase>,
+    folly::Executor*,
+    folly::Executor*) {
+  return std::make_shared<TestConnector>(id);
+}
+
+void TestDataSink::appendData(RowVectorPtr vector) {
+  if (vector) {
+    table_->addData(vector);
+  }
 }
 
 } // namespace facebook::velox::connector
--- a/axiom/optimizer/connectors/tests/TestConnector.h
+++ b/axiom/optimizer/connectors/tests/TestConnector.h
@@ -22,6 +22,9 @@ namespace facebook::velox::connector {
 
 class TestConnector;
 
+/// The Table and Connector objects to which this layout correspond
+/// are specified explicitly at init time. The sample API is
+/// overridden to provide placeholder counts.
 class TestTableLayout : public TableLayout {
  public:
   TestTableLayout(
@@ -34,77 +37,187 @@ class TestTableLayout : public TableLayout {
             table,
             connector,
             std::move(columns),
-            std::vector<const Column*>{},
-            std::vector<const Column*>{},
-            std::vector<SortOrder>{},
-            std::vector<const Column*>{},
-            true) {}
+            /*partitionColumns=*/{},
+            /*orderColumns=*/{},
+            /*sortOrder=*/{},
+            /*lookupKeys=*/{},
+            /*supportsScan=*/true) {}
 
   std::pair<int64_t, int64_t> sample(
-      const connector::ConnectorTableHandlePtr& /* handle */,
-      float /* pct */,
-      std::vector<core::TypedExprPtr> /* extraFilters */,
-      RowTypePtr /* outputType */ = nullptr,
-      const std::vector<common::Subfield>& /* fields */ = {},
-      HashStringAllocator* /* allocator */ = nullptr,
-      std::vector<ColumnStatistics>* /* statistics */ = nullptr) const {
+      const connector::ConnectorTableHandlePtr&,
+      float,
+      std::vector<core::TypedExprPtr>,
+      RowTypePtr,
+      const std::vector<common::Subfield>&,
+      HashStringAllocator*,
+      std::vector<ColumnStatistics>*) const override {
     return std::make_pair(1'000, 1'000);
   }
 };
 
+/// RowVectors are appended using the addData() interface and the vector
+/// of RowVectors are retrieved using the data() interface. Appended
+/// data is copied inside an internal memory pool associated with
+/// the table. Row count is determined dynamically using a summation
+/// of row counts for RowVectors currently stored within the table.
 class TestTable : public Table {
  public:
   TestTable(
       const std::string& name,
       const RowTypePtr& schema,
-      TestConnector* connector)
-      : Table(name), connector_{connector} {
-    type_ = schema;
-
-    columns_.reserve(schema->size());
-
-    std::vector<const Column*> columnPtrs;
-    columnPtrs.reserve(schema->size());
-
-    for (auto i = 0; i < schema->size(); ++i) {
-      auto column =
-          std::make_shared<Column>(schema->nameOf(i), schema->childAt(i));
-      columns_.emplace_back(column);
-      columnPtrMap_.emplace(column->name(), column.get());
-      columnPtrs.emplace_back(column.get());
-    }
-
-    layout_ =
-        std::make_shared<TestTableLayout>(name, this, connector, columnPtrs);
-    layoutPtrs_.emplace_back(layout_.get());
-  }
+      TestConnector* connector);
 
   const std::unordered_map<std::string, const Column*>& columnMap()
       const override {
-    return columnPtrMap_;
+    return columns_;
   }
 
   const std::vector<const TableLayout*>& layouts() const override {
-    return layoutPtrs_;
+    return layouts_;
   }
 
   uint64_t numRows() const override {
-    return 100;
+    uint64_t rows = 0;
+    for (const auto& vector : data_) {
+      rows += vector->size();
+    }
+    return rows;
+  }
+
+  const std::vector<RowVectorPtr>& data() const {
+    return data_;
+  }
+
+  /// Copy the specified RowVector into the internal data of the
+  /// table. The underlying types of the columns must match the
+  /// schema specified during initial table creation.
+  /// Data is copied on append so that vectors from temporary
+  /// memory pools can be appended. These copies are allocated
+  /// via the TestTable internal memory pool.
+  void addData(const RowVectorPtr& data) {
+    VELOX_CHECK(
+        data->type()->equivalent(*rowType()),
+        "appended data type {} must match table type {}",
+        data->type(),
+        rowType());
+    VELOX_CHECK(data->size() > 0, "cannot append empty RowVector");
+    auto copy = std::dynamic_pointer_cast<RowVector>(
+        BaseVector::copy(*data, pool_.get()));
+    data_.push_back(copy);
   }
 
  private:
-  TestConnector* connector_;
-  std::vector<std::shared_ptr<const Column>> columns_;
-  std::unordered_map<std::string, const Column*> columnPtrMap_;
-  std::shared_ptr<const TableLayout> layout_;
-  std::vector<const TableLayout*> layoutPtrs_;
+  connector::Connector* connector_;
+  std::unordered_map<std::string, const Column*> columns_;
+  std::vector<std::unique_ptr<Column>> exportedColumns_;
+  std::vector<const TableLayout*> layouts_;
+  std::vector<std::unique_ptr<TableLayout>> exportedLayouts_;
+  std::shared_ptr<memory::MemoryPool> pool_;
+  std::vector<RowVectorPtr> data_;
+};
+
+/// SplitSource generated via the TestSplitManager embedded in the
+/// TestConnector. Generates one default-initialized ConnectorSplit
+/// for each partition provided at initialization time. targetBytes
+/// are ignored when retrieving splits, each new call to getSplits
+/// returns just one ConnectorSplit.
+class TestSplitSource : public SplitSource {
+ public:
+  TestSplitSource(
+      const std::string& connectorId,
+      const std::vector<PartitionHandlePtr>& partitions)
+      : connectorId_(connectorId),
+        partitions_(partitions),
+        currentPartition_(0) {}
+
+  std::vector<SplitAndGroup> getSplits(uint64_t targetBytes) override;
+
+ private:
+  const std::string connectorId_;
+  const std::vector<PartitionHandlePtr> partitions_;
+  size_t currentPartition_;
+};
+
+/// SplitManager embedded in the TestConnector. Returns one
+/// default-initialized PartitionHandle upon call to listPartitions.
+/// Generates a TestSplitSource containing the provided partition
+/// handles upon call to getSplitSource.
+class TestSplitManager : public ConnectorSplitManager {
+ public:
+  std::vector<PartitionHandlePtr> listPartitions(
+      const ConnectorTableHandlePtr& tableHandle) override;
+
+  std::shared_ptr<SplitSource> getSplitSource(
+      const ConnectorTableHandlePtr& tableHandle,
+      std::vector<PartitionHandlePtr> partitions,
+      SplitOptions options = {}) override;
 };
 
 class TestColumnHandle : public ColumnHandle {
  public:
-  explicit TestColumnHandle(const std::string& name) : name_{name} {}
+  TestColumnHandle(const std::string& name, const TypePtr& type)
+      : name_(name), type_(type) {}
 
   const std::string& name() const override {
+    return name_;
+  }
+
+  const TypePtr& type() const {
+    return type_;
+  }
+
+ private:
+  const std::string name_;
+  const TypePtr type_;
+};
+
+/// The layout corresponding to the handle is provided at
+/// initialization time.
+class TestTableHandle : public ConnectorTableHandle {
+ public:
+  TestTableHandle(
+      const TableLayout& layout,
+      std::vector<ColumnHandlePtr> columnHandles,
+      std::vector<core::TypedExprPtr> filters = {})
+      : ConnectorTableHandle(layout.connector()->connectorId()),
+        layout_(layout),
+        columnHandles_(std::move(columnHandles)),
+        filters_(std::move(filters)) {}
+
+  const std::string& name() const override {
+    return layout_.table()->name();
+  }
+
+  std::string toString() const override {
+    return name();
+  }
+
+  const TableLayout& layout() const {
+    return layout_;
+  }
+
+  const std::vector<core::TypedExprPtr>& filters() const {
+    return filters_;
+  }
+
+  const std::vector<ColumnHandlePtr>& columnHandles() const {
+    return columnHandles_;
+  }
+
+ private:
+  const TableLayout& layout_;
+  const std::vector<ColumnHandlePtr> columnHandles_;
+  const std::vector<core::TypedExprPtr> filters_;
+};
+
+/// The TestInsertTableHandle should be populated using the table
+/// name as the name parameter so that lookups can be performed
+/// against the ConnectorMetadata table map.
+class TestInsertTableHandle : public ConnectorInsertTableHandle {
+ public:
+  explicit TestInsertTableHandle(const std::string& name) : name_(name) {}
+
+  std::string toString() const override {
     return name_;
   }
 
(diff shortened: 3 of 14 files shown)


axiom/optimizer/connectors/tests/TestConnectorTest.cpp
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  void SetUp() override {
+    connector_ = std::make_shared<TestConnector>("connector");
+  }
+
+  std::shared_ptr<TestConnector> connector_;
+};
+
+TEST_F(TestConnectorTest, ConnectorRegister) {
```
mbasmanova, line 45: naming: ConnectorRegister -> connectorRegister

ditto other places

mbasmanova (commented):
@hdikeman What's the benefit of running a query that returns no rows?

I think it is valuable to add an in-memory connector that allows to query in-memory tables (`std::vector<RowVectorPtr>`).

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,137 +34,230 @@ class TestTableLayout : public TableLayout {
             table,
             connector,
             std::move(columns),
-            std::vector<const Column*>{},
-            std::vector<const Column*>{},
-            std::vector<SortOrder>{},
-            std::vector<const Column*>{},
-            true) {}
+            {}, // partitionColumns
```
mbasmanova, line 37: nit: I believe this is supposed to be /*partitionColumns*/ {}.

CC: @xiaoxmeng

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,137 +34,230 @@ class TestTableLayout : public TableLayout {
...
+            {}, // orderColumns
+            {}, // sortOrder
+            {}, // lookupKeys
+            true) {} // supportsScan
+
+  const std::vector<const Column*>& columns() const {
```
mbasmanova, line 43: Why do we need this API? Is TableLayout::columns() API not sufficient?

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,137 +34,230 @@ class TestTableLayout : public TableLayout {
...
-      RowTypePtr /* outputType */ = nullptr,
-      const std::vector<common::Subfield>& /* fields */ = {},
-      HashStringAllocator* /* allocator */ = nullptr,
-      std::vector<ColumnStatistics>* /* statistics */ = nullptr) const {
-    return std::make_pair(1'000, 1'000);
-  }
+      const connector::ConnectorTableHandlePtr& handle,
+      float pct,
+      std::vector<core::TypedExprPtr> extraFilters,
+      RowTypePtr outputType = nullptr,
```
mbasmanova, line 51: I believe we do not need default values for the override, do we?

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,137 +34,230 @@ class TestTableLayout : public TableLayout {
...
  private:
-  TestConnector* connector_;
-  std::vector<std::shared_ptr<const Column>> columns_;
-  std::unordered_map<std::string, const Column*> columnPtrMap_;
-  std::shared_ptr<const TableLayout> layout_;
-  std::vector<const TableLayout*> layoutPtrs_;
+  std::vector<const Column*> getColumnVector() const;
+
+  connector::Connector* connector_;
+  uint64_t rows_;
```
mbasmanova, line 82: const ?

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,137 +34,230 @@ class TestTableLayout : public TableLayout {
...
+  std::unordered_map<std::string, const Column*> columns_;
+  std::vector<std::unique_ptr<Column>> exportedColumns_;
+  std::vector<const TableLayout*> layouts_;
+  std::vector<std::unique_ptr<TableLayout>> exportedLayouts_;
+};
+
+class TestSplit : public ConnectorSplit {
```
mbasmanova, line 89: Why do we need this class?

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,137 +34,230 @@ class TestTableLayout : public TableLayout {
...
+      const std::string& connectorId,
+      const std::vector<std::shared_ptr<const PartitionHandle>>& partitions)
+      : connectorId_(connectorId),
+        partitions_(partitions),
+        currentPartition_(0) {}
+
+  std::vector<SplitAndGroup> getSplits(uint64_t targetBytes) override;
+
+ private:
+  std::string connectorId_;
```
mbasmanova, line 107: const

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,137 +34,230 @@ class TestTableLayout : public TableLayout {
...
+  std::shared_ptr<SplitSource> getSplitSource(
+      const ConnectorTableHandlePtr& tableHandle,
+      std::vector<std::shared_ptr<const PartitionHandle>> partitions,
+      SplitOptions options = {}) override;
 };
 
 class TestColumnHandle : public ColumnHandle {
  public:
-  explicit TestColumnHandle(const std::string& name) : name_{name} {}
+  TestColumnHandle(const std::string& name, TypePtr type)
```
mbasmanova, line 125: consistency: either pass both by name and move or pass both by reference and copy

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,137 +34,230 @@ class TestTableLayout : public TableLayout {
...
     return name_;
   }
 
+  const TypePtr& type() const {
+    return type_;
+  }
+
  private:
-  const std::string name_;
+  std::string name_;
```
mbasmanova, line 137: const

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,137 +34,230 @@ class TestTableLayout : public TableLayout {
...
       std::vector<core::TypedExprPtr>& rejectedFilters,
-      RowTypePtr dataColumns,
-      std::optional<LookupKeys>) override;
+      RowTypePtr dataColumns = nullptr,
+      std::optional<LookupKeys> = std::nullopt) override;
+
+  void addTable(
+      const std::string& name,
+      const RowTypePtr& schema,
+      TestConnector* connector,
```
mbasmanova, line 211: Why do we need this parameter?

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,137 +34,230 @@ class TestTableLayout : public TableLayout {
...
       std::vector<ColumnHandlePtr> columnHandles,
       core::ExpressionEvaluator& evaluator,
       std::vector<core::TypedExprPtr> filters,
       std::vector<core::TypedExprPtr>& rejectedFilters,
-      RowTypePtr dataColumns,
-      std::optional<LookupKeys>) override;
+      RowTypePtr dataColumns = nullptr,
+      std::optional<LookupKeys> = std::nullopt) override;
+
+  void addTable(
```
mbasmanova, line 208: Would you document this API?

axiom/optimizer/connectors/tests/TestConnectorTest.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+#include <folly/init/Init.h>
+#include <gtest/gtest.h>
+
+#include "velox/common/base/tests/GTestUtils.h"
+#include "velox/expression/Expr.h"
+#include "velox/type/Type.h"
+
+using namespace facebook::velox;
+using namespace facebook::velox::connector;
```
mbasmanova, line 27: not needed

hdikeman:
>add automation
>@hdikeman What does this refer to ?

TestConnectorTest.cpp

>Can we decouple https://github.com/facebookexperimental/verax/pull/132 from this PR? Is there a dependency?

Yes there is a dependency, these new classes are used during the sampling query execution. I could have turned it off but chose not to. And would like to have it anyways for lightweight optimizer testing I thought. LocalHive is a bit heavy duty for most purposes

>TestConnector is not a framework, is it?

Maybe wrong word. The right one is on the tip of my tongue

axiom/optimizer/connectors/tests/TestConnector.h
hdikeman: I think I didn't always embed the connector in the ConnectorMetadata, you're right I can get it from the class field now. Thanks

axiom/optimizer/connectors/tests/TestConnector.h
hdikeman: We could get away with a vanilla ConnectorSplit for now

axiom/optimizer/connectors/tests/TestConnector.h
hdikeman: CPP reference says this:

>The overriders of [virtual](https://en.cppreference.com/w/cpp/language/virtual.html) functions do not acquire the default arguments from the base class declarations, and when the virtual function call is made, the default arguments are decided based on the static type of the object (note: this can be avoided with [non-virtual interface](http://www.gotw.ca/publications/mill18.htm) pattern).

I think we do

mbasmanova:
> Yes there is a dependency, these new classes are used during the sampling query execution. 

Got it. Would you update PR description to explain this? I keep forgetting that optimizer also executes the query (for sampling).

axiom/optimizer/connectors/tests/CMakeLists.txt
```diff
@@ -16,3 +16,21 @@ velox_add_library(velox_test_connector TestConnector.cpp)
 
 velox_link_libraries(velox_test_connector velox_connector_metadata
                      velox_common_base velox_memory velox_connector)
+
+add_executable(
+  velox_test_connector_test
```
mbasmanova, line 21: These names are in a flat namespace. Hence, we need some way to avoid collisions. I suggest to use folder path: velox_optimizer_connectors_tests_test_connector. If that's too long we may shorten it by replacing velox_optimizer_ with vo_.

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -22,6 +22,9 @@ namespace facebook::velox::connector {
 
 class TestConnector;
 
+/// TableLayout implementation for the TestConnector connector APIs. The
```
mbasmanova, line 25: > TableLayout implementation for the TestConnector connector APIs.

This is redundant. Let's drop.

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,256 @@ class TestTableLayout : public TableLayout {
...
+      const connector::ConnectorTableHandlePtr& handle,
+      float pct,
+      std::vector<core::TypedExprPtr> extraFilters,
+      RowTypePtr outputType = nullptr,
+      const std::vector<common::Subfield>& fields = {},
+      HashStringAllocator* allocator = nullptr,
+      std::vector<ColumnStatistics>* statistics = nullptr) const override;
 };
 
+/// Table implementation for the TestConnector connector APIs. Row count
```
mbasmanova, line 56: ditto

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,256 @@ class TestTableLayout : public TableLayout {
...
+class TestSplitSource : public SplitSource {
+ public:
+  TestSplitSource(
+      const std::string& connectorId,
+      const std::vector<std::shared_ptr<const PartitionHandle>>& partitions)
+      : connectorId_(connectorId),
+        partitions_(partitions),
+        currentPartition_(0) {}
+
+  std::vector<SplitAndGroup> getSplits(uint64_t targetBytes) override;
```
mbasmanova, line 102: I assume 'targetBytes' is ignored. Let's document that.

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,256 @@ class TestTableLayout : public TableLayout {
...
+  size_t currentPartition_;
+};
+
+/// SplitManager embedded in the TestConnector. Returns one
+/// default-initialized PartitionHandle upon call to listPartitions.
+/// Generates a TestSplitSource containing the provided partition
+/// handles upon call to getSplitSource.
+class TestSplitManager : public ConnectorSplitManager {
+ public:
+  std::vector<std::shared_ptr<const PartitionHandle>> listPartitions(
```
mbasmanova, line 116: Let's add PartitionHandlePtr typedef.

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,256 @@ class TestTableLayout : public TableLayout {
...
+  const TypePtr& type() const {
+    return type_;
+  }
+
  private:
   const std::string name_;
+  const TypePtr type_;
 };
 
+/// TableHandle implementation for the TestConnector. The layout
```
mbasmanova, line 145: > TableHandle implementation for the TestConnector.

Redundant. Drop. Ditto other places.

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,256 @@ class TestTableLayout : public TableLayout {
...
+/// corresponding to the handle as well as the set of pushed-down
+/// filters are provided at initialization time.
 class TestTableHandle : public ConnectorTableHandle {
  public:
-  explicit TestTableHandle(const std::string& connectorId)
-      : ConnectorTableHandle{connectorId} {}
+  TestTableHandle(
+      const TableLayout& layout,
+      std::vector<ColumnHandlePtr> columnHandles,
+      std::vector<core::TypedExprPtr> filters = {})
```
mbasmanova, line 153: Does TestConnector now support filter pushdown? It might be useful to control this behavior with a flag.

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,256 @@ class TestTableLayout : public TableLayout {
...
+    return filters_;
   }
+
+  const std::vector<ColumnHandlePtr>& columnHandles() const {
+    return columnHandles_;
+  }
+
+ private:
+  const TableLayout& layout_;
+  std::vector<ColumnHandlePtr> columnHandles_;
```
mbasmanova, line 181: const; ditto other places

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,256 @@ class TestTableLayout : public TableLayout {
...
+  std::vector<core::TypedExprPtr> filters_;
 };
 
+/// ConnectorMetadata for the TestConnector. Contains an in-memory
+/// map of TestTables which are inserted at runtime via the addTable
+/// API. Tables are retrieved by name using the findTable API. The
+/// splitManager API returns a TestSplitManager. createColumnHandle
+/// returns a TestColumnHandle for the specified layout and column.
+/// createTableHandle returns a TestTableHandle for the specified
+/// layout. If the backing Connector has filter pushdown enabled,
```
mbasmanova, line 191: > for the specified layout

I assume each table has only one layout. Let's clarify.

> If the backing Connector

What is "backing Connector" here?

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,256 @@ class TestTableLayout : public TableLayout {
...
       core::ExpressionEvaluator& evaluator,
       std::vector<core::TypedExprPtr> filters,
       std::vector<core::TypedExprPtr>& rejectedFilters,
-      RowTypePtr dataColumns,
-      std::optional<LookupKeys>) override;
+      RowTypePtr dataColumns = nullptr,
+      std::optional<LookupKeys> = std::nullopt) override;
+
+  /// Add a TestTable with the specified name, schema, and row count to the
+  /// in-memory map maintained in the connector metadata. If a table with
```
mbasmanova, line 225: > If a table with the specified name already exists, it is replaced.

Let's return a bool to indicate whether an insert or replace has occurred.

mbasmanova (commented):
@hdikeman I'm concerned that this implementation violates basic contracts. E.g. writing N rows to table T, then reading count(*) from T produces 0, not N.

It seems like this implementation only works for sampling... wondering if it is better to disable sampling and work towards more functional in-memory connector implementation.

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -177,32 +297,70 @@ class TestConnector : public Connector {
...
+      RowTypePtr inputType,
+      ConnectorInsertTableHandlePtr connectorInsertTableHandle,
+      ConnectorQueryCtx* connectorQueryCtx,
+      CommitStrategy commitStrategy) override;
+
+  /// Add a TestTable with the specified name, schema, and row count
+  /// to the TestConnectorMetadata corresponding to this connector.
+  void addTable(
+      const std::string& name,
+      const RowTypePtr& schema = ROW({}, {}),
```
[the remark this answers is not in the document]
hdikeman: Yes we do

e.g. see SchemaResolverTest. In that case I don't care about schema or row count. Sometimes the caller just wants a table which will be found during `metadata->findTable()`

hdikeman:
> @hdikeman I'm concerned that this implementation violates basic contracts. E.g. writing N rows to table T, then reading count(*) from T produces 0, not N.

@mbasmanova I wouldn't call that a contract violation, those are two separate queries and the Connector interface doesn't provide any guarantees about what happens to your data after write. Any expectations about read-after-write semantics are implementation-specific. It's fairly common and straightforward to have a storage test class which returns known data on read and dumps data on write

> It seems like this implementation only works for sampling... wondering if it is better to disable sampling and work towards more functional in-memory connector implementation.

I'm not sure I follow the logic. Is there a meaningful difference between the execution performed during sample() and the one during actual query execution

axiom/optimizer/connectors/tests/TestConnector.h
hdikeman: Sort of but these filters are not applied at the DataSource since the DataSource does not return any data aside from some placeholder data

I had a flag but removed given the later comments about controlling with config. Can just defer that and reject all filters for now

axiom/optimizer/connectors/tests/TestConnector.h
hdikeman: I had the semantic wrong when I added the docs. It won't replace the table if there's a name collision. I'll throw an error

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -177,32 +297,70 @@ class TestConnector : public Connector {
...
+  bool canAddDynamicFilter() const override {
+    return false;
   }
 
-  void addTable(const std::string& name, const RowTypePtr& schema) {
-    metadata_->addTable(name, schema);
+  /// Enable filter pushdown for the connector. If enabled, filters
+  /// are stashed in the TestTableHandle. Otherwise, filters are
+  /// rejected via the rejectedFilters parameter.
+  void setPushdownFilters(bool enabled) {
```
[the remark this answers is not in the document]
mbasmanova: > always reject filters for now

Sounds good.

mbasmanova:
>  Is there a meaningful difference between the execution performed during sample() and the one during actual query execution

The difference is sampling doesn't need to produce accurate results, while actual execution does.

>  wouldn't call that a contract violation, those are two separate queries and the Connector interface doesn't provide any guarantees about what happens to your data after write.

I find it hard to agree with this. It is nearly impossible to reason about a query engine that drops writes and always returns empty dataset on read.

hdikeman:
>TestConnector is not a framework, is it?

fixture! Ah this was driving me crazy, geez

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+}
+
+TestTable::TestTable(
+    const std::string& name,
+    const RowTypePtr& schema,
+    TestConnector* connector)
+    : Table(name), connector_(connector) {
+  type_ = schema;
+  for (auto i = 0; i < schema->size(); ++i) {
+    auto columnName = schema->nameOf(i);
```
mbasmanova, line 38: const &

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+    auto columnType = schema->childAt(i);
+    auto column = std::make_unique<Column>(columnName, columnType);
+    columns_[columnName] = column.get();
+    exportedColumns_.push_back(std::move(column));
```
mbasmanova, line 42: nit: a common pattern is something like

```
exportedColumns_.emplace_back(std::make_unique<Column>(columnName, columnType));
columns_[columnName] = exportedColumns_.back().get();
```

Please, reserve exportedColumns_ before the loop.

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+    const std::string& name,
+    const RowTypePtr& schema,
+    TestConnector* connector)
+    : Table(name), connector_(connector) {
+  type_ = schema;
+  for (auto i = 0; i < schema->size(); ++i) {
+    auto columnName = schema->nameOf(i);
+    auto columnType = schema->childAt(i);
+    auto column = std::make_unique<Column>(columnName, columnType);
+    columns_[columnName] = column.get();
```
mbasmanova, line 41: Add a check to make sure all names are unique and non-empty.

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+    exportedColumns_.push_back(std::move(column));
+  }
+  auto layout = std::make_unique<TestTableLayout>(
+      name_, this, connector_, getColumnVector());
+  layouts_.push_back(layout.get());
+  exportedLayouts_.push_back(std::move(layout));
+  pool_ = memory::memoryManager()->addLeafPool(name + "_table");
```
mbasmanova, line 48: Let's use the pool provided by Velox as the parent pool.

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+}
+
+std::vector<const Column*> TestTable::getColumnVector() const {
```
mbasmanova, line 51: This method is used only once in the ctor. Let't remove it and combine its logic with the existing loop.

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+
+std::vector<SplitSource::SplitAndGroup> TestSplitSource::getSplits(uint64_t) {
+  std::vector<SplitAndGroup> result;
+  if (currentPartition_ >= partitions_.size()) {
+    result.push_back({nullptr, kUngroupedGroupId});
+  } else {
+    result.push_back(
+        {std::make_shared<ConnectorSplit>(connectorId_), kUngroupedGroupId});
+  }
+  currentPartition_++;
```
mbasmanova, line 68: It looks like we can simply return a vector of 2 entries: a split + end marker. No need to maintain any state like currentPartition_.

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+std::vector<PartitionHandlePtr> TestSplitManager::listPartitions(
+    const ConnectorTableHandlePtr&) {
+  return {std::make_shared<PartitionHandle>()};
+}
+
+std::shared_ptr<SplitSource> TestSplitManager::getSplitSource(
+    const ConnectorTableHandlePtr& tableHandle,
+    std::vector<PartitionHandlePtr> partitions,
+    SplitOptions) {
+  return std::make_shared<TestSplitSource>(
```
mbasmanova, line 81: assert there is only one partition

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+
+ColumnHandlePtr TestConnectorMetadata::createColumnHandle(
+    const TableLayout& layout,
+    const std::string& columnName,
+    std::vector<common::Subfield>,
+    std::optional<TypePtr> castToType,
+    SubfieldMapping) {
+  auto column = layout.findColumn(columnName);
+  VELOX_CHECK_NOT_NULL(
+      column, "Column {} not found in layout {}", columnName, layout.name());
```
mbasmanova, line 104: layout -> table (for clarity)

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+  return std::make_shared<TestColumnHandle>(
+      columnName, castToType.value_or(column->type()));
```
mbasmanova, line 106: > castToType

Is there logic to actually implement this cast when reading the data?

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
   rejectedFilters = std::move(filters);
-  return std::make_shared<TestTableHandle>(connector_->connectorId());
+  return std::make_shared<TestTableHandle>(layout, std::move(columnHandles));
+}
+
+void TestConnectorMetadata::addTable(
+    const std::string& name,
+    const RowTypePtr& schema) {
+  auto table = std::make_unique<TestTable>(name, schema, connector_);
+  auto [_, emplaced] = tables_.emplace(name, std::move(table));
```
mbasmanova, line 125: a conventional pattern is

```
bool ok = tables_.emplace(name, std::move(table)).second;
VELOX_CHECK(ok, "table {} already exists", name);
```

It is shorter / more readable.

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+  VELOX_CHECK(emplaced, "table {} already exists", name);
+}
+
+void TestConnectorMetadata::addTableData(
+    const std::string& name,
+    const RowVectorPtr& data) {
+  auto it = tables_.find(name);
+  VELOX_CHECK(it != tables_.end(), "no table {} exists");
```
mbasmanova, line 133: Make sure to put runtime info at the end of the message: "Table not found: {}".

This allows for easier debugging as one can grep for the error message in the code base.

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+  it->second->addData(data);
+}
+
+TestDataSource::TestDataSource(
+    const RowTypePtr& outputType,
+    const ConnectorTableHandlePtr& tableHandle,
+    const Table* table)
+    : outputType_(outputType), tableHandle_(tableHandle) {
+  data_ = static_cast<const TestTable*>(table)->data();
```
mbasmanova, line 142: dynamic_cast + check for null

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+}
+
+void TestDataSource::addSplit(std::shared_ptr<ConnectorSplit> split) {
+  split_ = std::move(split);
```
mbasmanova, line 146: Add checks to make sure there is only one split.

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+}
+
+std::optional<RowVectorPtr> TestDataSource::next(
+    uint64_t,
+    velox::ContinueFuture&) {
+  VELOX_CHECK(split_, "no split added to DataSource");
+  if (data_.size() <= idx_) {
+    return nullptr;
+  }
+  return data_[idx_++];
```
mbasmanova, line 156: Don't we need to handle castToType and column pruning?

Let's add a check to the caller of next() that returned type matches expected.

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+
+void TestConnector::addTableData(
+    const std::string& name,
+    const RowVectorPtr& data) {
+  metadata_->addTableData(name, data);
+}
+
+void TestDataSink::appendData(RowVectorPtr vector) {
+  if (vector) {
+    static_cast<TestTable*>(table_)->addData(vector);
```
mbasmanova, line 204: Modify the type of table_ to TestTable.

axiom/optimizer/connectors/tests/TestConnectorTest.cpp
```diff
@@ -0,0 +1,241 @@
...
+    connector_ = std::make_shared<TestConnector>("connector");
+  }
+
+  std::shared_ptr<TestConnector> connector_;
+};
+
+TEST_F(TestConnectorTest, connectorRegister) {
+  registerConnector(connector_);
+
+  auto connector = getConnector("connector");
```
mbasmanova, line 46: > "connector"

this name is confusing; perhaps, change to something like "test" or "foo". Let's also add a temp variable to avoid copy-pasting.

axiom/optimizer/connectors/tests/TestConnectorTest.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+  void SetUp() override {
+    connector_ = std::make_shared<TestConnector>("connector");
+  }
+
+  std::shared_ptr<TestConnector> connector_;
+};
+
+TEST_F(TestConnectorTest, connectorRegister) {
+  registerConnector(connector_);
```
mbasmanova, line 44: perhaps, first call getConnector and verify it is not there.

axiom/optimizer/connectors/tests/TestConnectorTest.cpp
```diff
@@ -0,0 +1,241 @@
...
+  unregisterConnector("connector");
+  VELOX_ASSERT_THROW(
+      getConnector("connector"),
+      "Connector with ID 'connector' not registered");
+}
+
+TEST_F(TestConnectorTest, table) {
+  auto metadata = connector_->metadata();
+
+  auto schema = ROW({{"col1", INTEGER()}, {"col2", VARCHAR()}});
```
mbasmanova, line 60: avoid a1, a2, a3... names

perhaps, a, b

axiom/optimizer/connectors/tests/TestConnectorTest.cpp
```diff
@@ -0,0 +1,241 @@
...
+  connector_->addTable("table", schema);
+  auto table = metadata->findTable("table");
+  EXPECT_NE(table, nullptr);
+  EXPECT_EQ(table->name(), "table");
+  EXPECT_EQ(table->numRows(), 0);
+  EXPECT_EQ(table->columnMap().size(), 2);
+  EXPECT_NE(table->columnMap().find("col1"), table->columnMap().end());
```
mbasmanova, line 67: .contains

axiom/optimizer/connectors/tests/TestConnectorTest.cpp
```diff
@@ -0,0 +1,241 @@
...
+  auto splitManager = metadata->splitManager();
+  EXPECT_NE(splitManager, nullptr);
+}
+
+TEST_F(TestConnectorTest, splits) {
+  auto split = std::make_shared<ConnectorSplit>("connector");
+  EXPECT_EQ(split->connectorId, "connector");
+}
+
+TEST_F(TestConnectorTest, dataSink) {
```
mbasmanova, line 123: It might be helpful to change this test to use Velox's TableWriter. This will make sure all is wired up propertly. E.g. create Velox plan with values followed by table write. Run the plan. Read the data back and verify the results.

axiom/optimizer/connectors/tests/TestConnectorTest.cpp
```diff
@@ -0,0 +1,241 @@
...
+  EXPECT_EQ(table->numRows(), 5);
+
+  dataSink->appendData(nullptr);
+  EXPECT_EQ(table->numRows(), 5);
+
+  EXPECT_TRUE(dataSink->finish());
+  EXPECT_TRUE(dataSink->close().empty());
+}
+
+TEST_F(TestConnectorTest, dataSource) {
```
mbasmanova, line 150: let's test this by executing Velox plan to make sure things work end-to-end

mbasmanova (commented):
@hdikeman Thank you for extending TestConnector. Some comments.

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,290 @@ class TestTableLayout : public TableLayout {
...
+  std::vector<std::unique_ptr<Column>> exportedColumns_;
+  std::vector<const TableLayout*> layouts_;
+  std::vector<std::unique_ptr<TableLayout>> exportedLayouts_;
+  std::shared_ptr<memory::MemoryPool> pool_;
+  std::vector<RowVectorPtr> data_;
+};
+
+/// SplitSource generated via the TestSplitManager embedded in the
+/// TestConnector. Generates one default-initialized ConnectorSplit
+/// for each partition provided at initialization time. targetBytes
```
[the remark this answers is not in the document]
hdikeman: There are some details in the comments for SplitManager, but basically we just produce one partition always, and that one partition always produces one split which contains the data

To make that controllable we could do something fancy like add a partition parameter to the data append function or alternatively inject some configuration to the table to specify the number of partitions and splits we want plan execution to use. But it felt like jumping ahead to take a crack at that use-case

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,290 @@ class TestTableLayout : public TableLayout {
...
+    return name_;
+  }
+
+ private:
+  const std::string name_;
+};
+
+/// Contains an in-memory map of TestTables inserted via the addTable
+/// API. Tables are retrieved by name using the findTable API. The
+/// splitManager API returns a TestSplitManager. createColumnHandle
```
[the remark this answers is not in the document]
hdikeman: There is no elaborate splitting at the moment. One partition, one split

There is a single layout yes, it's populated with the schema specified during table init

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,290 @@ class TestTableLayout : public TableLayout {
...
+      const RowTypePtr& outputType,
+      const ConnectorTableHandlePtr& tableHandle,
+      const Table* table);
+
+  void addSplit(std::shared_ptr<ConnectorSplit> split) override;
+
+  std::optional<RowVectorPtr> next(uint64_t size, velox::ContinueFuture& future)
+      override;
+
+  void addDynamicFilter(
```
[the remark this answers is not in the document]
hdikeman: Perhaps NYI is more appropriate? I'll add that

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -177,32 +331,86 @@ class TestConnector : public Connector {
...
+      RowTypePtr inputType,
+      ConnectorInsertTableHandlePtr connectorInsertTableHandle,
+      ConnectorQueryCtx* connectorQueryCtx,
+      CommitStrategy commitStrategy) override;
+
+  /// Add a TestTable with the specified name, schema, and row count
+  /// to the TestConnectorMetadata corresponding to this connector.
+  void addTable(
+      const std::string& name,
+      const RowTypePtr& schema = ROW({}, {}));
```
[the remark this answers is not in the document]
hdikeman: That one actually gets used
I think I referenced in another comment during a previous review round. For some tests which only care about table lookup (e.g. SchemaResolverTest) and not schema, it's convenient to fill in a dummy schema

And this isn't one of those cases where the parent defaults take over, since not a virtual method

axiom/optimizer/connectors/tests/TestConnector.cpp
hdikeman: What pool provided by Velox?

axiom/optimizer/connectors/tests/TestConnector.cpp
hdikeman: I'd rather keep it to make life easy if or when someone decides to add support for more complex partition/split generation logic

axiom/optimizer/connectors/tests/TestConnector.cpp
hdikeman: Not sure we want to, we can handle >1

axiom/optimizer/connectors/tests/TestConnector.cpp
hdikeman: The caller of next() is plan execution

We ignore castToType for sure. Maybe should handle column pruning

axiom/optimizer/connectors/tests/TestConnector.h
```diff
@@ -34,139 +37,290 @@ class TestTableLayout : public TableLayout {
...
-    return it->second.get();
-  }
+  const Table* FOLLY_NULLABLE findTable(const std::string& name) override;
 
-  ConnectorSplitManager* splitManager() override {
-    VELOX_NYI();
-  }
+  /// Non-interface method which supplies a non-const Table reference
+  /// which is capable of performing writes to the underlying table.
+  Table* FOLLY_NULLABLE findTableInternal(const std::string& name);
```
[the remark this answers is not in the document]
hdikeman: Hmm don't think so. What if the table doesn't exist?

This method is also used internally by findTable

axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -9,25 +9,200 @@
...
+    std::vector<ColumnStatistics>*) const {
+  return std::make_pair(1'000, 1'000);
+}
+
+TestTable::TestTable(
+    const std::string& name,
+    const RowTypePtr& schema,
+    TestConnector* connector)
+    : Table(name), connector_(connector) {
+  type_ = schema;
```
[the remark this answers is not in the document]
hdikeman: Not sure I can since it's a parent attribute

I'll make it more explicit what I'm doing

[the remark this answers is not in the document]
mbasmanova: >  since it's a parent attribute

That's an anti pattern. Let's fix that in a separate PR, then rebase.

axiom/optimizer/connectors/tests/TestConnector.cpp
hdikeman: As mentioned previously we can handle >1 partition and >1 split so may leave the handling as-is, so any future change just needs to update the partition/split generation

axiom/optimizer/connectors/tests/TestConnectorTest.cpp
hdikeman: Hmm what is the reasoning behind not using "col1" etc. etc.

using "a" "b" etc. I would see as more unreadable since there's nothing obvious telling you it's the name of a column, so makes debugging harder

For now I'll update it though

axiom/optimizer/connectors/tests/TestConnector.cpp
hdikeman: We discussed offline, I think in this case since we want table data to persist we can't use a pool provided by the existing interfaces, which typically only live for the lifetime of a query

axiom/optimizer/tests/TestConnectorQueryTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+        write,
+        fragment.executionStrategy,
+        fragment.numSplitGroups,
+        fragment.groupedExecutionLeafNodeIds);
+    std::vector<runner::ExecutableFragment> fragments = {writeFragment};
+
+    return std::make_shared<runner::MultiFragmentPlan>(fragments, options_);
+  }
+
+  TestResult executePlan(const lp::LogicalPlanNodePtr& logicalPlan) {
```
mbasmanova, line 98: Why not use runVelox from the base class?

axiom/optimizer/tests/TestConnectorQueryTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+  std::unique_ptr<HashStringAllocator> allocator_;
+  std::unique_ptr<QueryGraphContext> context_;
+  runner::MultiFragmentPlan::Options options_;
+};
+
+TEST_F(TestConnectorQueryTest, selectFiltered) {
+  auto vector = makeRowVector({makeFlatVector<int64_t>({0, 1, 2})});
+  auto schema = ROW({"a"}, {BIGINT()});
+
+  connector_->createTable("readtable", schema);
```
mbasmanova, line 121: conventional names for test tables are "t", "u"

axiom/optimizer/tests/TestConnectorQueryTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+  auto expected = makeRowVector({makeFlatVector<int64_t>({1, 2})});
+  executePlanChecked(plan, expected);
+}
+
+TEST_F(TestConnectorQueryTest, writeFiltered) {
+  auto vector = makeRowVector(
+      {"b", "c"},
+      {makeFlatVector<int64_t>({0, 1, 2}),
+       makeFlatVector<StringView>({"str", "ing", "val"})});
+  auto schema = ROW({"b", "c"}, {BIGINT(), VARCHAR()});
```
mbasmanova, line 136: auto schema = vector->rowType()

axiom/optimizer/tests/TestConnectorQueryTest.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  auto table = dynamic_cast<connector::TestTable*>(
+      connector_->createTable("writetable", schema));
```
mbasmanova, line 139: Can we change createTable API to return TestTable* ?

mbasmanova (approved):
Thanks.

mbasmanova (approved):

PR #132, opened 2025-08-07
fix: unbreak logical plan optimization for multi-connector queries

hdikeman:
Summary:
For queries which reference tables residing on multiple connectors, we need a way to specify to the optimizer which connector to query for table metadata

Previously, with the physical input interfaces, we passed a fully-qualified table name (`[catalog].[schema].[query]`). Now there is a dedicated field in logical scan nodes for the connector ID

Modifying the schema lookup in Verax to respect this separately specified catalog by passing it explicitly. If the plan creator created the scan with a fully-qualified table name as before, the catalog specified there is compared against the separate connector ID parameter

Differential Revision: D79820585

### DIFF
--- a/axiom/optimizer/Schema.h
+++ b/axiom/optimizer/Schema.h
@@ -449,8 +449,11 @@ class Schema {
   /// Constructs a Schema for producing executable plans, backed by 'source'.
   Schema(Name _name, SchemaResolver* source, LocusCP locus);
 
-  /// Returns the table with 'name' or nullptr if not found.
-  SchemaTableCP findTable(std::string_view name) const;
+  /// Returns the table with 'name' or nullptr if not found, using
+  /// the connector specified by connectorId to perform table lookups.
+  /// An error is thrown if no connector with the specified ID exists.
+  SchemaTableCP findTable(std::string_view connectorId, std::string_view name)
+      const;
 
   Name name() const {
     return name_;
--- a/axiom/optimizer/SchemaResolver.cpp
+++ b/axiom/optimizer/SchemaResolver.cpp
@@ -16,19 +16,24 @@
 
 #include "axiom/optimizer/SchemaResolver.h"
 #include "axiom/optimizer/SchemaUtils.h"
+#include "velox/connectors/Connector.h"
 
 namespace facebook::velox::optimizer {
 
-const connector::Table* SchemaResolver::findTable(const std::string& name) {
+const connector::Table* SchemaResolver::findTable(
+    const std::string& catalog,
+    const std::string& name) {
   TableNameParser parser(name);
   VELOX_USER_CHECK(parser.valid(), "Invalid table name: '{}'", name);
 
-  connector::Connector* connector = parser.catalog().has_value()
-      ? connector::getConnector(parser.catalog().value()).get()
-      : defaultConnector_.get();
-
-  VELOX_USER_CHECK_NOT_NULL(
-      connector, "Connector not found for table: {}", name);
+  if (parser.catalog().has_value()) {
+    VELOX_USER_CHECK_EQ(
+        catalog,
+        parser.catalog().value(),
+        "Input catalog must match table catalog specifier");
+  }
+  connector::Connector* connector =
+      velox::connector::getConnector(catalog).get();
 
   std::string lookupName;
   if (parser.schema().has_value()) {
@@ -38,7 +43,6 @@ const connector::Table* SchemaResolver::findTable(const std::string& name) {
   } else {
     lookupName = parser.table();
   }
-
   return connector->metadata()->findTable(lookupName);
 }
 
--- a/axiom/optimizer/SchemaResolver.h
+++ b/axiom/optimizer/SchemaResolver.h
@@ -24,25 +24,25 @@ namespace facebook::velox::optimizer {
 
 class SchemaResolver {
  public:
-  SchemaResolver(
-      const std::shared_ptr<connector::Connector>& defaultConnector,
-      const std::string& defaultSchema)
-      : defaultConnector_(defaultConnector), defaultSchema_(defaultSchema) {}
+  SchemaResolver(const std::string& defaultSchema = "")
+      : defaultSchema_(defaultSchema) {}
 
   virtual ~SchemaResolver() = default;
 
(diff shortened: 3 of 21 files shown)


axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -236,6 +237,36 @@ TEST_F(PlanTest, rejectedFilters) {
   ASSERT_TRUE(matcher->match(plan));
 }
 
+TEST_F(PlanTest, multipleConnectors) {
+  auto extraConnector = std::make_shared<connector::TestConnector>("extra");
+  connector::registerConnector(extraConnector);
+  auto guard =
```
mbasmanova, line 243: perhaps, SCOPE_EXIT

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -236,6 +237,36 @@ TEST_F(PlanTest, rejectedFilters) {
...
+      folly::makeGuard([&] { connector::unregisterConnector("extra"); });
+
+  testConnector_->addTable("table1", ROW({"a"}, {BIGINT()}));
+  extraConnector->addTable("table2", ROW({"b"}, {BIGINT()}));
+
+  lp::PlanBuilder::Context context;
+  auto logicalPlan =
+      lp::PlanBuilder(context)
+          .tableScan(kTestConnectorId, "table1", {"a"})
```
mbasmanova, line 252: nit: this can be simplified to

```
lp::PlanBuilder::Context context(kTestConnectorId);

...tableScan("table1")

...tableScan("extra", "table2")
```

axiom/optimizer/Schema.h
```diff
@@ -452,7 +452,9 @@ class Schema {
   Schema(Name _name, SchemaResolver* source, LocusCP locus);
 
   /// Returns the table with 'name' or nullptr if not found.
-  SchemaTableCP findTable(std::string_view name) const;
+  SchemaTableCP findTable(
+      const std::optional<std::string>& connectorId,
```
mbasmanova, line 456: Would you update the comment to explain what 'connectorId' is for and what happens if it is not specified?

It might be more convenient to provide 2 separate APIs. One that takes just table name and another that takes both connector ID and table name.

axiom/optimizer/SchemaResolver.h
```diff
@@ -32,16 +32,18 @@ class SchemaResolver {
...
+  // exist. If a connector for the specified catalog doesn't exist, an error
+  // will be returned. Input table name can be any of the following formats:
   //   - "tablename"
   //   - "schema.tablename"
   //   - "catalog.schema.tablename"
   // If catalog is omitted, defaultConnector will be used for the table lookup.
   // If schema is omitted, defaultSchema will be prepended prior to lookup.
-  virtual const connector::Table* findTable(const std::string& name);
+  virtual const connector::Table* findTable(
+      const std::optional<std::string>& catalog,
```
mbasmanova, line 43: It is not clear what's the purpose of this new argument. 

Is catalog == connector ID? Let's clarify.

axiom/optimizer/SchemaResolver.cpp
```diff
@@ -16,19 +16,31 @@
...
+    const std::string& name) {
   TableNameParser parser(name);
-  VELOX_USER_CHECK(parser.valid(), "Invalid table name: '{}'", name);
+  VELOX_CHECK(parser.valid(), "Invalid table name: '{}'", name);
 
-  connector::Connector* connector = parser.catalog().has_value()
-      ? connector::getConnector(parser.catalog().value()).get()
-      : defaultConnector_.get();
+  if (catalog.has_value() && parser.catalog().has_value()) {
+    VELOX_CHECK_EQ(
```
mbasmanova, line 30: VELOX_USER_...

axiom/optimizer/tests/SchemaResolverTest.cpp
```diff
@@ -100,17 +109,27 @@ TEST_F(SchemaResolverTest, tablePlusSchemaPlusCatalog) {
...
-  table = resolver_->findTable(lookup);
+  table = resolver_->findTable(std::nullopt, lookup);
   ASSERT_NE(table, nullptr);
   ASSERT_EQ(table->name(), expect);
 }
 
+TEST_F(SchemaResolverTest, catalogMismatch) {
+  auto lookupName = "other.otherschema.table";
+  otherCatalog_.connector->addTable(lookupName);
+  EXPECT_THROW(resolver_->findTable("base", lookupName), VeloxException);
```
mbasmanova, line 127: VELOX_ASSERT_THROW

mbasmanova (commented):
@hdikeman  Thank you for fixing this.

axiom/optimizer/Schema.h
hdikeman: Considered two APIs, but one would go unused as soon as the physical plan input for Verax is removed

I'll update the comment

mbasmanova: > as soon as the physical plan input for Verax is removed

#135

hdikeman: Cool yeah so there's really no value in a second API

If your PR goes in first I will make the connector ID mandatory

hdikeman:
Now that the physical plan input is deprecated I made the catalog parameter non-optional

Since catalog will always be explicitly specified we no longer need the defaultConnector, so I'm taking that out too

axiom/optimizer/SchemaResolver.h
```diff
@@ -24,25 +24,25 @@ namespace facebook::velox::optimizer {
...
-  // exist. Input can be any of the following formats:
+  // exist. If a connector for the specified catalog doesn't exist, an error
+  // will be returned. Input table name can be any of the following formats:
   //   - "tablename"
   //   - "schema.tablename"
   //   - "catalog.schema.tablename"
-  // If catalog is omitted, defaultConnector will be used for the table lookup.
   // If schema is omitted, defaultSchema will be prepended prior to lookup.
-  virtual const connector::Table* findTable(const std::string& name);
+  // If the table name specifies a different catalog than the one specified
```
mbasmanova, line 39: nit: Throws an error if ..

mbasmanova (approved):
Thanks.

PR #144, opened 2025-08-08
Add support for LIMIT

mbasmanova:
Summary: Part of https://github.com/facebookexperimental/verax/issues/136

Remaining follow-ups:
- Handle consecutive limits with optional projects in between.
- Add local exchange when numDrivers > 1.
- Handle limit followed by order by. Currently it is planned as TopN, which is incorrect.

Differential Revision: D79911387

### DIFF
--- a/axiom/optimizer/Plan.cpp
+++ b/axiom/optimizer/Plan.cpp
@@ -827,6 +827,11 @@ void Optimization::addPostprocess(
     auto* project = make<Project>(plan, dt->exprs, dt->columns);
     plan = project;
   }
+  if (dt->orderBy == nullptr && dt->limit >= 0) {
+    auto limit = make<Limit>(plan, dt->limit, dt->offset);
+    state.addCost(*limit);
+    plan = limit;
+  }
 }
 
 namespace {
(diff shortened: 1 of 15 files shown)


axiom/optimizer/Plan.cpp
```diff
@@ -827,6 +827,11 @@ void Optimization::addPostprocess(
     auto* project = make<Project>(plan, dt->exprs, dt->columns);
     plan = project;
   }
+  if (dt->limit >= 0) {
+    auto limit = make<Limit>(plan, dt->limit, dt->offset);
+    state.addCost(*limit);
```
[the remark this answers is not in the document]
mbasmanova: I don't think so. Need to look into how to do that.

axiom/optimizer/tests/PlanMatcher.h
```diff
@@ -63,6 +63,14 @@ class PlanMatcherBuilder {
   PlanMatcherBuilder& localPartition(
       const std::shared_ptr<PlanMatcher>& matcher);
 
+  PlanMatcherBuilder& partitionedOutput();
+
+  PlanMatcherBuilder& exchange();
+
+  PlanMatcherBuilder& limit();
+
+  PlanMatcherBuilder& limit(int32_t offset, int32_t count);
```
MBkkt, line 72: `int64_t`?

mbasmanova: Nice catch. Will fix.

PR #168, opened 2025-08-11
Fix bug in expression canonicalization

pedroerp:
Summary:
Fixing a canonicalization bug found when integrating to PySpark. The
wrong condition would invert expressions based on the node ids only, at times
leaving expressions in non-canonicalized order (literal to the left, column
reference to the right). The wrong order would prevent these filters to be
pushed down.

Differential Revision: D80025735


mbasmanova (approved):
Thanks!

I was wondering about this when writing the tests.

PR #171, opened 2025-08-11
Do not use OrderBy (RelationOp) in DerivedTable

mbasmanova:
Summary:
Do not mix physical and logical plan nodes.

Next step is to remove AggregationPlan from DerivedTable.

Differential Revision: D80043274


xiaoxmeng (approved):
@mbasmanova thanks!

PR #172, opened 2025-08-11
fix: Push-down limit physical node

MBkkt:
If I understand right PR approach is better for common case (reasonable and small limit value) and it can be worse for some edge cases.
So can we merge it this way?
And create issue about we need some cost based choice about push down projection/ limit.

Also this PR makes behavior of `limit` more similar to `order by + limit`


MBkkt:
Closed in favor of this https://github.com/facebookexperimental/verax/pull/356

PR #173, opened 2025-08-11
Fix distribution type for Limit and OrderBy

mbasmanova:
Summary:
Distribution type for Limit and OrderBy should be "gather" as both of these operations gather all the data into a single thread.

Part of https://github.com/facebookexperimental/verax/issues/167

Differential Revision: D80045128

### DIFF
--- a/axiom/optimizer/LogicalPlanToGraph.cpp
+++ b/axiom/optimizer/LogicalPlanToGraph.cpp
@@ -864,7 +864,9 @@ PlanObjectP Optimization::addOrderBy(const lp::SortNode& order) {
     keys.push_back(translateExpr(field.expression));
   }
 
-  currentSelect_->orderBy = make<OrderBy>(nullptr, keys, orderType);
+  currentSelect_->orderByKeys = keys;
+  currentSelect_->orderByTypes = orderType;
+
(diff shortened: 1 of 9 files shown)


xiaoxmeng (approved):
@mbasmanova thanks for the fix!

axiom/optimizer/LogicalPlanToGraph.cpp
```diff
@@ -864,7 +864,9 @@ PlanObjectP Optimization::addOrderBy(const lp::SortNode& order) {
     keys.push_back(translateExpr(field.expression));
   }
 
-  currentSelect_->orderBy = make<OrderBy>(nullptr, keys, orderType);
+  currentSelect_->orderByKeys = keys;
```
MBkkt, line 867: move?

axiom/optimizer/LogicalPlanToGraph.cpp
```diff
@@ -864,7 +864,9 @@ PlanObjectP Optimization::addOrderBy(const lp::SortNode& order) {
...
+  currentSelect_->orderByTypes = orderType;
```
MBkkt, line 868: move?

PR #174, opened 2025-08-12
Fix distribution type for Values

mbasmanova:
Summary:
The distribution for Values should be "gather". 

Part of https://github.com/facebookexperimental/verax/issues/167

Differential Revision: D80053079


mbasmanova:
@pedroerp However, the plan is still not quite right. As Join(Values, Values) has 3 stages instead of 1. This would need more fixing. Perhaps, we can add plan shape assertions when we fix that.

xiaoxmeng (approved):
@mbasmanova thanks!

PR #176, opened 2025-08-12
Fix AggregationPlan

mbasmanova:
Summary: AggregationPlan is a logical representation of an aggregation in a Query Graph. It should not use physical representation (Aggregation : RelationOp).

Differential Revision: D80085816


xiaoxmeng (approved):
@mbasmanova  LGTM. Thanks!

PR #180, opened 2025-08-12
Fold RelationOp::setCost into ctor

mbasmanova:
Summary:
RelationOp stores the estimate for output cardinality in 2 places:
- distribution().cardinality
- cost_.inputCardinality * cost_.fanout

distribution().cardinality is set in ctor, while cost_ is set in a separate setCost() method called externally.

This leads to discrepancies and makes these values unreliable.

This change removes RelationOp::setCost() method and folds its logic into ctor. This way both distribution and cost are filled at the same time.

These values are still not quite right. A follow-up is needed to fix them up. We also need to figure out how to write tests to ensure these values are right.

Differential Revision: D80092069


xiaoxmeng (approved):
@mbasmanova thanks % nits!

PR #181, opened 2025-08-12
feat: Add bucketing and Hive partition support to local Hive metadata

oerling:
Adds table creation and insert to LocalHiveConnectorMetadata.

If a table directory has a .schema file, this file is read to initialize bucket properties and partitioning columns. All table directories are read recursively and a file list is gathered with file paths, values of partitioning columns extracted from the directory path and a bucket number parsed from the file name.

Directories with no .schema continue to work as before. The table layout is inferred from the files.

Adds implementation of ConnectorMetadata write functions for insertion. An insert table handle is created with the partition and bucketing information from the single layout of a Hive table.

### DIFF
--- a/axiom/optimizer/connectors/ConnectorMetadata.h
+++ b/axiom/optimizer/connectors/ConnectorMetadata.h
@@ -28,6 +28,8 @@ namespace facebook::velox::core {
 // core:: but implementations do.
 class ITypedExpr;
 using TypedExprPtr = std::shared_ptr<const ITypedExpr>;
+
+class PartitionFunctionSpec;
 } // namespace facebook::velox::core
 
 /// Base classes for schema elements used in execution. A
@@ -188,6 +190,11 @@ class Column {
   std::mutex mutex_;
 };
 
+//// Describes the kind of table, e.g. durable vs. temporary.
+enum class TableKind { kTable, kTempTable };
+
+VELOX_DECLARE_ENUM_NAME(TableKind);
+
 class Table;
 
 /// Represents sorting order. Duplicate of core::SortOrder. Connectors
@@ -252,7 +259,9 @@ class TableLayout {
   }
 
   /// List of columns present in this layout.
-  const std::vector<const Column*>& columns() const;
+  const std::vector<const Column*>& columns() const {
+    return columns_;
+  }
 
   /// Set of partitioning columns. The values in partitioning columns determine
   /// the location of the row. Joins on equality of partitioning columns are
@@ -375,12 +384,24 @@ class Table {
   /// Returns an estimate of the number of rows in 'this'.
   virtual uint64_t numRows() const = 0;
 
+  virtual const std::unordered_map<std::string, std::string>& options() const {
+    return options_;
+  }
+
+  TableKind kind() const {
+    return kind_;
+  }
+
  protected:
   const std::string name_;
 
   // Discovered from data. In the event of different types, we take the
   // latest (i.e. widest) table type.
   RowTypePtr type_;
+
+  TableKind kind_{TableKind::kTable};
+
+  std::unordered_map<std::string, std::string> options_;
 };
 
 /// Describes a single partition of a TableLayout. A TableLayout has at least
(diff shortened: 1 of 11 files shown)


mbasmanova (commented):
@oerling Orri, would it be possible to add a README file that describes how HiveConnector works and how to use it to axiom/optimizer/connectors/hive/ directory?

axiom/optimizer/connectors/ConnectorMetadata.h
```diff
@@ -188,6 +190,9 @@ class Column {
   std::mutex mutex_;
 };
 
+//// Describes the kind of table, e.g. durable vs. temporary.
```
mbasmanova, line 193: Why does the engine need to know whether the table is temporary or not? What kind of logic depends on that?

axiom/optimizer/connectors/ConnectorMetadata.h
```diff
@@ -507,6 +526,53 @@ struct LookupKeys {
...
 };
 
+/// Describes how to repartition data before a TableWriter.
+struct WritePartitioning {
+  /// Columns for partitioning,. Names refer to the column names in the insert
+  /// table handle. Empty if any worker can write any row.
+  const std::vector<std::string> columns;
+
+  /// Specifies the partition function. nullptr if 'columns' is empty.
+  const std::shared_ptr<const core::PartitionFunctionSpec> partitionSpec;
```
mbasmanova, line 536: Can we add PartitionFunctionSpecPtr typedef?

axiom/optimizer/connectors/ConnectorMetadata.h
```diff
@@ -570,6 +636,90 @@ class ConnectorMetadata {
...
+  /// data is added to the table, finishWrite must be called after the
+  /// last writer is finished. Whether this autocommits a transaction
+  /// depends on the connector and session settings.
+  virtual ConnectorInsertTableHandlePtr createInsertTableHandle(
+      const TableLayout& layout,
+      const RowTypePtr& rowType,
+      const std::unordered_map<std::string, std::string>& options,
+      WriteKind kind,
+      const ConnectorSessionPtr& session) {
+    VELOX_UNSUPPORTED();
```
mbasmanova, line 692: It would be nicer to define required APIs as pure virtual to allow the compiler to point out when some are missing in the implementation.

axiom/optimizer/connectors/ConnectorMetadata.h
```diff
@@ -570,6 +636,90 @@ class ConnectorMetadata {
...
+      const TableLayout& layout,
+      const ConnectorInsertTableHandlePtr& handle,
+      const std::vector<RowVectorPtr>& writerResult,
+      WriteKind kind,
+      const ConnectorSessionPtr& session) {
+    VELOX_UNSUPPORTED();
+  }
+
+  /// Returns column handles whose value uniquely identifies a row for creating
+  /// and update or delete record. These may be for example some connector
```
mbasmanova, line 716: typo: and update -> an update

axiom/optimizer/connectors/hive/CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ velox_link_libraries(
   velox_hive_connector_metadata
   velox_connector_metadata
   velox_hive_connector
+  velox_type_fbhive
```
mbasmanova, line 22: Can we avoid using velox_type_fbhive ?

axiom/optimizer/connectors/hive/HiveConnectorMetadata.h
```diff
@@ -41,7 +42,13 @@ struct HivePartitionHandle : public PartitionHandle {
   const std::optional<int32_t> tableBucketNumber;
 };
 
+class HiveConnectorSession : public connector::ConnectorSession {
+ public:
+  ~HiveConnectorSession() override = default;
```
mbasmanova, line 47: I assume this dtor doesn't need to be defined explicitly.

axiom/optimizer/connectors/hive/HiveConnectorMetadata.h
```diff
@@ -121,7 +128,31 @@ class HiveConnectorMetadata : public ConnectorMetadata {
...
+
  protected:
+  virtual void ensureInitialized() const {}
+
+  virtual void validateOptions(
+      const std::unordered_map<std::string, std::string>& options) const;
+
+  virtual std::shared_ptr<connector::hive::LocationHandle> makeLocationHandle(
+      std::string targetDirectory,
+      std::optional<std::string> writeDirectory = std::nullopt,
```
mbasmanova, line 146: Can we drop default values for these arguments to make call sites easier to understand?

axiom/optimizer/connectors/hive/HiveConnectorMetadata.cpp
```diff
@@ -108,4 +108,78 @@ ConnectorTableHandlePtr HiveConnectorMetadata::createTableHandle(
...
+    inputColumns.push_back(std::static_pointer_cast<const HiveColumnHandle>(
+        createColumnHandle(layout, rowType->nameOf(i))));
+  }
+
+  std::shared_ptr<const HiveBucketProperty> bucketProperty;
+  auto* hiveLayout = reinterpret_cast<const HiveTableLayout*>(&layout);
+  if (hiveLayout->numBuckets().has_value()) {
+    std::vector<std::string> names;
+    std::vector<TypePtr> types;
+    for (auto& c : layout.partitionColumns()) {
```
mbasmanova, line 137: c -> column

mbasmanova (approved):
Thanks.

oerling:
The underlying interfaces need that.This is somewhere in the handles below. The commit protocol (renames in ws) depend on that. So, to create tables you must specify if they are temporary.  We  want an API that has all the parameters we expect to have.

PR #186, opened 2025-08-13
feat: Add support for cross joins

### LINKED ISSUE #169: Equi-join between 2 Values fails with "No cross joins"
```
SQL> explain select * from (select 1, 2) as t(a, b), (select 1, 3) as u(c, d) where t.a = u.c;
Logical plan:
- Project: -> ROW<_p4:INTEGER,_p5:INTEGER,_p6:INTEGER,_p7:INTEGER>
    _p4 := _p
    _p5 := _p0
    _p6 := _p2
    _p7 := _p3
  - Filter: eq(_p, _p2) -> ROW<_p:INTEGER,_p0:INTEGER,_p2:INTEGER,_p3:INTEGER>
    - Join INNER:  -> ROW<_p:INTEGER,_p0:INTEGER,_p2:INTEGER,_p3:INTEGER>
      - Project: -> ROW<_p:INTEGER,_p0:INTEGER>
          _p := 1
          _p0 := 2
        - Values: 1 rows -> ROW<_c:INTEGER>
      - Project: -> ROW<_p2:INTEGER,_p3:INTEGER>
          _p2 := 1
          _p3 := 3
        - Values: 1 rows -> ROW<_c1:INTEGER>

E0811 12:56:13.896360 454072 Exceptions.h:66] Line: fbcode/axiom/optimizer/Plan.cpp:1399, Function:crossJoin, Expression:  No cross joins, Source: RUNTIME, ErrorCode: NOT_IMPLEMENTED
Error: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: NOT_IMPLEMENTED
Reason: No cross joins
Retriable: False
Function: crossJoin
File: fbcode/axiom/optimizer/Plan.cpp
Line: 1399
```

CC: @pedroerp @feilong-liu

### LINKED ISSUE #358: Verax optimizer does not support cross join node
As title

https://github.com/facebookexperimental/verax/blob/9781c57a82fa99408fff997b627289604b481391/axiom/optimizer/Optimization.cpp#L1061

pashandor789:
- Implemented basic cross join logic at optimizer 
- Added nestedLoopJoin() to PlanMatcher

Termin:
Connected-Component - All the existing joins at the optimizer forms an edge between some relations except one - xJoin. So a join graph with xJoins would look like an islands of regular joins - each of these island we call connected component. These connected components can be also connected to each other  via xjoin.  

Example: (a join b) x (c join d join e) x (f join g)
connected components will be: a-b and c-d-e and f-g

Pipeline of the implementation  works as follows:
* If there are no neighboring tables left to process in the current connected component, we consider all remaining unplaced tables as candidates for a cross join.
* For each set of cross join candidates, we evaluate all possible combinations.

Considering all the combinations make sence for push downing complex filters depending on many tables
Example: 
A x B x C filter f(A, C) 
Combinations will be considered:
A x B x C 
A x C x B <- filter will be pushed down
B x A x C <- filter will be pushed down
B x C x A <- filter will be pushed down
C x A x B <- filter will be pushed down
C x B x A

Cost-formula explanation

```
cost_.unitCost = fanout *
        (Costs::kColumnRowCost + numRightColumns * Costs::kColumnByteCost)
```
to access one row of the crossjoin we actually have to consider that the right row was broadcasted. And it's a crossjoin so every row will produce x fanout rows 

```
cost_.setupCost = rightCardinality * rightByteSize;
```
setup we consider that the right part is broadcasted, this will encourage big tables to be chosen as the most left part of crossjoin

Follow-ups

- Considering all the combinations may be wasteful if we've no complex filter above to pushdown. So there's a way t fix it - https://github.com/facebookincubator/axiom/issues/489

Fixes #169


axiom/optimizer/Plan.cpp
```diff
@@ -1396,7 +1396,55 @@ void Optimization::crossJoin(
...
+    broadcastColumns.unionSet(availableColumns(buildTable));
+    broadcastTables.add(buildTable);
+  }
+  state.columns.unionSet(broadcastColumns);
+
+  auto memoKey = MemoKey{
+      candidate.tables[0], broadcastColumns, broadcastTables, candidate.existences};
+  
+  auto broadcast = Distribution::broadcast(
+      plan->distribution().distributionType, plan->resultCardinality());
```
pashandor789, line 1413: @mbasmanova 

So as I understand the candidate is a join tree that we want to join with the plan. In broadcast, we need to specify its cardinality - how do we access?

axiom/optimizer/tests/QuerySqlParser.cpp
```diff
@@ -16,9 +16,21 @@
 
 #include "axiom/optimizer/tests/QuerySqlParser.h"
 #include "velox/duckdb/conversion/DuckConversion.h"
-#include "velox/parse/DuckLogicalOperator.h"
 #include "velox/parse/PlanNodeIdGenerator.h"
 
+#include <duckdb/planner/operator/logical_get.hpp> // @manual
```
mbasmanova, line 21: Is this necessary? Can we avoid this?

axiom/optimizer/tests/QuerySqlParser.cpp
```diff
@@ -508,33 +520,34 @@ lp::LogicalPlanNodePtr toPlanNode(
...
-lp::LogicalPlanNodePtr toPlanNode(
-    ::duckdb::LogicalComparisonJoin& join,
-    memory::MemoryPool* pool,
-    std::vector<lp::LogicalPlanNodePtr> sources,
-    QueryContext& queryContext) {
-  VELOX_CHECK_EQ(2, sources.size());
-
-  lp::JoinType joinType = lp::JoinType::kInner;
-  switch (join.join_type) {
+lp::JoinType ConvertToLogicalPlanJoinType(
```
mbasmanova, line 523: Make sure to follow coding style. Function names start with lower case.

ConvertToLogicalPlanJoinType -> toJoinType

axiom/optimizer/tests/QuerySqlParser.cpp
```diff
@@ -555,6 +568,25 @@ lp::LogicalPlanNodePtr toPlanNode(
...
 }
 
+lp::LogicalPlanNodePtr toPlanNode(
+    ::duckdb::LogicalAnyJoin& join,
+    memory::MemoryPool* pool,
+    std::vector<lp::LogicalPlanNodePtr> sources,
+    QueryContext& queryContext) {
+  VELOX_CHECK_EQ(2, sources.size());
+
+  lp::JoinType joinType = ConvertToLogicalPlanJoinType(join.join_type);
```
mbasmanova, line 578: ConvertToLogicalPlanJoinType is very long. Rename to toJoinType.

Use auto: auto joinType = toJoinType(...);

axiom/optimizer/tests/QuerySqlParser.cpp
```diff
@@ -555,6 +568,25 @@ lp::LogicalPlanNodePtr toPlanNode(
...
+
+  lp::ExprPtr filter;
+  if (join.condition) {
+    const auto joinInputType = sources[0]->outputType()->unionWith(sources[1]->outputType());
```
mbasmanova, line 582: nit: perhaps, extract a helper function joinInputType(sources) to reduce copy-paste

mbasmanova (commented):
@pashandor789 Would add some tests? perhaps, to axiom/optimizer/tests/HiveQueriesTest.cpp

axiom/optimizer/tests/QuerySqlParser.cpp
pashandor789: otherwise it couldn't be compiled 

I've added AnyJoin here, but there's no any join in `#include "velox/parse/DuckLogicalOperator.h"`, so I've added import of it - `#include <duckdb/planner/operator/logical_any_join.hpp>` and it caused a conflict 

```
/usr/local/include/duckdb/planner/operator/logical_join.hpp: At global scope:
/usr/local/include/duckdb/planner/operator/logical_join.hpp:19:7: error: redefinition of ‘class duckdb::LogicalJoin’
   19 | class LogicalJoin : public LogicalOperator {
      |       ^~~~~~~~~~~
In file included from /workspaces/verax/axiom/optimizer/tests/QuerySqlParser.cpp:19:
/workspaces/verax/velox/velox/parse/DuckLogicalOperator.h:259:7: note: previous definition of ‘class duckdb::LogicalJoin’
  259 | class LogicalJoin : public LogicalOperator {
```

mbasmanova: Let's update velox/parse/DuckLogicalOperator.h like in https://github.com/facebookincubator/velox/pull/14453

axiom/optimizer/Plan.cpp
mbasmanova: The problem here is that Distribution struct is used for 2 different purposes: (1) describe desired distribution, in which case cardinality is not needed and doesn't make sense and therefore is ignored; (2) describe a dataset on disk or in memory, in which cardinality is useful.

We need to refactor Distribution to remove 'cardinality' member variable.

mbasmanova: #201

axiom/optimizer/tests/QuerySqlParser.cpp
```diff
@@ -537,33 +537,36 @@ lp::LogicalPlanNodePtr toPlanNode(
...
+      return lp::JoinType::kFull;
     default:
       VELOX_NYI(
           "Unsupported Duck join type: {}",
-          static_cast<int32_t>(join.join_type));
+          static_cast<int32_t>(join));
   }
+}
+
+std::shared_ptr<const RowType> joinInputType(const std::vector<lp::LogicalPlanNodePtr>& sources) {
```
mbasmanova, line 560: use RowTypePtr for brevity

I'm wondering if we can also move this logic into a method in LogicalPlanNode class.

axiom/optimizer/RelationOp.h
```diff
@@ -335,6 +335,11 @@ struct Join : public RelationOp {
   const QGstring& historyKey() const override;
 
   std::string toString(bool recursive, bool detail) const override;
+
+  static Join* makeCrossJoin(
```
mbasmanova, line 339: Let's use this in placeSingleRowDt.

Would you extract this refactoring into a separate PR so we can land that quickly?

axiom/optimizer/Plan.cpp
```diff
@@ -1410,7 +1410,47 @@ void Optimization::crossJoin(
     const JoinCandidate& candidate,
     PlanState& state,
     std::vector<NextJoin>& toTry) {
-  VELOX_NYI("No cross joins");
+  PlanStateSaver save(state, candidate);
+
+  PlanObjectSet broadcastTables;
```
mbasmanova, line 1415: Have you thought through about how do we enumerate all possible combinations of these joins? I assume we may want to sort the inputs by cardinality desc, then join left to right. Are we already doing that here?

Also, it seems like we would want to change the overall workflow to first break up inputs into connected components, produce best plans for these, then sort by cardinality and x-join.

What do you think?

axiom/optimizer/tests/QuerySqlParser.cpp
```diff
@@ -537,33 +537,38 @@ lp::LogicalPlanNodePtr toPlanNode(
...
-lp::LogicalPlanNodePtr toPlanNode(
-    ::duckdb::LogicalComparisonJoin& join,
-    memory::MemoryPool* pool,
-    std::vector<lp::LogicalPlanNodePtr> sources,
-    QueryContext& queryContext) {
-  VELOX_CHECK_EQ(2, sources.size());
-
-  lp::JoinType joinType = lp::JoinType::kInner;
-  switch (join.join_type) {
+lp::JoinType toJoinType(
```
mbasmanova, line 540: Perhaps, extract changing to  the parser into a separate PR and add a test that just parses SQL into logical plan without running the optimizer. We can land this sooner while we figure out proper impl for he optimizer.

oerling:
We looked at Thomas Neumann's publications way back and did not find any open source implementation to check out. We have friends who have been to Neumann's school and know about the subject matter, though. 

Over a decade ago I did the OpenLink Virtuoso QO that gets tons of joins, like 100's of them for RDF workloads (anybody remember linked data or the so called semantic web?)  It also get the relational benchmarks. So from experience we know that we can get good plans fast with our current algorithm. The other thing we need is to run cost model on physical plans and not some abstracion without cost for shuffles and detailsof join method. So we are not certain that taking joined pairs and treating them independently and then bubbling them up into bigger plans would do that very well. 

I expect you could do dp hyper with phhysical ops and index paths. But even so, there is the other thing that we want for fast planning time which is rapid cutoff. Once you have a somewhat good full plan, any partial plan that is worse is not worth pursuing. The bottom up dp does not seem to lend itself to that. Not in an obvious way at least. Also there is top k pushdown which is more natural to do top down as we do. 

The reasons may be somewhat subjective. But in any case we absolutely know we can hit the benchmarks and we can hit every late materialization and every very large number of joins case with what we have.  So doing what we do  is on one side familiar and on the other proven from experience to hit where we need to go. 

We can always tweak this to do more propagation of restrictions between join sides and so on. But we also have to be good with heterogenous access methods, because the world is no longer hash join only and big AI applications are bringing back physical design.


On the other hand, the query graph and join edge structure we have is a suitable input for dp hyper. So if somebody is ambitious and wants to show us how it's done, that would be a possibility.

pashandor789:
@oerling 
I used to work at YDB, where we implemented the DPHyp algorithm and also adapted it for shuffling. It handles 100+ joins efficiently in chain topology. If the input graph isn't a chain, we can transform it into one using the LinDP++ algorithm (Apache 2.0 licensed). Here’s the [link](https://github.com/ydb-platform/ydb/blob/aaa2358f4b9e680c511de94dedfc81fa1eea55b3/ydb/library/yql/dq/opt/dq_opt_dphyp_solver.h#L4) to the implementation.

axiom/optimizer/Plan.cpp
```diff
@@ -1417,20 +1417,61 @@ void Optimization::crossJoin(
...
+      candidate.tables[0], broadcastColumns, broadcastTables, candidate.existences};
+  
+  auto broadcast = Distribution::broadcast(plan->distribution().distributionType);
+  PlanObjectSet empty;
+  bool needsShuffle = false;
+  auto* rightPlan = makePlan(memoKey, broadcast, empty, candidate.existsFanout, state, needsShuffle);
+
+  RelationOpPtr rightOp = rightPlan->op;
+  PlanState broadcastState(state.optimization, state.dt, rightPlan);
+  if (needsShuffle && !isSingleWorker()) {
```
MBkkt, line 1440:
```suggestion
  if (needsShuffle) {
```

axiom/optimizer/Plan.cpp
pashandor789: @mbasmanova so what do you think 🤔

pashandor789: ok I will fix it

pashandor789: @mbasmanova 
but what about interesting orderings ?
we may shuffle more bytes but also we can save shuffling of the left operator .

pashandor789: @mbasmanova
for example cost(a x b) < cost(b x a), but shuffling of the b we may need on the top operator so we can eliminate redundant shuffle operator

pashandor789:
@mbasmanova I think this is done

mbasmanova (commented):
@pashandor789 Would you update the PR description to explain the changes? This will make it easier to review.

Is there sufficient test coverage? Does a single x-join cover all code paths and conditions?

axiom/optimizer/Plan.cpp
```diff
@@ -1412,25 +1416,91 @@ void Optimization::joinByHashRight(
...
     PlanState& state,
     std::vector<NextJoin>& toTry) {
-  VELOX_NYI("No cross joins");
+  if (crossJoins.empty()) {
+    return;
+  }
+
+  PlanStateSaver save(state);
+
+  std::sort(
```
mbasmanova, line 1430: When joining A x B x C, I believe, we want the largest table first, followed by tables in the increasing order of cardinality, e.g. A is largest, B is smallest, C is next smallest. Does this logic achieve that?

pashandor789: ye it does

pashandor789:
@mbasmanova have changed the description

axiom/optimizer/Optimization.cpp
```diff
@@ -362,15 +363,44 @@ std::vector<JoinCandidate> Optimization::nextJoins(PlanState& state) {
...
     state.dt->startTables.forEach([&](PlanObjectCP object) {
       if (!state.placed.contains(object)) {
-        candidates.emplace_back(nullptr, object, tableCardinality(object));
+        crossJoins.emplace_back(nullptr, object, tableCardinality(object));
       }
     });
   }
-  return candidates;
+
+  return {candidates, crossJoins};
```
MBkkt, line 377: move pair arguments

axiom/optimizer/Optimization.cpp
```diff
@@ -700,6 +730,52 @@ void Optimization::addPostprocess(
   }
 }
 
+RelationOpPtr Optimization::processCrossJoin(
+    const RelationOpPtr& plan,
+    const JoinCandidate& candidate,
+    PlanState& state) {
+  PlanObjectSet broadcastTables;
+  PlanObjectSet broadcastColumns;
+  for (auto buildTable : candidate.tables) {
```
MBkkt, line 739: `const auto*` otherwise it's unclear there's no unnecessary copies here?

axiom/optimizer/Plan.cpp
mbasmanova: @pashandor789 the PR description says

```
- process connected components
- sort them by cardinality desc
- cross join these connected components
```

, which is different from "When joining A x B x C, I believe, we want the largest table first, followed by tables in the increasing order of cardinality, e.g. A is largest, B is smallest, C is next smallest".

Would you clarify what is the behavior implemented in this PR?

axiom/optimizer/RelationOp.cpp
```diff
@@ -344,11 +344,16 @@ Join::Join(
       joinType(_joinType),
       right(std::move(_right)),
       leftKeys(std::move(_leftKeys)),
-      rightKeys(std::move(rightKeys)),
+      rightKeys(std::move(_rightKeys)),
       filter(std::move(filter)) {
   cost_.inputCardinality = inputCardinality();
   cost_.fanout = fanout;
 
+  if (method == JoinMethod::kCross) {
```
MBkkt, line 352: both cost function cross and hash looks wrong to me.

@mbasmanova can you explain current hash join cost code in details?


I don't understand why this is part of unitCost `cost_.fanout * rowCost`

mbasmanova: uniCost * inputCardinality should provide overall cost of the operator.

The formula makes sense if for each input row the operator processes fanout rows spending rowCost for each.

MBkkt: Ok I think I understand now

for query like

`select * from t1 join t2 on t1.x = t2.x`

fanout will be between 0 and 1, right?

axiom/optimizer/Plan.cpp
pashandor789: is It right now ?

mbasmanova:
@pashandor789 

I'm seeing that x-join of nation, region, orders uses join order orders x nation x region. I would expect orders x region x nation because card(region) < card(nation).

``
SQL> explain  SELECT 1 FROM nation, region, orders WHERE r_name like 'A%';
Fragment 0: stage1 numWorkers=4:
-- PartitionedOutput[6][SINGLE Presto] -> expr:BIGINT
  -- Project[5][expressions: (expr:BIGINT, 1)] -> expr:BIGINT
    -- NestedLoopJoin[4][INNER] -> r_name:VARCHAR
      -- NestedLoopJoin[2][INNER] ->
        -- TableScan[0][table: orders, scaleFactor: 0.01] ->
        -- TableScan[1][table: nation, scaleFactor: 0.01] ->
      -- TableScan[3][table: region, scaleFactor: 0.01, filter: like("r_name",A%)] -> r_name:VARCHAR

Fragment 1:  numWorkers=1:
-- Exchange[7][Presto] -> expr:BIGINT

Inputs:  7 <- stage1
```

Also seeing a crash in JoinCandidate::toString() when JoinCandidate.join is null.

Linter warning can probably be ignored.

<img width="816" height="237" alt="Screenshot 2025-08-30 at 5 40 56 PM" src="https://github.com/user-attachments/assets/985bf9f1-f178-499e-9119-1aafe9716826" />

**Comment by mbasmanova** [comment]
>     -- NestedLoopJoin[4][INNER] -> r_name:VARCHAR

Also, it seems redundant to project r_name from this join.

**Comment by mbasmanova** [comment]
Adding this log causes a crash:

```
    LOG(ERROR) << "Cross join: " << join.toString();
    crossJoinPlan = processCrossJoin(crossJoinPlan, join, state);
```

File: axiom/optimizer/Optimization.h [hunk 15]
```diff
@@ -235,6 +236,17 @@ class Optimization {
   // their different permutations.
   void tryNextJoins(PlanState& state, const std::vector<NextJoin>& nextJoins);
 
+  void processCrossJoins(
+      const RelationOpPtr& plan,
+      std::vector<JoinCandidate> crossJoins,
```
mbasmanova, line 241: const &

axiom/optimizer/Optimization.h
```diff
@@ -235,6 +236,17 @@ class Optimization {
   // their different permutations.
   void tryNextJoins(PlanState& state, const std::vector<NextJoin>& nextJoins);
 
+  void processCrossJoins(
```
mbasmanova, line 239: for consistency, perhaps, rename to crossJoins and crossJoin

pashandor789:
```
I'm seeing that x-join of nation, region, orders uses join order orders x nation x region. I would expect orders x region x nation because card(region) < card(nation).
```

does it make sense ? 
I think that it is important to take the least broadcast size cross join
so we take the first table the biggest and the others doesn't matter 

ordering 
orders x region x nation
orders x nation x region 

will have the same broadcast data

axiom/optimizer/Optimization.cpp
```diff
@@ -362,15 +363,45 @@ std::vector<JoinCandidate> Optimization::nextJoins(PlanState& state) {
...
+  std::sort(
+      crossJoins.begin(),
+      crossJoins.end(),
+      [](const JoinCandidate& left, const JoinCandidate& right) {
+        return left.fanout > right.fanout;
+      });
+
+  auto crossJoinPlan = plan;
+  for (const auto& join : crossJoins) {
+    LOG(ERROR) << "Cross join: " << join.toString();
```
mbasmanova, line 400: Let's remember to remove this.

axiom/optimizer/Plan.cpp
```diff
@@ -387,7 +387,10 @@ bool JoinCandidate::isDominantEdge(PlanState& state, JoinEdgeP edge) {
 
 std::string JoinCandidate::toString() const {
   std::stringstream out;
-  out << join->toString() << " fanout " << fanout;
+  if (join) {
+    out << join->toString();
```
mbasmanova, line 391: This avoids the crash, but doesn't produce a meaningful output. For x-join, join edge is null, hence, we need to print the first table in 'tables', which is not printed for other joins as it is included in the join edge.

axiom/optimizer/tests/HiveQueriesTest.cpp
```diff
@@ -75,6 +75,77 @@ TEST_F(HiveQueriesTest, basic) {
...
+                         "n_nationkey",
+                         "n_name",
+                         "n_regionkey",
+                         "n_comment",
+                         "r_regionkey",
+                         "r_name",
+                         "r_comment"})
+                    .limit(0, 10, false)
+                    .planNode();
+    checkResults("SELECT * FROM customer, nation, region LIMIT 10", plan);
```
mbasmanova, line 131: checkResults verifies that the query can be planned and executed successfully and produces the right results; it doesn't verify the plan shape and therefore is not enough to ensure the optimizer produces the right plan

we need to add more verification to ensure the plan shape is right

mbasmanova:
@pashandor789 Would you elaborate on what have changed since last review, how you are testing / verifying these changes, how do you ensure the functionality is complete?

mbasmanova:
@pashandor789 Would you respond to each earlier comment to clarify how and if that comment was addressed? Please, provide detailed answers to help reduce the amount of reverse engineering and guessing during review.

Specifically, what did you do about https://github.com/facebookexperimental/verax/pull/186#issuecomment-3239576207 ?

pashandor789:
about partitiong I mean that two tables may have almost the same size but one produce a useful partitioning (which maybe used for aggregation and we won't shuffle data) and the other will produce a partitioning which is not compatible with the top aggregation (its for example, it can be not only aggregation)

it's not wasteful, only cross chain length check

pashandor789:
well there's an example 

сross join for tables t1(10), t2(100), t3(1000), t4(10000) we will consider these cases:

t1, t2, t3, t4 
t2, t1, t3, t4
t3, t1, t2, t4
t4, t1, t2, t3

Like because the final partitioning will be the same as in the first table, and we might benefit from it for collocated joining/aggregation and e.t.c.

we don't consider all the combinations because broadcast size won't be changed and there won't be any new partitioning except these (t1, t2, t3, t4)

mbasmanova:
> For the left part we enumerate all the possible partitioning cases 

@pashandor789 Do we already have tests to verify this logic? If not, let's add.

axiom/optimizer/tests/PlanMatcher.cpp
```diff
@@ -557,6 +557,39 @@ class HashJoinMatcher : public PlanMatcherImpl<HashJoinNode> {
...
+      const std::shared_ptr<PlanMatcher>& right,
+      JoinType joinType)
+      : PlanMatcherImpl<NestedLoopJoinNode>({left, right}),
+        joinType_{joinType} {}
+
+  bool matchDetails(const NestedLoopJoinNode& plan) const override {
+    SCOPED_TRACE(plan.toString(true, false));
+
+    if (joinType_.has_value()) {
+      EXPECT_EQ(
```
mbasmanova, line 578: Can we not use EXPECT_EQ(plan.joinType(), joinType_.value())?

pashandor789: do you want to remove it ?

I just was thinking about case with left join with non equi filter 
it will matter

mbasmanova: I'm asking why can't we compare enum values directly? Why do we need to convert these into strings?

pashandor789:
@mbasmanova 

Hello there! I was writing a test for the implementation we're discussing and found a bag where a join condition was lost during enumeration because of incorrect processing of connected components, also this approach works bad with complex filters which depends on many tables - it doesn't push down it.

Also, the bug was non-trivial to fix so I've decided to make a new approach. You can read about it in description. It's a little disadvantage which is going to be fixed in follow-up (I've provided there a design doc about efficient cross join enumeration you can click the link).

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1352,6 +1352,71 @@ TEST_F(PlanTest, lastProjection) {
   ASSERT_TRUE(matcher->match(plan));
 }
 
+TEST_F(PlanTest, xJoinFilterPushdown) {
+  auto aType = ROW({"a_id", "a_field"}, {BIGINT(), BIGINT()});
```
mbasmanova, line 1356: {BIGINT(), BIGINT()} -> BIGINT()

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1352,6 +1352,71 @@ TEST_F(PlanTest, lastProjection) {
...
+  auto cType = ROW({"c_id", "c_field"}, {BIGINT(), BIGINT()});
+
+  testConnector_->addTable("a", aType);
+  testConnector_->addTable("b", bType);
+  testConnector_->addTable("c", cType);
+
+  lp::PlanBuilder::Context context(kTestConnectorId);
+  auto logicalPlan =
+      lp::PlanBuilder(context)
+          .tableScan("a", aType->names())
```
mbasmanova, line 1367: drop aType->names() argument; ditto other places

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1352,6 +1352,71 @@ TEST_F(PlanTest, lastProjection) {
...
+          .crossJoin(lp::PlanBuilder(context).tableScan("b", bType->names()))
+          .crossJoin(lp::PlanBuilder(context).tableScan("c", cType->names()))
+          .filter("a_field != b_field")
```
mbasmanova, line 1370: Note that it is Ok to have duplicate column names across tables. PlanBuilder provides "as" API to disambiguate. Check out examples in PlanPrinterTest.cpp

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1352,6 +1352,71 @@ TEST_F(PlanTest, lastProjection) {
...
+          .build();
+
+  auto plan = toSingleNodePlan(logicalPlan);
```
mbasmanova, line 1373: Can we also test multi-threaded and distributed plans?

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1352,6 +1352,71 @@ TEST_F(PlanTest, lastProjection) {
...
+
+  auto matcher =
+      core::PlanMatcherBuilder()
```
mbasmanova, line 1376: consider using lambda variable to reduce boiler plate; see UnnestTest

```
auto startMatcher = [](const auto& tableName) {
   return core::PlanMatcherBuilder().tableScan(tableName);

auto matcher = startMatcher("a")
   .nestedLoopJoin(startMatcher("b").build())
   .filter("a_field != b_field")
   .nestedLoopJoin(startMatcher("c").build())
   .build();
};
```

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1352,6 +1352,71 @@ TEST_F(PlanTest, lastProjection) {
...
+}
+
+TEST_F(PlanTest, xJoinManyTables) {
+  auto nationType =
+      ROW({"n_nationkey", "n_name", "n_regionkey"},
+          {BIGINT(), VARCHAR(), BIGINT()});
+  auto regionType = ROW({"r_regionkey", "r_name"}, {BIGINT(), VARCHAR()});
+  auto customerType = ROW(
+      {"c_custkey", "c_name", "c_nationkey"}, {BIGINT(), VARCHAR(), BIGINT()});
+  auto lineitemType = ROW({"l_orderkey", "l_partkey"}, {BIGINT(), BIGINT()});
```
mbasmanova, line 1392: use HiveQueriesTestBase::getSchema

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1352,6 +1352,71 @@ TEST_F(PlanTest, lastProjection) {
...
+
+  const auto connectorId = exec::test::kHiveConnectorId;
+  const auto connector = velox::connector::getConnector(connectorId);
+
+  lp::PlanBuilder::Context context;
+  auto logicalPlan = lp::PlanBuilder(context)
+                         .tableScan(connectorId, "nation", nationType->names())
```
mbasmanova, line 1399: drop nationType->names() argument here and in all other places

mbasmanova (commented):
> Also, the final partitioning will be the same as in the first table, and we might benefit from it for collocated joining/aggregation and e.t.c.

Are we missing test coverage for this functionality?

axiom/optimizer/tests/PlanMatcher.cpp
```diff
@@ -889,4 +920,27 @@ PlanMatcherBuilder& PlanMatcherBuilder::orderBy(
   return *this;
 }
 
+PlanMatcherBuilder& PlanMatcherBuilder::nestedLoopJoin() {
+  VELOX_USER_CHECK_NOT_NULL(matcher_);
+  matcher_ = std::make_shared<NestedLoopJoinMatcher>(
+      matcher_, std::make_shared<PlanMatcherImpl<>>());
```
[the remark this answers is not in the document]
pashandor789: I think that it may be useful for cases when we want to pin the first table and it doesn't matter what others will be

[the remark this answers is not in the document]
pashandor789: ye
that's the case 
```
 auto matcher = core::PlanMatcherBuilder()
                     .tableScan("lineitem")
                     .nestedLoopJoin()
                     .nestedLoopJoin()
                     .nestedLoopJoin()
                     .project()
                     .build();
```
here I think lineitem must be the first because we'll broadcast everything there and I think that it's not important in which order they'll be broadcasted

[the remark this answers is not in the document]
mbasmanova: I see. I guess this only works when right side is a single leaf node. If it is a tree, then it won't work because default matcher used for the right side expect 0 sources. Assuming that's the case, this API is very limited and therefore not very useful. Let's remove it.

> lineitem must be the first because we'll broadcast everything there and I think that it's not important in which order they'll be broadcasted

It may not be important what the order is, but we do want to ensure deterministic order. Hence, it is beneficial to verify the actual order in the test.

PR #193, opened 2025-08-13
Fix ci to run verax tests

feilong-liu:


axiom/optimizer/tests/HiveQueriesTest.cpp
```diff
@@ -25,7 +25,7 @@ namespace {
 
 class HiveQueriesTest : public test::HiveQueriesTestBase {};
 
-TEST_F(HiveQueriesTest, basic) {
+TEST_F(HiveQueriesTest, DISABLED_basic) {
```
mbasmanova, line 28: Please, do not disable tests. We don't want to lose all coverage.

mbasmanova (changes requested):

mbasmanova:
Verax is an informal project name. Let's not use it in CMake targets or anywhere else in the code.

.github/workflows/build_and_test.yml
```diff
@@ -103,5 +103,7 @@ jobs:
       # Run Verax tests (excluding velox tests)
       - name: Run Debug Tests
         working-directory: _build/debug
+        env:
+          GTEST_FILTER: "-TpchPlanTest.q01:TpchPlanTest.q03:TpchPlanTest.q05:HiveQueriesTest.basic:PlanTest.filterImport:PlanTest.intersect:PlanTest.except"
```
feilong-liu, line 107: These tests are failing. Tried to fix the tests, for example TpchPlanTest.q01 failed with error C++ exception with description "ExtractPlan can only prepare a single statement" thrown in the test body. which I previously thought is due to the additional ; in the query, however removing it does not solve the problem, adding code change in sql read function to get rid of ; also does not work.

Considering that this PR will 1) at least let us to run other tests, and better than no tests running for now 2) will not disable tests in meta internal infrastructure 3) also shorten the compile time since we are disabling build of velox test. I think perhaps we should merge this change while investigating why the tests failed in CI.

feilong-liu: Issue created for the failed tests https://github.com/facebookexperimental/verax/issues/204

feilong-liu: @mbasmanova  how do you think?

mbasmanova: Let's add a TODO with a link to GitHub issue and prioritize fixing.

mbasmanova:
@feilong-liu Would you update PR description to say how long CI took before and after this change?

feilong-liu:
> @feilong-liu Would you update PR description to say how long CI took before and after this change?

The CI duration decreased from 1 hour to 30 minutes, updated in the PR

.github/workflows/build_and_test.yml
feilong-liu: Added TODO as comment

kgpai (approved):
LGTM.

PR #196, opened 2025-08-13
Bug fix for complex types in ValuesNode

pedroerp:
Summary:
Complex types in value nodes were failing because the type pointer was
not being acquired in QueryGraphContext.

Differential Revision: D80222042

### DIFF
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -220,7 +220,7 @@ void ToGraph::getExprForField(
       auto it = renames_.find(name);
       VELOX_CHECK(it != renames_.end());
       auto maybeColumn = it->second;
-      VELOX_CHECK(maybeColumn->type() == PlanType::kColumnExpr);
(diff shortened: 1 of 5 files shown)


axiom/optimizer/ToGraph.cpp
```diff
@@ -1171,7 +1171,7 @@ PlanObjectP ToGraph::makeValuesTable(const lp::ValuesNode& values) {
     VELOX_DCHECK_LT(i, type->size());
 
     const auto& name = names[i];
-    Value value{type->childAt(i).get(), cardinality};
+    Value value{toType(type->childAt(i)), cardinality};
```
pedroerp, line 1174: @MBkkt small bug fix on acquiring type pointers for ValuesNode

pedroerp:
@oerling fix for the bug we chatted about offline.

pedroerp:
@MBkkt my understanding from reading the code is that any Type objects you carry inside the query graph need to be deduplicated and allocated in the graph arena. The toType() function maps from the pointer back to the original shared_ptr. This worked for primitive types because they are always statically allocated - but this isn't true for complex types.

PR #197, opened 2025-08-14
Fix output names and drop redundant projects

mbasmanova:
Summary:
Fix output names to match logical plan. Remove some redundant projects.

- In ToGraph, store names from logical plan in Column.alias.
- In ToVelox, use Column.alias for names in Velox plan.

Fixes https://github.com/facebookexperimental/verax/issues/195

Differential Revision: D80229006


xiaoxmeng (approved):
@mbasmanova thanks!

PR #198, opened 2025-08-14
Add cache option in CMake for CI build

feilong-liu:


mbasmanova (commented):
@feilong-liu Would you clarify what does this do and how did you verify that it indeed has the desired effect?

CC: @kgpai

feilong-liu:
> @feilong-liu Would you clarify what does this do and how did you verify that it indeed has the desired effect?
> 
> CC: @kgpai

Had a meeting with @kgpai, @kgpai suspected that we need to checkin the code as cache will need to be written from main branch. So we decide to check in the changes with debug code, and we will check and iterate based on it. cc @kgpai in case I missed anything or interpret it wrong.

.github/workflows/build_and_test.yml
```diff
@@ -43,16 +43,46 @@ jobs:
...
     steps:
+      - name: Debug Initial Environment
+        run: |
+          echo "=== Cache Environment Debug ==="
+          echo "CCACHE_DIR: $CCACHE_DIR"
+          echo "Workspace: ${{ github.workspace }}"
+          echo "Runner OS: ${{ runner.os }}"
+          echo "Runner Architecture: ${{ runner.arch }}"
+          echo "Current directory: $(pwd)"
+          echo "Available disk space:"
```
kgpai, line 54: nit: have you run out of disk space at any point to have to output this ?

.github/workflows/build_and_test.yml
```diff
@@ -43,16 +43,46 @@ jobs:
       VELOX_DEPENDENCY_SOURCE: SYSTEM
       ICU_SOURCE: SYSTEM
     steps:
+      - name: Debug Initial Environment
+        run: |
+          echo "=== Cache Environment Debug ==="
+          echo "CCACHE_DIR: $CCACHE_DIR"
+          echo "Workspace: ${{ github.workspace }}"
+          echo "Runner OS: ${{ runner.os }}"
```
kgpai, line 51: nit: Runner os, architecture are all output by default when the workflow runs

.github/workflows/build_and_test.yml
```diff
@@ -66,27 +96,112 @@ jobs:
...
+          echo "Available memory:"
+          free -h
+          echo "Available CPU cores:"
+          nproc
+          echo ""
+          echo "=== Sample Compilation Commands (first 5) ==="
+          echo "Showing first few compilation commands to verify ccache usage:"
+          # Build with verbose output but limit the output to first few commands
+          (cd _build/debug && \
+          make VERBOSE=1 -j1 2>&1 | head -20 | grep -E "(ccache|g\+\+|clang)") || echo "No compiler commands found in first 20 lines"
```
kgpai, line 143: nit: I dont understand why do this, you can see that anyway when you do make below..

CMakeLists.txt
```diff
@@ -112,6 +114,76 @@ endif()
...
+        "  - CMAKE_C_COMPILER_LAUNCHER already set: ${CMAKE_C_COMPILER_LAUNCHER}"
+    )
+  endif()
+  if(CMAKE_CXX_COMPILER_LAUNCHER)
+    message(
+      STATUS
+        "  - CMAKE_CXX_COMPILER_LAUNCHER already set: ${CMAKE_CXX_COMPILER_LAUNCHER}"
+    )
+  endif()
+  if(CMAKE_CUDA_COMPILER_LAUNCHER)
```
kgpai, line 174: Does verax need cuda compiler ?

kgpai (approved):
Approving to unblock so that we can validate that we hit the cache with main builds.

PR #212, opened 2025-08-16
feat: Add AnyJoin parsing

pashandor789:


axiom/optimizer/tests/HiveQueriesTest.cpp
```diff
@@ -68,6 +68,10 @@ TEST_F(HiveQueriesTest, basic) {
           .planNode());
 }
 
+TEST_F(HiveQueriesTest, AnyJoin) {
+  checkLogicalPlanParsing("SELECT * FROM nation JOIN region on true");
```
mbasmanova, line 72: nit: on -> ON for consistency

axiom/optimizer/tests/HiveQueriesTest.cpp
```diff
@@ -68,6 +68,16 @@ TEST_F(HiveQueriesTest, basic) {
           .planNode());
 }
 
+TEST_F(HiveQueriesTest, anyJoin) {
+  std::string sql = "SELECT * FROM nation JOIN region ON true";
+  SCOPED_TRACE(sql);
```
mbasmanova, line 73: this is not needed since there is only one query in this test.

mbasmanova (commented):
Thanks.

PR #213, opened 2025-08-16
feat: Add constant folding for special form expression

feilong-liu:
Fixes https://github.com/facebookexperimental/verax/issues/200

Differential Revision: D80386643

### DIFF
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -97,6 +97,17 @@ void ToGraph::setDtUsedOutput(
   }
 }
 
+namespace {
+bool isConstantTrue(ExprCP expr) {
+  if (expr->type() != PlanType::kLiteralExpr) {
+    return false;
+  }
+  const auto& variant = expr->as<Literal>()->literal();
+  return variant.kind() == TypeKind::BOOLEAN && !variant.isNull() &&
+      variant.value<bool>();
+}
+} // namespace
+
 void ToGraph::translateConjuncts(const lp::ExprPtr& input, ExprVector& flat) {
   if (!input) {
     return;
@@ -106,22 +117,21 @@ void ToGraph::translateConjuncts(const lp::ExprPtr& input, ExprVector& flat) {
       translateConjuncts(child, flat);
     }
   } else {
-    flat.push_back(translateExpr(input));
+    auto translatedExpr = translateExpr(input);
+    if (!isConstantTrue(translatedExpr)) {
+      flat.push_back(translatedExpr);
+    }
   }
 }
 
 ExprCP ToGraph::tryFoldConstant(
-    const lp::CallExpr* call,
-    const lp::SpecialFormExpr* cast,
+    const TypePtr& returnType,
+    const std::string& callName,
     const ExprVector& literals) {
   try {
-    Value value(call ? toType(call->type()) : toType(cast->type()), 1);
+    Value value(toType(returnType), 1);
     auto* veraxExpr = make<Call>(
-        PlanType::kCallExpr,
-        cast ? toName("cast") : toName(call->name()),
-        value,
-        literals,
-        FunctionSet());
+        PlanType::kCallExpr, toName(callName), value, literals, FunctionSet());
     auto typedExpr = queryCtx()->optimization()->toTypedExpr(veraxExpr);
     auto exprSet = evaluator_.compile(typedExpr);
     auto first = exprSet->exprs().front().get();
@@ -492,8 +502,15 @@ BuiltinNames::BuiltinNames()
       gte(toName("gte")),
       plus(toName("plus")),
       multiply(toName("multiply")),
-      _and(toName("and")),
-      _or(toName("or")) {
+      _and(toName(SpecialFormCallNames::kAnd)),
+      _or(toName(SpecialFormCallNames::kOr)),
+      cast(toName(SpecialFormCallNames::kCast)),
+      tryCast(toName(SpecialFormCallNames::kTryCast)),
+      _try(toName(SpecialFormCallNames::kTry)),
+      coalesce(toName(SpecialFormCallNames::kCoalesce)),
+      _if(toName(SpecialFormCallNames::kIf)),
+      _switch(toName(SpecialFormCallNames::kSwitch)),
+      in(toName(SpecialFormCallNames::kIn)) {
   canonicalizable.insert(eq);
   canonicalizable.insert(lt);
   canonicalizable.insert(lte);
@@ -600,29 +617,6 @@ ExprCP ToGraph::makeConstant(const lp::ConstantExpr& constant) {
 }
 
 namespace {
-const char* specialFormCallName(const lp::SpecialFormExpr* form) {
-  switch (form->form()) {
-    case lp::SpecialForm::kAnd:
-      return "and";
-    case lp::SpecialForm::kOr:
-      return "or";
-    case lp::SpecialForm::kCast:
-      return "cast";
-    case lp::SpecialForm::kTryCast:
-      return "trycast";
-    case lp::SpecialForm::kCoalesce:
-      return "coalesce";
-    case lp::SpecialForm::kIf:
-      return "if";
-    case lp::SpecialForm::kSwitch:
-      return "switch";
-    case lp::SpecialForm::kIn:
-      return "in";
-    default:
-      VELOX_UNREACHABLE(lp::SpecialFormName::toName(form->form()));
-  }
-}
-
 // Returns bits describing function 'name'.
 FunctionSet functionBits(Name name) {
   if (auto* md = functionMetadata(name)) {
@@ -652,6 +646,10 @@ ExprCP ToGraph::translateExpr(const lp::ExprPtr& expr) {
     return path.value();
   }
 
+  if (expr->isLambda()) {
+    return translateLambda(expr->asUnchecked<lp::LambdaExpr>());
+  }
+
   ToGraphContext ctx(expr.get());
   ExceptionContextSetter s(makeExceptionContext(&ctx));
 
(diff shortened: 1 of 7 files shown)


axiom/optimizer/ToGraph.cpp
```diff
@@ -106,19 +106,63 @@ void ToGraph::translateConjuncts(const lp::ExprPtr& input, ExprVector& flat) {
...
+    case lp::SpecialForm::kCoalesce:
+      return "coalesce";
+    case lp::SpecialForm::kIf:
+      return "if";
+    case lp::SpecialForm::kSwitch:
+      return "switch";
+    case lp::SpecialForm::kIn:
+      return "in";
+    default:
+      VELOX_UNREACHABLE(lp::SpecialFormName::toName(form->form()));
```
mbasmanova, line 136: Let's remove 'default' case altogether.

axiom/optimizer/ToVelox.cpp
```diff
@@ -411,6 +411,11 @@ core::TypedExprPtr ToVelox::toTypedExpr(ExprCP expr) {
             toTypePtr(expr->value().type), std::move(inputs), false);
       }
 
+      if (call->name() == toName("trycast")) {
```
mbasmanova, line 414: It is error prone to hard-codes names in multiple places. Let's introduce constants.

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -214,6 +214,206 @@ TEST_F(PlanTest, rejectedFilters) {
...
+                       .project()
+                       .build();
+
+    auto plan = toSingleNodePlan(logicalPlan);
+    ASSERT_TRUE(matcher->match(plan));
+  }
+  {
+    auto logicalPlan = lp::PlanBuilder()
+                           .tableScan(kTestConnectorId, "numbers")
+                           .filter("if(2>1, true, false)")
```
mbasmanova, line 250: 2>1 -> 2 > 1 for readability

mbasmanova: This example is a bit strange. Can we have something more realistic? Can we just use filter("2 > 1")?

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -214,6 +214,206 @@ TEST_F(PlanTest, rejectedFilters) {
...
+                       .tableScan()
+                       .filter("false")
+                       .project()
+                       .build();
+
+    auto plan = toSingleNodePlan(logicalPlan);
+    ASSERT_TRUE(matcher->match(plan));
+  }
+  {
+    auto logicalPlan = lp::PlanBuilder()
```
mbasmanova, line 353: There is a lot of repetition here, which makes it hard to understand what's being tested. Would you refactor?

axiom/optimizer/tests/PlanTest.cpp
mbasmanova: I see, can we maybe write it as if(1 > 2, 3 + 4, 4 + 5)? 

Should we also test partially foldable expressions? if (a > b, 1 + 2, 3 + c)?

axiom/optimizer/ToGraph.cpp
mbasmanova: If we don't handle all enum values, then why we say "unreachable"? Shouldn't we handle these?

feilong-liu: Change it from VELOX_UNREACHABLE to VELOX_FAIL

axiom/optimizer/ToVelox.cpp
feilong-liu: Since `specialFormCallName` is the source of truth and used in both logical plan to verax plan, and verax plan to velox plan, I pull it out to be a util function and used in both ToGraph and ToVelox, to replace the hard-code string here.

axiom/optimizer/tests/PlanTest.cpp
feilong-liu: Simplified the test logic.

axiom/optimizer/ToVelox.cpp
```diff
@@ -971,7 +984,9 @@ namespace {
 core::TypedExprPtr toAndWithAliases(
     const std::vector<core::TypedExprPtr>& exprs,
     const BaseTable* baseTable) {
-  auto result = std::make_shared<core::CallTypedExpr>(BOOLEAN(), exprs, "and");
+  auto result = exprs.size() == 1
+      ? exprs.at(0)
+      : std::make_shared<core::CallTypedExpr>(BOOLEAN(), exprs, "and");
```
mbasmanova, line 989: can we replace hard-coded "and" here?

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -214,6 +214,64 @@ TEST_F(PlanTest, rejectedFilters) {
...
+      {"try_cast(1 as BIGINT) > 4", "false"},
+      {"try(10 / 1) > 4", "true"},
+      {"coalesce(cast(null as boolean), false)", "false"},
+      {"case when 1 > 2 then true else false end", "false"},
+      {"try_cast('a' as BIGINT) > 4", "null"},
+      {"if(a > b, 1 + 2, c) > b + 3", "if(a > b, 3, c) > b + 3"},
+      {"if(a > b, 1 + 2, 3 + 4) > b + 3", "if(a > b, 3, 7) > b + 3"},
+  };
+
+  auto testConstantFold =
```
mbasmanova, line 248: this lambda is used only once; let's fold it into the caller

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -214,6 +214,64 @@ TEST_F(PlanTest, rejectedFilters) {
...
+      [&](const std::string& filter,
+          const std::optional<std::string>& expectedFilter) {
+        auto logicalPlan = lp::PlanBuilder()
+                               .tableScan(kTestConnectorId, "numbers")
+                               .filter(filter)
+                               .map({"a + 2"})
+                               .build();
+
+        auto matcher = expectedFilter == "true"
```
mbasmanova, line 257: it is a bit more robust to change expectedFilter to std::optional.

mbasmanova (commented):
@feilong-liu Thank you for iterating. This change is shaping up nicely. The tests case is a lot easier to understand now.

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -214,6 +214,64 @@ TEST_F(PlanTest, rejectedFilters) {
...
+TEST_F(PlanTest, specialFormConstantFold) {
+  testConnector_->createTable(
+      "numbers", ROW({"a", "b", "c"}, {BIGINT(), BIGINT(), BIGINT()}));
+
+  struct TestCase {
+    std::string filter;
+    std::string expectedFilter;
+  };
+
+  std::vector<TestCase> testCases = {
```
mbasmanova, line 226: Why do we only tests constant folding in filter? Should we also test that it works in project?

axiom/optimizer/tests/PlanTest.cpp
feilong-liu: Done

axiom/optimizer/tests/PlanTest.cpp
feilong-liu: Done

axiom/optimizer/tests/PlanTest.cpp
feilong-liu: Added tests for project too

axiom/optimizer/PlanUtils.h
```diff
@@ -108,4 +108,20 @@ std::optional<int64_t> maybeIntegerLiteral(
...
 std::string conjunctsToString(const ExprVector& conjuncts);
 
+bool isConstantTrue(ExprCP expr);
+
+const char* specialFormCallName(const logical_plan::SpecialForm& form);
+
+namespace SpecialFormCallNames {
+constexpr const char* kAnd = "and";
+constexpr const char* kOr = "or";
+constexpr const char* kCast = "cast";
```
mbasmanova, line 118: Should we expand BuiltinNames?

```
struct BuiltinNames {
  BuiltinNames();

  Name reverse(Name op) const;

  bool isCanonicalizable(Name name) const {
    return canonicalizable.find(name) != canonicalizable.end();
  }

  Name eq;
  Name lt;
  Name lte;
  Name gt;
  Name gte;
  Name plus;
  Name multiply;
  Name _and;
  Name _or;

  folly::F14FastSet<Name> canonicalizable;
};
```

axiom/optimizer/PlanUtils.h
```diff
@@ -108,4 +108,20 @@ std::optional<int64_t> maybeIntegerLiteral(
 
 std::string conjunctsToString(const ExprVector& conjuncts);
 
+bool isConstantTrue(ExprCP expr);
+
+const char* specialFormCallName(const logical_plan::SpecialForm& form);
+
+namespace SpecialFormCallNames {
```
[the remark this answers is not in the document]
feilong-liu: Changed to struct and moved to next to Call expression

axiom/optimizer/ToVelox.cpp
```diff
@@ -396,7 +403,7 @@ core::TypedExprPtr ToVelox::toTypedExpr(ExprCP expr) {
       std::vector<core::TypedExprPtr> inputs;
       auto call = expr->as<Call>();
 
-      if (call->name() == toName("in")) {
+      if (call->name() == toName(SpecialFormCallNames::kIn)) {
```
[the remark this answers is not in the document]
feilong-liu: Update the BuiltinNames to include special form, and use the Name defined in special form directly now to avoid calling toName

axiom/optimizer/PlanUtils.h
```diff
@@ -108,4 +108,8 @@ std::optional<int64_t> maybeIntegerLiteral(
 
 std::string conjunctsToString(const ExprVector& conjuncts);
 
+bool isConstantTrue(ExprCP expr);
```
mbasmanova, line 111: Assuming this is used only once in ToGraph.cpp, it can go into anonymous namespace in that .cpp

axiom/optimizer/PlanUtils.h
```diff
@@ -108,4 +108,8 @@ std::optional<int64_t> maybeIntegerLiteral(
...
+
+const char* specialFormCallName(const logical_plan::SpecialForm& form);
```
mbasmanova, line 113: This can be a method on SpecialFormCallNames

SpecialFormCallNames::toName(form)

In general, we want to avoid putting too many things into generic Utils and prefer better organization.

mbasmanova (approved):
Thanks.

PR #214, opened 2025-08-16
Fix incorrect and strange choice of best plan

MBkkt:
I don't understand current `PlanSet::best(`

Now
1. It chooses best from plans with expected distribution
2. It doesn't account distribution for single, because in single shuffle never required.
3. It does return matched plan when cost of shuffle + best is larger than matched


axiom/optimizer/Plan.cpp
```diff
@@ -1838,80 +1844,91 @@ PlanP Optimization::makePlan(
...
-    if (isSingleWorker_) {
-      RelationOpPtr result = make<UnionAll>(inputs);
-      Aggregation* distinct = nullptr;
-      if (isDistinct) {
-        result = makeDistinct(result);
-        distinct = result->as<Aggregation>();
-      }
-      return unionPlan(inputStates, inputPlans, result, distinct);
-    }
+    return makeUnionPlan(
```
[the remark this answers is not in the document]
MBkkt: > NOTE: I recommend to disable whitespace diff when you will review this PR

I only put this union construction code in a separate function

axiom/optimizer/Plan.cpp
```diff
@@ -323,28 +322,34 @@ PlanP PlanSet::best(const Distribution& distribution, bool& needsShuffle) {
...
+    auto update = [&](PlanP& current, float& currentCost) {
+      if (!current || cost < currentCost) {
+        current = plan.get();
+        currentCost = cost;
+      }
+    };
+
+    update(best, bestCost);
+    if (!single && plan->op->distribution().isSamePartition(distribution)) {
+      update(match, matchCost);
```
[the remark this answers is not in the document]
MBkkt: > Is this an actual fix?

I made three fixes (maybe 2, I'm not sure about old code behavior in case of single):
This is the first.
The second: https://github.com/facebookexperimental/verax/pull/214#discussion_r2284147241
The third we don't account "same distribution" in single mode

> Now we keep lowest-cost plan with desired distribution. Do I get it right?

Yes, this is true.

axiom/optimizer/Plan.cpp
```diff
@@ -323,28 +322,34 @@ PlanP PlanSet::best(const Distribution& distribution, bool& needsShuffle) {
...
 
   VELOX_DCHECK_NOT_NULL(best);
 
-  if (best != match && match) {
-    const float shuffle = shuffleCost(best->op->columns()) * best->cost.fanout;
-    if (bestCost + shuffle < matchCost) {
-      needsShuffle = true;
+  if (single || best == match) {
+    return best;
+  }
```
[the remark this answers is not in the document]
mbasmanova: @MBkkt Got it. Thank you for detailed explanation.

axiom/optimizer/Plan.cpp
```diff
@@ -1567,6 +1572,7 @@ void Optimization::placeDerivedTable(DerivedTableCP from, PlanState& state) {
   if (reduction < 0.9) {
     key.tables = reducingSet;
     key.columns = state.downstreamColumns();
+    ignore = false;
```
mbasmanova, line 1575: It might be more robust to not rely on the caller to provide the right default for the output parameter, but rather set it first thing in the implementation of makePlan.

axiom/optimizer/Plan.cpp
```diff
@@ -323,28 +322,34 @@ PlanP PlanSet::best(const Distribution& distribution, bool& needsShuffle) {
...
     }
   }
 
   VELOX_DCHECK_NOT_NULL(best);
 
-  if (best != match && match) {
-    const float shuffle = shuffleCost(best->op->columns()) * best->cost.fanout;
-    if (bestCost + shuffle < matchCost) {
-      needsShuffle = true;
+  if (single || best == match) {
```
mbasmanova, line 341: I assume we don't need to check 'single' here. if (best == match) should be sufficient.

mbasmanova:
> I tried to explain it, but please write to me, if it's not clear, maybe I'm wrong in some assumptions.

It is clear now. Appreciate detailed explanations. Next time, it would be helpful to put these in the PR description. That will make review easier and faster. Thanks!

I see that there is not easy way to write tests for this logic now. We don't need to block this fix on that, but let's think about how to reorganized the code to allow such testing. CC: @oerling

axiom/optimizer/DerivedTable.cpp
```diff
@@ -1022,9 +1022,8 @@ void DerivedTable::makeInitialPlan() {
 
   optimization->makeJoins(nullptr, state);
 
-  Distribution emptyDistribution;
-  bool needsShuffle;
-  auto plan = state.plans.best(emptyDistribution, needsShuffle)->op;
+  bool ignore = false;
+  auto plan = state.plans.best({}, ignore)->op;
```
mbasmanova, line 1026: I see multiple places that call "best" with empty distribution and needsShuffle that's ignored. Let's refactor (in a separate PR) to introduce a version of "best" that takes no arguments. This will make the calling code simpler.

axiom/optimizer/Plan.cpp
MBkkt: Sounds good to me, I will replace dcheck(!needsShuffle) in makePlan with `needsShuffle = false;`

PR #216, opened 2025-08-18
feat(optimizer): Add support for Unnest

MBkkt:
I don't think it's perfect, but at least for me it's better than nothing.

Unnest modeled as Join (this commit: https://github.com/facebookexperimental/verax/pull/216/commits/d49333ae70d0a8ce6e8bb9f313f95fa7c2b1fab0) inspired by this PR https://github.com/facebookexperimental/verax/pull/260

It also fix this issue: https://github.com/facebookexperimental/verax/issues/255 (It doesn't allow unnest to be leaf node)

There's some limitations
1. I don't compute cardinality of unnested columns, just assume it's same as input array counts for now (same as every array have length 1)
2. I don't support ordinality column for Unnest in optimizer, because it needs more code and tests, so I decided someone who need this can support it, it's not needed for us for now.
3. Joins cannot be mixed with Unnest right now, there's tests for this and issue to fix this https://github.com/facebookexperimental/verax/issues/286

Test coverage:
```
  // I test following cases:
  //  If something is after unnest
  //   it can depend and not depend on unnested columns.
  //  And we should also check that any expressions are allowed inside unnest,
  //   not only input column references.
  // 1. unnest (with constant and map mixed with array)
  // 2. unnest after unnest
  // 3. project before and after unnest
  // 4. filter before and after unnest
  // 5. group by before and after unnest
  // 6. order by before and after unnest
  // 7. limit before and after unnest
  // 8. join before and after unnest
  // 9. there's no extra columns in projections before unnest
```

### DIFF
--- a/axiom/optimizer/tests/UnnestTest.cpp
+++ b/axiom/optimizer/tests/UnnestTest.cpp
@@ -30,10 +30,9 @@ class UnnestTest : public PlanTest {
   void SetUp() override {
     PlanTest::SetUp();
 
-    dummyRow = std::make_shared<RowVector>(
-        pool_.get(), ROW({}), BufferPtr{}, 1, std::vector<VectorPtr>{});
+    dummyRow_ = makeRowVector(ROW({}), 1);
 
-    rowVector = makeRowVector(
+    rowVector_ = makeRowVector(
         names,
         {
             makeFlatVector<int64_t>({
@@ -61,14 +60,14 @@ class UnnestTest : public PlanTest {
   }
 
   void TearDown() override {
-    rowVector.reset();
-    dummyRow.reset();
+    rowVector_.reset();
+    dummyRow_.reset();
     PlanTest::TearDown();
   }
 
   const std::vector<std::string> names{"x", "a_a_y", "a_a_z"};
-  RowVectorPtr dummyRow;
-  RowVectorPtr rowVector;
+  RowVectorPtr dummyRow_;
+  RowVectorPtr rowVector_;
 };
 
 // We need to test the following cases:
@@ -93,7 +92,7 @@ TEST_F(UnnestTest, unnest) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -109,7 +108,7 @@ TEST_F(UnnestTest, unnest) {
 
     auto referencePlanUnnest =
         exec::test::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .project({
                 "x",
                 "a_a_y",
@@ -127,7 +126,7 @@ TEST_F(UnnestTest, unnest) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -152,7 +151,7 @@ TEST_F(UnnestTest, unnest) {
 
     auto referencePlanUnnest =
         exec::test::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .project({
                 "x",
                 "a_a_y",
@@ -184,7 +183,7 @@ TEST_F(UnnestTest, unnest) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -215,7 +214,7 @@ TEST_F(UnnestTest, unnest) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
             })
@@ -231,7 +230,7 @@ TEST_F(UnnestTest, unnest) {
 
     auto referencePlanUnnest =
         exec::test::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .project({"array_distinct(a_a_y) AS a_a_y_d"})
             .unnest({}, {"a_a_y_d"})
             .planNode();
@@ -249,7 +248,7 @@ TEST_F(UnnestTest, unnest) {
                                  .build();
 
     auto referencePlanUnnest = exec::test::PlanBuilder{pool_.get()}
-                                   .values({dummyRow})
+                                   .values({dummyRow_})
                                    .project({
                                        "array[1, 2, 3] AS a",
                                    })
@@ -283,7 +282,7 @@ TEST_F(UnnestTest, unnest) {
 
     auto referencePlanUnnest =
         exec::test::PlanBuilder{pool_.get()}
-            .values({dummyRow})
+            .values({dummyRow_})
             .project({
                 "array[1, 2, 3] AS a",
                 "map(array['1', '2'], array[10, 20]) AS m",
@@ -315,7 +314,7 @@ TEST_F(UnnestTest, project) {
     SCOPED_TRACE("project before unnest");
 
     auto logicalPlanUnnest = lp::PlanBuilder{}
-                                 .values({rowVector})
+                                 .values({rowVector_})
                                  .project({
                                      "x + 1 AS x1",
                                      "array_distinct(a_a_y) AS a_a_y_d",
@@ -342,7 +341,7 @@ TEST_F(UnnestTest, project) {
 
     auto referencePlanUnnest =
         exec::test::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .project({
                 "x + 1 AS x1",
                 "array_distinct(a_a_y) AS a_a_y_d",
@@ -358,7 +357,7 @@ TEST_F(UnnestTest, project) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -379,7 +378,7 @@ TEST_F(UnnestTest, project) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .project({
                                        "x",
                                        "array_distinct(a_a_y) AS a_a_y_d",
@@ -396,7 +395,7 @@ TEST_F(UnnestTest, project) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -417,7 +416,7 @@ TEST_F(UnnestTest, project) {
 
     auto referencePlanUnnest =
         exec::test::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .project({
                 "x",
                 "array_distinct(a_a_y) AS a_a_y_d",
@@ -437,7 +436,7 @@ TEST_F(UnnestTest, filter) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .filter("x % 2 = 0")
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
@@ -459,7 +458,7 @@ TEST_F(UnnestTest, filter) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .filter("x % 2 = 0")
                                    .project({
                                        "x",
@@ -477,7 +476,7 @@ TEST_F(UnnestTest, filter) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -499,7 +498,7 @@ TEST_F(UnnestTest, filter) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .project({
                                        "x",
                                        "array_distinct(a_a_y) AS a_a_y_d",
@@ -517,7 +516,7 @@ TEST_F(UnnestTest, filter) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -545,7 +544,7 @@ TEST_F(UnnestTest, filter) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .project({
                                        "x",
                                        "array_distinct(a_a_y) AS a_a_y_d",
@@ -569,7 +568,7 @@ TEST_F(UnnestTest, filter) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -597,7 +596,7 @@ TEST_F(UnnestTest, filter) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .project({
                                        "x",
                                        "array_distinct(a_a_y) AS a_a_y_d",
@@ -621,7 +620,7 @@ TEST_F(UnnestTest, filter) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -649,7 +648,7 @@ TEST_F(UnnestTest, filter) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .project({
                                        "x",
                                        "array_distinct(a_a_y) AS a_a_y_d",
@@ -676,7 +675,7 @@ TEST_F(UnnestTest, groupBy) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .aggregate(names, {})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
@@ -697,7 +696,7 @@ TEST_F(UnnestTest, groupBy) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .singleAggregation(names, {})
                                    .project({
                                        "x",
@@ -714,7 +713,7 @@ TEST_F(UnnestTest, groupBy) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -735,7 +734,7 @@ TEST_F(UnnestTest, groupBy) {
 
     auto referencePlanUnnest =
         exec::test::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .project({
                 "x",
                 "array_distinct(a_a_y) AS a_a_y_d",
@@ -755,7 +754,7 @@ TEST_F(UnnestTest, orderBy) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .orderBy(names)
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
@@ -776,7 +775,7 @@ TEST_F(UnnestTest, orderBy) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .orderBy(names, {})
                                    .project({
                                        "x",
@@ -793,7 +792,7 @@ TEST_F(UnnestTest, orderBy) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -815,7 +814,7 @@ TEST_F(UnnestTest, orderBy) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .project({
                                        "x",
                                        "array_distinct(a_a_y) AS a_a_y_d",
@@ -832,7 +831,7 @@ TEST_F(UnnestTest, orderBy) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -855,7 +854,7 @@ TEST_F(UnnestTest, orderBy) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .project({
                                        "x",
                                        "array_distinct(a_a_y) AS a_a_y_d",
@@ -875,7 +874,7 @@ TEST_F(UnnestTest, limit) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .limit(1, 1)
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
@@ -896,7 +895,7 @@ TEST_F(UnnestTest, limit) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .limit(1, 1, {})
                                    .project({
                                        "x",
@@ -913,7 +912,7 @@ TEST_F(UnnestTest, limit) {
 
     auto logicalPlanUnnest =
         lp::PlanBuilder{}
-            .values({rowVector})
+            .values({rowVector_})
             .unnest({
                 lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                 lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
@@ -934,7 +933,7 @@ TEST_F(UnnestTest, limit) {
     }
 
     auto referencePlanUnnest = exec::test::PlanBuilder{}
-                                   .values({rowVector})
+                                   .values({rowVector_})
                                    .project({
                                        "x",
                                        "array_distinct(a_a_y) AS a_a_y_d",
@@ -957,11 +956,11 @@ TEST_F(UnnestTest, join) {
     lp::PlanBuilder::Context ctx;
     auto logicalPlanUnnest =
         lp::PlanBuilder{ctx}
-            .values({rowVector})
+            .values({rowVector_})
             .project({"x AS x1", "a_a_y AS a_a_y1", "a_a_z AS a_a_z1"})
             .join(
                 lp::PlanBuilder{ctx}
-                    .values({rowVector})
+                    .values({rowVector_})
                     .project({"x AS x2", "a_a_y AS a_a_y2", "a_a_z AS a_a_z2"}),
                 "x1 = x2",
                 lp::JoinType::kInner)
@@ -975,19 +974,57 @@ TEST_F(UnnestTest, join) {
             })
             .project(expectedNames)
             .build();
-    auto plan = toSingleNodePlan(logicalPlanUnnest, 1);
+    {
+      auto plan = toSingleNodePlan(logicalPlanUnnest, 1);
 
-    auto matcher = core::PlanMatcherBuilder{}
-                       .values()
-                       .hashJoin(core::PlanMatcherBuilder{}.values().build())
-                       .project()
-                       .unnest()
-                       .project()
-                       .unnest()
-                       .project()
-                       .build();
-    ASSERT_TRUE(matcher->match(plan)) << plan->toString(true, true);
-    ASSERT_EQ(plan->outputType()->names(), expectedNames);
+      auto matcher = core::PlanMatcherBuilder{}
+                         .values()
+                         .hashJoin(core::PlanMatcherBuilder{}.values().build())
+                         .project()
+                         .unnest()
+                         .project()
+                         .unnest()
+                         .project()
+                         .build();
+      ASSERT_TRUE(matcher->match(plan)) << plan->toString(true, true);
+      ASSERT_EQ(plan->outputType()->names(), expectedNames);
+    }
+
+    auto generator = std::make_shared<core::PlanNodeIdGenerator>();
+    auto referencePlanUnnest =
+        exec::test::PlanBuilder{generator}
+            .values({rowVector_})
+            .project({"x AS x1", "a_a_y AS a_a_y1", "a_a_z AS a_a_z1"})
+            .hashJoin(
+                {"x1"},
+                {"x2"},
+                exec::test::PlanBuilder{generator}
+                    .values({rowVector_})
+                    .project({"x AS x2", "a_a_y AS a_a_y2", "a_a_z AS a_a_z2"})
+                    .planNode(),
+                "",
+                {"x1", "a_a_y1", "a_a_z1", "x2", "a_a_y2", "a_a_z2"})
+            .project({
+                "x1",
+                "a_a_y2",
+                "a_a_z2",
+                "array_distinct(a_a_y1) AS a_a_y1_d",
+                "array_distinct(a_a_z1) AS a_a_z1_d",
+            })
+            .project({
+                "x1",
+                "a_a_y1_d",
+                "a_a_z1_d",
+                "array_distinct(a_a_y2) AS a_a_y2_d",
+                "array_distinct(a_a_z2) AS a_a_z2_d",
+            })
+            .unnest({"x1", "a_a_y2_d", "a_a_z2_d"}, {"a_a_y1_d", "a_a_z1_d"})
+            .unnest(
+                {"x1", "a_a_y1_d_e", "a_a_z1_d_e"}, {"a_a_y2_d", "a_a_z2_d"})
+            .project({"x1", "a_a_y1_d_e AS a_y1", "a_a_z2_d_e AS a_z2"})
+            .planNode();
+
+    checkSame(logicalPlanUnnest, referencePlanUnnest);
   }
   {
     SCOPED_TRACE("join before unnest (dependent on unnested columns)");
@@ -997,11 +1034,11 @@ TEST_F(UnnestTest, join) {
     lp::PlanBuilder::Context ctx;
     auto logicalPlanUnnest =
         lp::PlanBuilder{ctx}
-            .values({rowVector})
+            .values({rowVector_})
             .project({"x AS x1", "a_a_y AS a_a_y1", "a_a_z AS a_a_z1"})
             .join(
                 lp::PlanBuilder{ctx}
-                    .values({rowVector})
+                    .values({rowVector_})
                     .project({"x AS x2", "a_a_y AS a_a_y2", "a_a_z AS a_a_z2"}),
                 "a_a_y1 = a_a_y2",
                 lp::JoinType::kInner)
@@ -1015,19 +1052,57 @@ TEST_F(UnnestTest, join) {
             })
             .project(expectedNames)
             .build();
-    auto plan = toSingleNodePlan(logicalPlanUnnest, 1);
+    {
+      auto plan = toSingleNodePlan(logicalPlanUnnest, 1);
 
-    auto matcher = core::PlanMatcherBuilder{}
-                       .values()
-                       .hashJoin(core::PlanMatcherBuilder{}.values().build())
-                       .project()
-                       .unnest()
-                       .project()
-                       .unnest()
-                       .project()
-                       .build();
-    ASSERT_TRUE(matcher->match(plan)) << plan->toString(true, true);
-    ASSERT_EQ(plan->outputType()->names(), expectedNames);
+      auto matcher = core::PlanMatcherBuilder{}
+                         .values()
+                         .hashJoin(core::PlanMatcherBuilder{}.values().build())
+                         .project()
+                         .unnest()
+                         .project()
+                         .unnest()
+                         .project()
+                         .build();
+      ASSERT_TRUE(matcher->match(plan)) << plan->toString(true, true);
+      ASSERT_EQ(plan->outputType()->names(), expectedNames);
+    }
+
+    auto generator = std::make_shared<core::PlanNodeIdGenerator>();
+    auto referencePlanUnnest =
+        exec::test::PlanBuilder{generator}
+            .values({rowVector_})
+            .project({"x AS x1", "a_a_y AS a_a_y1", "a_a_z AS a_a_z1"})
+            .hashJoin(
+                {"a_a_y1"},
+                {"a_a_y2"},
+                exec::test::PlanBuilder{generator}
+                    .values({rowVector_})
+                    .project({"x AS x2", "a_a_y AS a_a_y2", "a_a_z AS a_a_z2"})
+                    .planNode(),
+                "",
+                {"x1", "a_a_y1", "a_a_z1", "x2", "a_a_y2", "a_a_z2"})
+            .project({
+                "x1",
+                "a_a_y2",
+                "a_a_z2",
+                "array_distinct(a_a_y1) AS a_a_y1_d",
+                "array_distinct(a_a_z1) AS a_a_z1_d",
+            })
+            .project({
+                "x1",
+                "a_a_y1_d",
+                "a_a_z1_d",
+                "array_distinct(a_a_y2) AS a_a_y2_d",
+                "array_distinct(a_a_z2) AS a_a_z2_d",
+            })
+            .unnest({"x1", "a_a_y2_d", "a_a_z2_d"}, {"a_a_y1_d", "a_a_z1_d"})
+            .unnest(
+                {"x1", "a_a_y1_d_e", "a_a_z1_d_e"}, {"a_a_y2_d", "a_a_z2_d"})
+            .project({"x1", "a_a_y1_d_e AS a_y1", "a_a_z2_d_e AS a_z2"})
+            .planNode();
+
+    checkSame(logicalPlanUnnest, referencePlanUnnest);
   }
   {
     SCOPED_TRACE("join after unnest (independent on unnested columns)");
@@ -1037,7 +1112,7 @@ TEST_F(UnnestTest, join) {
     lp::PlanBuilder::Context ctx;
     auto logicalPlanUnnest =
         lp::PlanBuilder{ctx}
-            .values({rowVector})
+            .values({rowVector_})
             .project({"x AS x1", "a_a_y AS a_a_y1", "a_a_z AS a_a_z1"})
             .unnest({
                 lp::Sql("array_distinct(a_a_y1)").unnestAs("a_y1"),
@@ -1045,7 +1120,7 @@ TEST_F(UnnestTest, join) {
             })
             .join(
                 lp::PlanBuilder{ctx}
-                    .values({rowVector})
+                    .values({rowVector_})
                     .project({"x AS x2", "a_a_y AS a_a_y2", "a_a_z AS a_a_z2"})
                     .unnest({
                         lp::Sql("array_distinct(a_a_y2)").unnestAs("a_y2"),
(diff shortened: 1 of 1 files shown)


axiom/optimizer/QueryGraph.h
```diff
@@ -879,4 +879,26 @@ class AggregationPlan : public PlanObject {
 
 using AggregationPlanCP = const AggregationPlan*;
 
+class UnnestPlan : public PlanObject {
+ public:
+  UnnestPlan(ExprVector unnestExprs, ColumnVector unnestedColumns)
+      : PlanObject{PlanType::kUnnestNode},
+        unnestExprs_{std::move(unnestExprs)},
```
mbasmanova, line 886: Are we missing ordinalityColumn?

mbasmanova (commented):
@MBkkt Thank you for working on adding support for Unnest. Some comments and questions.

axiom/optimizer/Subfields.cpp
```diff
@@ -70,6 +70,36 @@ void ToGraph::markFieldAccessed(
       return;
     }
 
+    if (kind == lp::NodeKind::kUnnest) {
+      const auto* unnest = source.planNode->asUnchecked<lp::UnnestNode>();
+      const auto& input = unnest->onlyInput();
+      if (ordinal < input->outputType()->size()) {
```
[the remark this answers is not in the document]
MBkkt: Well, partially yes, if I didn't write, it will crash/exception in my test

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1054,6 +1054,48 @@ TEST_F(PlanTest, values) {
...
+    auto makeLogicalPlan = [&] {
+      return lp::PlanBuilder(ctx).values({rowVector});
+    };
+
+    auto logicalPlanUnnest = makeLogicalPlan()
+                                 .unnest({"regions"}, {{"region"}})
+                                 .project({"nation", "region"})
+                                 .build();
+
+    auto referencePlanUnnest = exec::test::PlanBuilder(pool_.get())
```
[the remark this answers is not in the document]
MBkkt: ok

axiom/optimizer/ToGraph.cpp
```diff
@@ -1624,7 +1661,25 @@ PlanObjectP ToGraph::makeQueryGraph(
...
     }
-    case lp::NodeKind::kUnnest:
+
+    case lp::NodeKind::kUnnest: {
+      if (!contains(allowedInDt, PlanType::kUnnestNode)) {
+        return wrapInDt(node);
+      }
+
+      // Multiple unnest is allowed in a DT.
+      // If arrives after groupBy, orderBy, limit, then starts a new DT.
```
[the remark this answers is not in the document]
MBkkt: Ok

axiom/optimizer/ToVelox.h
```diff
@@ -127,6 +127,12 @@ class ToVelox {
       velox::core::PlanNodePtr source,
       std::vector<velox::core::FieldAccessTypedExprPtr>& result);
 
+  // Makes a Velox UnnestNode for a RelationOp.
+  core::PlanNodePtr makeUnnest(
+      Unnest& op,
```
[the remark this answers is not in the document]
MBkkt: Ok, but I will do it for other methods too

axiom/optimizer/ToGraph.cpp
```diff
@@ -844,6 +846,40 @@ ExprVector ToGraph::translateColumns(const std::vector<lp::ExprPtr>& source) {
...
+    channel -= replicateType.size();
+    for (size_t i = 0; const auto& name : logicalUnnest.unnestedNames()) {
+      if (channel >= name.size()) {
+        channel -= name.size();
+        ++i;
+        continue;
+      }
+      auto expr = logicalUnnest.unnestExpressions()[i];
+      const auto& inputName =
+          expr->asUnchecked<lp::InputReferenceExpr>()->name();
```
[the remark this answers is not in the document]
MBkkt: But I will add it here too

axiom/optimizer/ToVelox.cpp
```diff
@@ -1134,6 +1134,54 @@ velox::core::PlanNodePtr ToVelox::makeJoin(
...
 }
 
+core::PlanNodePtr ToVelox::makeUnnest(
+    Unnest& op,
+    ExecutableFragment& fragment,
+    std::vector<ExecutableFragment>& stages) {
+  auto input = makeFragment(op.input(), fragment, stages);
+
+  // TODO: This code is suboptimal, because unfortunately currently in optimizer
+  // we don't have ability to check that column referenced only for current
```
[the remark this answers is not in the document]
MBkkt: Yes, but I thought we cannot use it in `ToVelox`.
But yes, my idea was about use downstreamColumns somewhere here
```cpp
void Optimization::addPostprocess(
    DerivedTableCP dt,
    RelationOpPtr& plan,
    PlanState& state) {
  for (auto* unnest : dt->unnests) {
    auto* unnestOp =
        make<Unnest>(plan, unnest->unnestExprs(), unnest->unnestedColumns());
    state.addCost(*unnestOp);
    plan = unnestOp;
  }

```

axiom/logical_plan/PlanBuilder.cpp
```diff
@@ -322,6 +318,51 @@ PlanBuilder& PlanBuilder::with(const std::vector<ExprApi>& projections) {
...
+
+  std::vector<ExprPtr> outputUnnestExpressions;
+  std::vector<std::vector<std::string>> outputUnnestNames;
+
+  for (size_t i = 0; const auto& untypedExpr : unnestExpressions) {
+    auto expr = resolveScalarTypes(untypedExpr.expr());
+    VELOX_USER_CHECK(expr->isInputReference());
+    const auto& id = expr->asUnchecked<InputReferenceExpr>()->name();
+    const auto& unnestName = unnestNames[i++];
+    VELOX_USER_CHECK(!unnestName.empty());
```
mbasmanova, line 346: It would be nice to not require names and auto-generate some if not specified. This would be consistent with how other APIs work (project, aggregate, etc.).

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1054,6 +1054,48 @@ TEST_F(PlanTest, values) {
...
+
+  lp::PlanBuilder::Context ctx{exec::test::kHiveConnectorId};
+
+  {
+    auto makeLogicalPlan = [&] {
+      return lp::PlanBuilder(ctx).values({rowVector});
+    };
+
+    auto logicalPlanUnnest = makeLogicalPlan()
+                                 .unnest({"regions"}, {{"region"}})
```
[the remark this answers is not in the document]
mbasmanova: We'll be writing and maintaining a lot of tests. Hence, having a convenient concise way for writing these is important for long-term health of the project.

axiom/optimizer/ToVelox.cpp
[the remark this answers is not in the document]
mbasmanova: > I thought we cannot use it in ToVelox.

I expect we would use it in the core optimizer (Plan.cpp) and indicate what columns to project out of Unnest in the Unnest RelationOp.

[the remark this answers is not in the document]
MBkkt: > indicate what columns to project out of Unnest in the Unnest RelationOp.

> somewhere here

It's same place, ok

axiom/optimizer/tests/PlanTest.cpp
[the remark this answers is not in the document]
MBkkt: I made autogenerated strings, so users have two options:
1. pass empty array => names will be autogenerated
2. pass array of names

I think it should be enough to write tests.
Other options, like "AS some_name" I think it's unnecessary more complicated because they require to parse strings and then parse resulting ast.

axiom/optimizer/ToVelox.cpp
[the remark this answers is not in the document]
MBkkt: Fixed

axiom/optimizer/QueryGraph.h
MBkkt: Added VELOX_USER_CHECK in translateUnnest

axiom/logical_plan/PlanBuilder.cpp
```diff
@@ -322,6 +318,51 @@ PlanBuilder& PlanBuilder::with(const std::vector<ExprApi>& projections) {
...
+      newOutputMapping->add(name, id);
+    }
+  }
+
+  std::vector<ExprPtr> outputUnnestExpressions;
+  std::vector<std::vector<std::string>> outputUnnestNames;
+
+  for (size_t i = 0; const auto& untypedExpr : unnestExpressions) {
+    auto expr = resolveScalarTypes(untypedExpr.expr());
+    VELOX_USER_CHECK(expr->isInputReference());
```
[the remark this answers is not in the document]
MBkkt: Fixed

axiom/optimizer/ToGraph.cpp
```diff
@@ -844,6 +846,40 @@ ExprVector ToGraph::translateColumns(const std::vector<lp::ExprPtr>& source) {
   return result;
 }
 
+void ToGraph::translateUnnest(const lp::UnnestNode& logicalUnnest) {
+  const auto& replicateType = *logicalUnnest.onlyInput()->outputType();
+  ExprVector unnestExprs;
+  ColumnVector unnestedColumns;
+  for (auto channel : usedChannels(logicalUnnest)) {
```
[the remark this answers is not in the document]
MBkkt: Fixed this

MBkkt:
@mbasmanova I fixed most of your comments and answered on two other

axiom/optimizer/QueryGraph.h
```diff
@@ -920,4 +920,26 @@ class AggregationPlan : public PlanObject {
 
 using AggregationPlanCP = const AggregationPlan*;
 
+class UnnestPlan : public PlanObject {
+ public:
+  UnnestPlan(ExprVector unnestExprs, ColumnVector unnestedColumns)
```
mbasmanova, line 925: What's the relationship between unnestExprs and unnestedColumns? Is it that each expr has one or more columns (dependending on type)? one column per array, 2 per map?

Would it be possible to add a comment to clarify?

Should we add some checks in this ctor to ensure compatibility between unnestExprs and unnestedColumns?

naming: can we drop 'unnested' since this is implied from the class name? exprs & columns

axiom/optimizer/RelationOp.cpp
```diff
@@ -504,6 +513,16 @@ Aggregation::Aggregation(
   cost_.totalBytes = nOut * rowBytes;
 }
 
+std::string Unnest::toString(bool recursive, bool detail) const {
+  std::stringstream out;
+  if (recursive) {
+    out << input()->toString(true, detail) << " ";
+  }
+  out << "unnest ";
```
mbasmanova, line 521: Should we include details such as replicated columns and unnested columns?

axiom/optimizer/Subfields.cpp
```diff
@@ -41,120 +41,191 @@ PathCP stepsToPath(const std::vector<Step>& steps) {
...
+  markSubfields(
+      project.expressionAt(ordinal),
+      steps,
+      isControl,
+      std::array{input->outputType().get()},
+      std::array{LogicalContextSource{.planNode = input.get()}});
+}
+
+void ToGraph::markFieldAccessed(
+    const lp::UnnestNode& unnest,
```
mbasmanova, line 77: Do we have test coverage for this logic?

axiom/optimizer/ToGraph.cpp
```diff
@@ -829,6 +831,45 @@ ExprVector ToGraph::translateColumns(const std::vector<lp::ExprPtr>& source) {
   return result;
 }
 
+void ToGraph::translateUnnest(const lp::UnnestNode& logicalUnnest) {
+  if (logicalUnnest.ordinalityName().has_value()) {
+    VELOX_NYI(
+        "Unnest ordinality column is not supported in Verax optimizer. Unnest node: {}",
```
mbasmanova, line 837: Let's create GitHub issue to add this support.

axiom/optimizer/ToGraph.cpp
```diff
@@ -829,6 +831,45 @@ ExprVector ToGraph::translateColumns(const std::vector<lp::ExprPtr>& source) {
...
+  }
+  const auto& replicateType = *logicalUnnest.onlyInput()->outputType();
+  ExprVector unnestExprs;
+  ColumnVector unnestedColumns;
+  for (auto channel : usedChannels(logicalUnnest)) {
+    if (channel < replicateType.size()) {
+      continue;
+    }
+    channel -= replicateType.size();
+    for (size_t i = 0; const auto& name : logicalUnnest.unnestedNames()) {
```
mbasmanova, line 848: name -> names

otherwise, it is confusing as it reads that we compare column index with a length of a name string

axiom/optimizer/ToGraph.cpp
```diff
@@ -829,6 +831,45 @@ ExprVector ToGraph::translateColumns(const std::vector<lp::ExprPtr>& source) {
...
+      if (channel >= name.size()) {
+        channel -= name.size();
+        ++i;
+        continue;
+      }
+      const auto* expr = translateExpr(logicalUnnest.unnestExpressions()[i]);
+      const auto& outputName = name[channel];
+      // TODO: Cardinality here should be multiply input column cardinality by
```
mbasmanova, line 856: typo? multiply -> a multiple of

axiom/optimizer/ToGraph.h
```diff
@@ -370,25 +372,60 @@ class ToGraph {
...
+
+  void markFieldAccessed(
+      const logical_plan::UnnestNode& unnest,
+      int32_t ordinal,
+      std::vector<Step>& steps,
+      bool isControl,
+      std::span<const RowType* const> context,
+      std::span<const LogicalContextSource> sources);
+
+  void markFieldAccessed(
```
mbasmanova, line 410: Thank you for breaking up a long method into smaller ones.

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1148,6 +1148,130 @@ TEST_F(PlanTest, values) {
     checkSame(logicalPlan, referencePlan);
   }
 }
+
+TEST_F(PlanTest, unnest) {
```
mbasmanova, line 1152: Do we ensure that column pruning happens properly? I assume not. Would it be possible to check that?

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1148,6 +1148,130 @@ TEST_F(PlanTest, values) {
...
+  {
+    auto logicalPlanUnnest =
+        lp::PlanBuilder{}.values({rowVector}).unnest({"regions"}).build();
+
+    auto referencePlanUnnest = exec::test::PlanBuilder{}
+                                   .values({rowVector})
+                                   .unnest({"nation", "regions"}, {"regions"})
+                                   .planNode();
+
+    checkSame(logicalPlanUnnest, referencePlanUnnest);
```
mbasmanova, line 1182: This ensures that the plan produces correct results, but doesn't verify that the plan is "optimal". Is that so? Would it be possible to verify the plan itself as well?

mbasmanova (commented):
@MBkkt Thank you for iterating. Some follow-up questions.

axiom/optimizer/Subfields.cpp
MBkkt: 1) code is covered, you can attach debugger and run my test, it will go here in both branches
2) but subfields push down isn't covered, but it's same for a lot of other code

axiom/optimizer/tests/PlanTest.cpp
MBkkt: Ok

axiom/optimizer/ToGraph.cpp
MBkkt: Maybe bad grammar, not typo. I meant that cardinality for this Value should be input array column cardinality * expected array len

axiom/optimizer/QueryGraph.h
MBkkt: > What's the relationship between unnestExprs and unnestedColumns? Is it that each expr has one or more columns (dependending on type)? one column per array, 2 per map?
Would it be possible to add a comment to clarify?
Should we add some checks in this ctor to ensure compatibility between unnestExprs and unnestedColumns?

Added checks with message



> naming: can we drop 'unnested' since this is implied from the class name? exprs & columns

I don't want this. I understand that this node, doesn't have other fields now, but it's still feels very inconsistent in case of renaming.
Lets keep it consistent with logical_plan, RelationOp, execution nodes.

axiom/optimizer/tests/PlanTest.cpp
MBkkt: Added

axiom/optimizer/tests/PlanTest.cpp
MBkkt: Added

axiom/optimizer/RelationOp.cpp
MBkkt: Done

axiom/logical_plan/PlanBuilder.h
```diff
@@ -224,6 +226,24 @@ class PlanBuilder {
     return with(std::vector<ExprApi>{projections});
   }
 
+  PlanBuilder& unnest(
```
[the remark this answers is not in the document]
MBkkt: Cool!

I'm ok with almost any API for this, but if possible I want minimize time that I need to spend to develop this API. 
Because as I mentioned I don't see the big difference in how tests should be written and we don't use PlanBuilder in our system.

If mentioned PR will be merged before this PR I will do rebase ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯

axiom/optimizer/ToGraph.cpp
```diff
@@ -1611,7 +1657,25 @@ PlanObjectP ToGraph::makeQueryGraph(
...
-    case lp::NodeKind::kUnnest:
+
+    case lp::NodeKind::kUnnest: {
+      if (!contains(allowedInDt, PlanType::kUnnestNode)) {
+        return wrapInDt(node);
+      }
+
+      // Multiple unnest is allowed in a DT.
+      // If arrives after groupBy, orderBy, limit, then starts a new DT.
+      makeQueryGraph(*node.onlyInput(), allowedInDt);
```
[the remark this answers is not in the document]
MBkkt: I don't expect unnest will be leaf node, it's just not convenient (a lot of unnecessary conditions).

It's same as projection node is never leaf node.

We can have dummy values node for this 
https://github.com/facebookexperimental/verax/blob/f7d41fe040ccbc003c67a097b91321ca17d57b1f/axiom/optimizer/tests/QuerySqlParser.cpp#L189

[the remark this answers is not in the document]
MBkkt: 1. `SELECT ARRAY[1, 2, 3]` -- values => project
2. `SELECT UNNEST(ARRAY[1, 2, 3])` -- values => "temporary" project => unnest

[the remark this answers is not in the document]
MBkkt: Also in velox both project and unnest nodes are not leaf node.
And IMO leaf nodes are more complicated in optimizer (a lot of special cases for Values/Base/DerivedTable nodes), so I think better to don't make unnest leaf node.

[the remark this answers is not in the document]
MBkkt: If it's necessary to write tests, we can add this dummy values node in PlanBuilder

mbasmanova:
I thought about the unnest some more and I believe it needs to be handled / modeled as a join and it needs to participate in the join order. It is not part of post-processing.

For example, consider (t + unnest(foo)) join (u + unnest(bar)) on t.k = u.k. If "t join u on t.k = u.k" is selective, then optimal plan would join t with u firs, then apply 2 unnests.

MBkkt:
@mbasmanova 

> I thought about the unnest some more and I believe it needs to be handled / modeled as a join and it needs to participate in the join order. It is not part of post-processing.

I don't fully agree, because right now we don't have cardinality of array inside row.
So it's difficult to account unnest in desired join order.

Another part is join pushdown, but I think it should be done not only for unnest, but for other post process nodes too.
So it can be done separately.

> For example, consider (t + unnest(foo)) join (u + unnest(bar)) on t.k = u.k. If "t join u on t.k = u.k" is selective, then optimal plan would join t with u first, then apply 2 unnests.

I don't understand example query, can you provide it in more standard sql way?
Also "current" and "desired" plans will be helpful.


Another thing that I don't understand, this PR adds initial implementation, initial tests and doesn't change unnest interface (although to me it's desired).
So I don't see the real reason why does opportunity to make it more optimal internally should prevent this PR from merge.
If someone later will remove this entire code (but keep tests working) I'm completely ok with it.
But right now I cannot write tests on some part of stateless execution because unnest isn't supported at all.
With this PR verax get this initial support of unnest that can be changed when someone will have time for this.
What am I missing?

MBkkt:
@mbasmanova 

Did you mean query like this?

```cpp
    lp::PlanBuilder::Context ctx;
    auto logicalPlanUnnest =
        lp::PlanBuilder{ctx}
            .values({rowVector})
            .unnest({"array_distinct(regions)"}, {{"distinct_regions"}})
            .unnest({"array_distinct(distinct_regions)"}, {{"distinct_region"}})
            .project({"nation AS nation1", "distinct_region AS region1"})
            .join(
                lp::PlanBuilder{ctx}
                    .values({rowVector})
                    .unnest({"array_distinct(regions)"}, {{"distinct_regions"}})
                    .unnest(
                        {"array_distinct(distinct_regions)"},
                        {{"distinct_region"}})
                    .project(
                        {"nation AS nation2", "distinct_region AS region2"}),
                "nation1 = nation2",
                lp::JoinType::kInner)
            .build();
    auto plan = toSingleNodePlan(logicalPlanUnnest);
    std::cout << "Plan:\n" << plan->toString(true, true) << std::endl;
```

It's controled by this code

```cpp
      if (!contains(allowedInDt, PlanType::kUnnestNode)) {
        return wrapInDt(node);
      }
```

Now plan looks like this, which is behave similar to how AggregationNode behave

```
Plan:
-- Project[11][expressions: (nation1:BIGINT, "nation"), (region1:BIGINT, "distinct_region"), (nation2:BIGINT, "nation_0"), (region2:BIGINT, "distinct_region_3")] -> nation1:BIGINT, region1:BIGINT, nation2:BIGINT, region2:BIGINT
  -- HashJoin[10][INNER nation=nation_0] -> nation:BIGINT, distinct_region:BIGINT, nation_0:BIGINT, distinct_region_3:BIGINT
    -- Unnest[4][__r2] -> nation:BIGINT, distinct_region:BIGINT
      -- Project[3][expressions: (nation:BIGINT, "nation"), (__r2:ARRAY<BIGINT>, array_distinct("distinct_regions"))] -> nation:BIGINT, __r2:ARRAY<BIGINT>
        -- Unnest[2][__r2] -> nation:BIGINT, distinct_regions:ARRAY<BIGINT>
          -- Project[1][expressions: (nation:BIGINT, "nation"), (__r2:ARRAY<ARRAY<BIGINT>>, array_distinct("regions"))] -> nation:BIGINT, __r2:ARRAY<ARRAY<BIGINT>>
            -- Values[0][4 rows in 1 vectors] -> nation:BIGINT, regions:ARRAY<ARRAY<BIGINT>>
    -- Unnest[9][__r2] -> nation_0:BIGINT, distinct_region_3:BIGINT
      -- Project[8][expressions: (nation_0:BIGINT, "nation_0"), (__r2:ARRAY<BIGINT>, array_distinct("distinct_regions_2"))] -> nation_0:BIGINT, __r2:ARRAY<BIGINT>
        -- Unnest[7][__r2] -> nation_0:BIGINT, distinct_regions_2:ARRAY<BIGINT>
          -- Project[6][expressions: (nation_0:BIGINT, "nation_0"), (__r2:ARRAY<ARRAY<BIGINT>>, array_distinct("regions_1"))] -> nation_0:BIGINT, __r2:ARRAY<ARRAY<BIGINT>>
            -- Values[5][4 rows in 1 vectors] -> nation_0:BIGINT, regions_1:ARRAY<ARRAY<BIGINT>>
```


But opposite to AggregationNode Unnest doesn't decrease cardinality.
So maybe we should remove this branch in such case this plan will be used

```
```

Let's make it same way as project allow to push

MBkkt:
@mbasmanova It's still not ideal (unnest push down not happens) but it's UnnestTable + JoinEdge now

Unfortunately I understand this idea only after saw it in your WIP PR.

I think it works now and modeled as you asked.
Can it be merged now? Or some other things needs to be fixed/added first?

axiom/optimizer/Subfields.cpp
```diff
@@ -41,120 +41,191 @@ PathCP stepsToPath(const std::vector<Step>& steps) {
...
-      auto mark = [&](const lp::ExprPtr& expr) {
-        markSubfields(expr, subSteps, isControl, inputContext, inputSources);
-      };
-
-      const auto& keys = agg->groupingKeys();
-      if (ordinal < keys.size()) {
-        mark(keys[ordinal]);
-        return;
-      }
+void ToGraph::markFieldAccessed(
```
mbasmanova, line 106: @MBkkt Would you like to extract this refactoring into a separate PR? Smaller PRs are easier to review.

MBkkt:
> I'm seeing this issue now. Still looking into why this is happening and how to fix.

I think in case of Unnest is join I understand why.

We have such join edges

vt1 <= ut1
vt2 <= ut2
ut1 <=> ut2


Optimizer can start join from vt1 or vt2

vt1 <= ut1 (now it's impossible to put anything except vt2, vt2 will cause cross join)
vt2 <= ut2

Same here

axiom/optimizer/Subfields.cpp
MBkkt: https://github.com/facebookexperimental/verax/pull/273 Done

axiom/optimizer/Optimization.cpp
```diff
@@ -1047,6 +1050,42 @@ void Optimization::crossJoin(
   VELOX_NYI("No cross joins");
 }
 
+void Optimization::unnestJoin(
+    RelationOpPtr plan,
+    const JoinCandidate& candidate,
+    PlanState& state,
+    std::vector<NextJoin>& toTry) {
+  for (const auto* table : candidate.tables) {
```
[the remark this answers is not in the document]
MBkkt: > Do we ever have multiple entries here? I believe candidate.tables is either a single table or has additional tables that reduce the first table cardinality. Would that happen when we have multiple unnests joined together? 

It's needed otherwise my tests doesn't work.

> Would be nice to add a test case.

There is test case

`values().unnest(...).unnest(use result of previos unnest)`

axiom/optimizer/Optimization.cpp
```diff
@@ -1047,6 +1050,42 @@ void Optimization::crossJoin(
...
+    state.downstreamColumns().forEach<Column>([&](auto column) {
+      if (state.columns.contains(column)) {
+        replicateColumns.push_back(column);
+      }
+    });
+
+    // We don't use downstreamColumns() for unnestExprs/unnestedColumns
+    // because column unnest even when columns isn't used
+    // can change cardinality of the output for other columns
+    const auto& unnestExprs = candidate.join->filter();
```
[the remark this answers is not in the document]
MBkkt: But it's not used to conjuct? 
Filter is probably best what we can to do?
left keys doesn't work, because there's a lot of places that rely on left keys size equal to right keys size?

axiom/optimizer/Optimization.cpp
```diff
@@ -1346,6 +1391,10 @@ float startingScore(PlanObjectCP table) {
     return table->as<ValuesTable>()->cardinality();
   }
 
+  if (table->is(PlanType::kUnnestTableNode)) {
+    VELOX_FAIL("UnnestTable cannot be a starting table");
```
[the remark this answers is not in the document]
MBkkt: I will add the comment

axiom/optimizer/ToGraph.cpp
```diff
@@ -1610,7 +1663,28 @@ PlanObjectP ToGraph::makeQueryGraph(
       currentDt_->tableSet.add(setDt);
       return currentDt_;
     }
-    case lp::NodeKind::kUnnest:
+
+    case lp::NodeKind::kUnnest: {
+      // TODO: Unnest cannot be in single DT with Join right now.
```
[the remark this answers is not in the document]
MBkkt: https://github.com/facebookexperimental/verax/issues/286

axiom/optimizer/Optimization.cpp
```diff
@@ -1047,6 +1050,42 @@ void Optimization::crossJoin(
   VELOX_NYI("No cross joins");
 }
 
+void Optimization::unnestJoin(
```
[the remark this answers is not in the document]
mbasmanova: It is conventional to call it a cross join: https://prestodb.io/docs/current/sql/select.html#unnest

axiom/optimizer/Optimization.cpp
```diff
@@ -1047,6 +1050,42 @@ void Optimization::crossJoin(
...
+    std::vector<NextJoin>& toTry) {
+  for (const auto* table : candidate.tables) {
+    VELOX_CHECK(table->is(PlanType::kUnnestTableNode));
+    // We add unnest table before compute downstream columns because
+    // we're not interested in the replicating columns needed only for unnest.
+    state.placed.add(table);
+
+    ColumnVector replicateColumns;
+    state.downstreamColumns().forEach<Column>([&](auto column) {
+      if (state.columns.contains(column)) {
```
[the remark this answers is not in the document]
MBkkt: downstreamColumns contains all columns referenced by not processed yet nodes.

So it also contains columns that produced by next node
Project(x, a_a_i)
Unnest(a_a_i => a_i)  downstream here (after I added current node to state) contains `{x, a_i, i}`, but state.columns contains only `{x, a_a_i}`, so replicated columns will be their intersection
Unnest(a_i => i)  
Project(x, i)

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1165,6 +1165,273 @@ TEST_F(PlanTest, values) {
     checkSame(logicalPlan, referencePlan);
   }
 }
+
+TEST_F(PlanTest, unnest) {
```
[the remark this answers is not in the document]
MBkkt: Added

axiom/optimizer/Optimization.cpp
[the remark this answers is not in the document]
MBkkt: I separated this to another PR

axiom/optimizer/ToGraph.cpp
```diff
@@ -849,6 +850,58 @@ ExprVector ToGraph::translateColumns(const std::vector<lp::ExprPtr>& source) {
...
+      auto* column = make<Column>(columnName, unnestTable, value, columnName);
+      unnestTable->columns.push_back(column);
+      renames_[columnName] = column;
+    }
+  }
+
+  auto* edge = make<JoinEdge>(
+      leftTable,
+      unnestTable,
+      JoinEdge::Spec{.filter = std::move(unnestExprs), .directed = true});
```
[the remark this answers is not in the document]
MBkkt: Ok, I will try to put it in a left or right keys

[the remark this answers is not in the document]
MBkkt: @mbasmanova I made it works https://github.com/facebookexperimental/verax/pull/216/commits/5c35b7e91064a862ab710fbf75c43f319a6c63f0

Although I'm still not sure it's better

axiom/optimizer/ToGraph.cpp
```diff
@@ -1629,7 +1681,27 @@ PlanObjectP ToGraph::makeQueryGraph(
       currentDt_->tableSet.add(setDt);
       return currentDt_;
     }
-    case lp::NodeKind::kUnnest:
+
+    case lp::NodeKind::kUnnest: {
+      // Unnest doesn't mixed with Join.
```
mbasmanova, line 1686: perhaps, add a link the issue you created.

axiom/optimizer/QueryGraph.h
```diff
@@ -570,6 +571,11 @@ class JoinEdge {
     return rightTable_;
   }
 
+  void setKeys(ExprVector keys) {
+    leftKeys_ = keys;
+    rightKeys_ = std::move(keys);
```
mbasmanova, line 576: rightKeys_ are expected to come from the 'right' side of the join, no?

I think we need to set leftKeys_ and leave rightKeys_ empty. I do understand that we'll need to make more changes to accommodate for leftKeys_ not matching rightKeys_. I did that in my WIP PR and it wasn't too bad.

axiom/optimizer/Subfields.cpp
```diff
@@ -144,6 +176,9 @@ void ToGraph::markFieldAccessed(
   auto* fields = isControl ? &controlSubfields_ : &payloadSubfields_;
 
   const auto* path = stepsToPath(steps);
+  const auto& nodePaths = fields->nodeFields[source.planNode];
+  const bool nodePathsWasEmpty =
+      fields->nodeFields[source.planNode].resultPaths.empty();
```
mbasmanova, line 181: nit: use nodePaths

axiom/optimizer/Subfields.cpp
```diff
@@ -73,6 +73,38 @@ void ToGraph::markFieldAccessed(
...
+    std::span<const LogicalContextSource> sources,
+    bool nodePathsWasEmpty) {
+  const auto& input = unnest.onlyInput();
+  if (ordinal < input->outputType()->size()) {
+    markFieldAccessed(
+        {.planNode = input.get()}, ordinal, steps, isControl, context, sources);
+  }
+  // We need this to mark all unnest expressions as used, because even
+  // if they're not referenced they still can affect output cardinality.
+  // Also we don't want to call this multiple times for the same unnest.
```
mbasmanova, line 91: > // Also we don't want to call this multiple times for the same unnest.

Would you clarify what this refers to?

>    // We need to start unnestSteps from empty, because Unnest operator isn't
>    // projection, so steps after it cannot be pushdown-ed.

I'm not sure that's accurate.

```
SELECT x.foo
FROM t, unnest(a) as u(x)
```

Here, we do want to pushdown 'foo' into 't' and avoid reading / projecting other fields from the struct.

axiom/optimizer/Subfields.cpp
```diff
@@ -73,6 +73,38 @@ void ToGraph::markFieldAccessed(
...
+  if (nodePathsWasEmpty) {
+    // We need to start unnestSteps from empty, because Unnest operator isn't
+    // projection, so steps after it cannot be pushdown-ed.
+    std::vector<Step> unnestSteps;
```
mbasmanova, line 95: > unnestSteps

I'm not sure this name makes sense. This is an empty vector that is passed to markSubfields just to provide a temp storage. This doesn't store any unnest-specific value on the input and doesn't store any value on the output. Am I missing something?

(I'm thinking that it might be cleaner to provide a version of these methods that do not take 'steps' parameter and create empty vector internally as needed.)

axiom/optimizer/RelationOp.h
```diff
@@ -401,6 +402,20 @@ struct HashBuild : public RelationOp {
...
+struct Unnest : public RelationOp {
+  Unnest(
+      RelationOpPtr input,
+      ColumnVector replicateColumns,
+      ExprVector unnestExprs,
+      ColumnVector unnestedColumns);
+
+  ColumnVector replicateColumns;
+  ExprVector unnestExprs;
+  ColumnVector unnestedColumns;
```
mbasmanova, line 414: Let's document that unnestedColumns correspond to unnestExprs, but not 1:1 because each map has 2 entries.

axiom/optimizer/RelationOp.cpp
```diff
@@ -476,6 +476,26 @@ std::string Repartition::toString(bool recursive, bool detail) const {
   return out.str();
 }
 
+namespace {
+ColumnVector unionColumns(const ColumnVector& lhs, const ColumnVector& rhs) {
```
mbasmanova, line 480: nit: this is more like a concat then a union

axiom/optimizer/Optimization.cpp
```diff
@@ -1047,6 +1050,43 @@ void Optimization::crossJoin(
...
+        replicateColumns.push_back(column);
+      }
+    });
+
+    // We don't use downstreamColumns() for unnestExprs/unnestedColumns
+    // because column unnest even when columns isn't used
```
mbasmanova, line 1072: typos

axiom/optimizer/Optimization.cpp
```diff
@@ -1047,6 +1050,43 @@ void Optimization::crossJoin(
...
+    // can change cardinality of the output for other columns.
+    const auto& unnestExprs = candidate.join->leftKeys();
+    const auto& unnestedColumns = table->as<UnnestTable>()->columns;
+
+    // Plan is updated here, because it can multiple unnest.
```
mbasmanova, line 1077: typos

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1211,6 +1211,672 @@ TEST_F(PlanTest, parallelCse) {
...
+              "[[10, 30], [100, 300]]",
+              "[[2, 1], [1, 2]]",
+              "[[100, 300], [10, 30]]",
+              "[[0, 0], [0, 0]]",
+              "[[2, 1], [1, 2]]",
+          }),
+      });
+  // We need to test following cases:
+  //  If something after unnest
+  //  it can depend and and not depend on unnested columns.
```
mbasmanova, line 1244: typos

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1211,6 +1211,672 @@ TEST_F(PlanTest, parallelCse) {
...
+  // 2. unnest after unnest
+  // 3. project before and after unnest
+  // -- after this we will start to use project to simplify plans --
+  // 4. filter before and after unnest
+  // 5. group by before and after unnest
+  // 6. order by before and after unnest
+  // 7. limit before and after unnest
+  // 8. join before and after unnest
+
+  std::cerr << "1. unnest\n";
```
mbasmanova, line 1257: replace with SCOPED_TRACE

mbasmanova (commented):
@MBkkt Overall looks good to me. Thank you for iterating patiently.

(I do worry about needing to re-design this when adding support for join orders, but don't have bandwidth to sort this out now.)

I have a few nits and 1 comment re: dummy values node.

axiom/optimizer/Subfields.cpp
MBkkt: I don't understand how we can pushdown `x.foo` because if we have
`a = [{"foo": 1}, {"foo": 2}, {"foo": 3}]`, just pushdown `x.foo` will be incorrect we need to do `unnest` on `a`

axiom/optimizer/ToGraph.cpp
```diff
@@ -849,6 +850,57 @@ ExprVector ToGraph::translateColumns(const std::vector<lp::ExprPtr>& source) {
...
+    maxCardinality = std::max(maxCardinality, unnestExpr->value().cardinality);
+  }
+
+  auto* unnestTable = make<UnnestTable>();
+  unnestTable->cname = newCName("ut");
+  unnestTable->columns.reserve(
+      unnest.outputType()->size() - unnest.onlyInput()->outputType()->size());
+  for (size_t i = 0; const auto& unnestedNames : unnest.unnestedNames()) {
+    const auto* unnestExpr = unnestExprs[i++];
+    for (size_t j = 0; const auto& unnestedName : unnestedNames) {
```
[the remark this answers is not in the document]
mbasmanova: I assume unnestExpr is only used to get the type and that is available in unnest.outputType(), hence, I think unnestExpr is not needed.

axiom/optimizer/Subfields.cpp
mbasmanova: Let's say "a" is array(row(foo, bar)). 

> [{foo: 1, bar: 10}, {foo: 2, bar: 20}]

Unnest does need all elements of the array, but it doesn't need the whole value of each element. It is sufficient to unnest [{foo: 1}, {foo: 2}], no need to unnest the whole [{foo: 1, bar: 10}, {foo: 2, bar: 20}].

MBkkt:
> This creates an extra joins and leads to unnecessarily more complex join order path.

@mbasmanova

I don't think exists execution plan difference between values dummy node/unnest allowed as leaf node.

It creates additional join, but only for simple queries without table scans at all. So I think it should be quite ok?

MBkkt:
> Unnest isn't leaf node. I think I already wrote my opinion on this. Maybe if you can provide example or explain in what cases it's bad it will be helpful to me to understand why do you think it's needed.

Also example where we need constant unnest to be common join edge in the optimizer:

```sql
SELECT * FROM 
   (SELECT x AS x1, UNNEST(generate_series(1, 10000)) as s1 FROM t) AS t1
  JOIN 
   (SELECT  x AS x2, UNNEST(generate_series(1, 10000)) as s2 FROM t) AS t2
  ON x1 = x2 
```

MBkkt:
@mbasmanova In last commit I fixed logic around Subfields for Unnest.

I missed `markControl` and implemented it logic (with my hack to call it only once) inside `markFieldAccessed`.

Now it's fixed.

`markControl` used for all `unnestExpressions` and `markFieldAccessed` used only for replicated columns

axiom/optimizer/ToGraph.cpp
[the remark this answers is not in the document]
mbasmanova: Wondering, is something like this not working?

```
const auto numInputs = unnest.input()->outputType()->size();
const auto numOutputs = unnest.outputType()->size();
for (auto i = numInputs; i < numOutputs; ++i) {
      const auto* unnestedType = unnest.outputType()->childAt(i).get();

      Value value{unnestedType, maxCardinality};
      const auto* columnName = toName(unnest.outputType()->nameOf(i));
      auto* column = make<Column>(columnName, unnestTable, value, columnName);
      unnestTable->columns.push_back(column);
      renames_[columnName] = column;
}  
```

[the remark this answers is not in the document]
MBkkt: Not exactly, because `unnest.outputType()->childAt(i).get();` cannot be used in optimizer.
It needs to be lookuped in `queryCtx()` in such case, something like this will work, but will make extra lookups:
`const auto* unnestedType = toType(unnest.outputType()->childAt(i));`

I also think having access to unnestExpr here can be useful in future.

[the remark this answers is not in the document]
mbasmanova: Got it.

If we keep these 2 loops, would it be possible to make increments for i and j a bit easier to spot. Currently, these are "hidden" inside some dense code and difficult to find.

axiom/optimizer/QueryGraph.h
```diff
@@ -570,6 +571,13 @@ class JoinEdge {
     return rightTable_;
   }
 
+  // This method should be removed, see
+  // https://github.com/facebookexperimental/verax/issues/298
+  void setKeys(ExprVector keys) {
```
mbasmanova, line 576: Would it make sense to introduce JoinEdge::makeUnnest static method to encapsulate the details of how join edges for Unnest are configured?

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1211,6 +1211,688 @@ TEST_F(PlanTest, parallelCse) {
...
+  // -- after this we will start to use project to simplify plans --
+  // 4. filter before and after unnest
+  // 5. group by before and after unnest
+  // 6. order by before and after unnest
+  // 7. limit before and after unnest
+  // 8. join before and after unnest
+  // 9. there's no extra columns in projections before unnest
+
+  {
+    SCOPED_TRACE("1. unnest");
```
mbasmanova, line 1259: Using ordinal numbers like this is fragile as these easily become wrong during refactoring or when tests are added in the middle. Can we drop these numbers?

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1211,6 +1211,688 @@ TEST_F(PlanTest, parallelCse) {
   EXPECT_TRUE(matcher->match(plan));
 }
 
+TEST_F(PlanTest, unnest) {
+  const std::vector<std::string> names{"x", "a_a_y", "a_a_z"};
```
mbasmanova, line 1215: > a_a_y

Does this stand for array-of-arrays?

MBkkt: Yes

axiom/optimizer/tests/PlanTest.cpp
MBkkt: >  Can we drop these numbers?

Ok

axiom/optimizer/QueryGraph.h
MBkkt: Done

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1211,6 +1211,688 @@ TEST_F(PlanTest, parallelCse) {
   EXPECT_TRUE(matcher->match(plan));
 }
 
+TEST_F(PlanTest, unnest) {
```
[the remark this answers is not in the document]
MBkkt: Done

axiom/optimizer/tests/UnnestTest.cpp
```diff
@@ -0,0 +1,710 @@
...
+                                       "array_distinct(a_a_z) AS a_a_z_d",
+                                   })
+                                   .unnest({"x"}, {"a_a_y_d", "a_a_z_d"})
+                                   .project({"x", "a_a_y_d_e"})
+                                   .planNode();
+
+    checkSame(logicalPlanUnnest, referencePlanUnnest);
+  }
+  {
+    SCOPED_TRACE("filter after unnest (independent on unnested columns)");
```
mbasmanova, line 252: Specifically, here, it seems important to verify that filter got pushed down before the unnest. Otherwise, the results may be correct, but the plan is not optimal.

axiom/optimizer/tests/UnnestTest.cpp
```diff
@@ -0,0 +1,710 @@
...
+#include "axiom/optimizer/tests/PlanMatcher.h"
+#include "axiom/optimizer/tests/PlanTest.h"
+#include "velox/exec/tests/utils/PlanBuilder.h"
+
+namespace lp = facebook::velox::logical_plan;
+
+namespace facebook::velox::optimizer {
+namespace {
+
+TEST_F(PlanTest, unnest) {
```
mbasmanova, line 27: PlanTest -> UnnestTest

MBkkt: `(UnnestTest, unnest)`?

axiom/optimizer/tests/UnnestTest.cpp
```diff
@@ -0,0 +1,710 @@
...
+  {
+    SCOPED_TRACE("there's no extra columns in projections before unnest");
+
+    const std::vector<std::string> expectedNames{"x", "y"};
+
+    auto logicalPlanUnnest =
+        lp::PlanBuilder{}
+            .values({rowVector})
+            .unnest({
+                lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
```
mbasmanova, line 683: Can we add tests for unnest with constant expressions? unnest(array[1,2,3]) or unnest(sequence(1, 10))?

Can we also add tests for unnesting maps. Unnesting arrays is simpler because there is 1:1 mapping between unnest expr and output column, but for maps the mapping is 2:1. Would be nice to make sure this all works fine.

axiom/optimizer/tests/UnnestTest.cpp
mbasmanova: I would break this up further. It looks like there are different groups of tests: filters, agg, sort, join. I would put each group into its own test method.

```
class UnnestTest..

TEST_F(UnnestTest, basic) 

TEST_F(UnnestTest, filter) 

TEST_F(UnnestTest, agg) 

TEST_F(UnnestTest, sort) 

TEST_F(UnnestTest, join) 
```

Hope it makes sense.

axiom/logical_plan/LogicalPlanNode.h
```diff
@@ -610,8 +610,7 @@ using SetNodePtr = std::shared_ptr<const SetNode>;
 /// optional ordinality column of type BIGINT.
 class UnnestNode : public LogicalPlanNode {
  public:
-  /// @param input Optional input node. If not specified, 'unnestExpressions'
-  /// must be constant expressions.
+  /// @param input Input node.
```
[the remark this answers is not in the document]
MBkkt: I fixed this

axiom/optimizer/Subfields.cpp
MBkkt: This code removed, so I resolve it

axiom/optimizer/Subfields.cpp
MBkkt: code removed

axiom/optimizer/tests/UnnestTest.cpp
MBkkt: Done for both

axiom/optimizer/tests/UnnestTest.cpp
```diff
@@ -0,0 +1,710 @@
...
+                "x",
+                "a_a_y",
+                "a_a_z",
+                "array_distinct(a_a_y) AS a_a_y_d",
+                "array_distinct(a_a_z) AS a_a_z_d",
+            })
+            .unnest({"x", "a_a_y", "a_a_z"}, {"a_a_y_d", "a_a_z_d"})
+            .planNode();
+
+    checkSame(logicalPlanUnnest, referencePlanUnnest);
```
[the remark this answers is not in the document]
MBkkt: It's fixed, now all unnest tests verify plan shape

MBkkt:
I see, something broken after multiple rebase, I will check

MBkkt:
>  coverage for 'constant' Unnest is added. 

Yes it is.

MBkkt:
> Perhaps, advance Velox to see if that reproduces the error?

I tried, and there's no difference.
My branch is also last github commit.

mbasmanova:
@MBkkt The PR description seems outdated. Would you double check?

> We have one issue, join => unnest (in single dt) is possible and can be reordered, but unnest => join will be in different dt (so right now cannot be reordered). But it relates not to this PR, but to the current main state about joins (same for an example that right part of non-inner join tree will be in a separate dt, so cannot be reordered)

Unless you recent comments on the issue, it wasn't clear to me that this is a generic problem and not the one that is specific to unnest. There are no blockers for this PR other than I just need to read it one more time and make sure tests are passing (PrestoParser test used to fail earlier).

Thank you for your patience.

axiom/optimizer/QueryGraph.h
```diff
@@ -506,6 +507,19 @@ class JoinEdge {
...
 
+  static JoinEdge* makeUnnest(
+      PlanObjectCP leftTable,
+      PlanObjectCP rightTable,
+      ExprVector unnestExprs) {
+    VELOX_DCHECK_NOT_NULL(leftTable);
+    auto* edge = make<JoinEdge>(leftTable, rightTable, Spec{.directed = true});
+    edge->leftKeys_ = std::move(unnestExprs);
+    // TODO Not sure to what values fanout need to be set,
+    // (1, 1) looks ok, but tests don't produce expected plans.
```
mbasmanova, line 518: > but tests don't produce expected plans

What does this refer to? Should we create issues to follow-up?

axiom/optimizer/tests/UnnestTest.cpp
```diff
@@ -0,0 +1,1121 @@
...
+#include "axiom/optimizer/tests/PlanTest.h"
+#include "velox/exec/tests/utils/PlanBuilder.h"
+
+namespace facebook::axiom::optimizer {
+namespace {
+
+using namespace velox;
+namespace lp = facebook::axiom::logical_plan;
+
+class UnnestTest : public PlanTest {
```
mbasmanova, line 28: PlanTest should not be extended. We should extend test::QueryTestBase or similar. I'll change that.

axiom/optimizer/tests/UnnestTest.cpp
```diff
@@ -0,0 +1,1121 @@
...
+                                   })
+                                   .unnest({"x"}, {"a_a_y_d", "a_a_z_d"})
+                                   .limit(1, 1, {})
+                                   .planNode();
+
+    checkSame(logicalPlanUnnest, referencePlanUnnest);
+  }
+}
+
+TEST_F(UnnestTest, join) {
```
mbasmanova, line 951: Any particular reason these tests do not verify that produced plans can run and generate correct results?

axiom/optimizer/QueryGraph.h
MBkkt: Well idea was simple, if you write here (1, 1), one test will fail, it expect that unnest will be after inner joins, but for some reason, if here 1, 1, it's mixed in a strange way

MBkkt:
@mbasmanova I added execution tests for join, and made two readability improvements that you suggested for tests.

axiom/optimizer/tests/UnnestTest.cpp
MBkkt: > I'll change that.

Ok, ask me if I need to do something with it

axiom/optimizer/tests/UnnestTest.cpp
```diff
@@ -0,0 +1,1121 @@
...
+            })
+            .unnest({
+                lp::Sql("array_distinct(a_y)").unnestAs("y"),
+                lp::Sql("array_distinct(a_z)").unnestAs("z"),
+            })
+            .project(expectedNames)
+            .build();
+    auto plan = toSingleNodePlan(logicalPlanUnnest, 1);
+
+    // names like this because they're autogenerated by the optimizer
```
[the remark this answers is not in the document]
mbasmanova: The matcher should allow us to provide aliases for all these columns. The matcher then needs to map these properly to actual columns names in the plan. The code in Presto I referenced allows exactly that.

PR #224, opened 2025-08-19
refactor: Add PlanObject::is(PlanType) helper method

mbasmanova:
Summary:
Before:

> object->type() == PlanType::kDerivedTableNode

After:

> object->is(PlanType::kDerivedTableNode)

Differential Revision: D80570004


xiaoxmeng (approved):
@mbasmanova thanks!

PR #225, opened 2025-08-19
refactor: Rename PlanPtr to PlanP

mbasmanova:
Summary: Ptr suffix is for smart pointer. P is for raw pointer.

Differential Revision: D80578407


xiaoxmeng (approved):
@mbasmanova thanks!

PR #227, opened 2025-08-20
feat: Move findTable to smart pointer return type

hdikeman:
Summary:
was hard to manage the lifecycle of tables when the Optimizer (or other findTable callers) could dereference the pointer at any time. e.g. if a second concurrent query determines the table information is stale, or planning is stuck and the table is pushed out of cache, the server can segfault

Changing the return type to shared_ptr<const Table> (ConnectorTablePtr) and updating all the callers

Also, adding a set in the Optimizer to capture a reference to all the Tables used in Optimization, to prevent reclaimation prior to Optimization completion

Differential Revision: D80525688


hdikeman:
had to fix the title and some license headers for files I touched

peterenescu (approved):

PR #230, opened 2025-08-20
build(CI): Add Antlr4 support in CI

feilong-liu:
Make the following changes:
* Update the file generated in sql/presto from Antlr 4.9.3 to 4.13.2, because 4.9.3 is too old and does not build in verax environment (due to CMake version too old and cmake policy violation)
* Enable build and test of code in sql/presto, to make sure that Antlr4 indeed works
* Update CI file to include Antlr4


CMakeLists.txt
```diff
@@ -140,3 +146,4 @@ add_subdirectory(velox)
 add_subdirectory(axiom/logical_plan)
 add_subdirectory(axiom/runner)
 add_subdirectory(axiom/optimizer)
+add_subdirectory(axiom/sql/presto)
```
feilong-liu, line 149: Enable build and test of axiom/sql/presto

PR #238, opened 2025-08-21
feat: Add table write logical plan node and partition type

oerling:
Adds a logical plan node for table write and generate write plans and produce a write finishing function.

Adds a PartitionType to ConnectorMetadata. This supports flexible determiniation of possible copartitioning.

- Uses  connector::PartitionType to describe a partition function.

- Adds a ConnectorMetadata method for getting the result type of a table write.

- Adds failure recovery to LocalHiveConnectorMetadata. Files are first staged in a temp directory and moved to the destination only at in the success case of the write finishing function.

- Adds a separate write execution path to LocalRunner. This silenntly gathers results from the write query and passes these to the write finishing function. Calls the write finishing function with success or with failure if the query has an error.

### DIFF
--- a/axiom/logical_plan/LogicalPlanNode.cpp
+++ b/axiom/logical_plan/LogicalPlanNode.cpp
@@ -24,6 +24,7 @@ const auto& nodeKindNames() {
   static const folly::F14FastMap<NodeKind, std::string_view> kNames = {
       {NodeKind::kValues, "VALUES"},
       {NodeKind::kTableScan, "TABLE_SCAN"},
+      {NodeKind::kTableWrite, "TABLE_WRITE"},
       {NodeKind::kFilter, "FILTER"},
       {NodeKind::kProject, "PROJECT"},
       {NodeKind::kAggregate, "AGGREGATE"},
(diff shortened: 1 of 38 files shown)


axiom/logical_plan/LogicalPlanNode.cpp
```diff
@@ -351,4 +351,31 @@ void UnnestNode::accept(
...
 
+void TableWriteNode::accept(
+    const PlanNodeVisitor& visitor,
+    PlanNodeVisitorContext& context) const {
+  visitor.visit(*this, context);
+}
+
+RowTypePtr TableWriteNode::makeWriteType() {
+  return ROW(
+      {"numWrittenRows", "fragment", "tableCommitContext"},
```
mbasmanova, line 362: Why do we hard-code this type? It might be cleaner to let connector define the output type of the writer.

axiom/logical_plan/PlanBuilder.cpp
```diff
@@ -1073,6 +1073,20 @@ PlanBuilder& PlanBuilder::offset(int64_t offset) {
...
+PlanBuilder& PlanBuilder::tableWrite(
+    const std::string& connectorId,
+    const std::string& tableName,
+    WriteKind kind,
+    const std::vector<std::string>& columnNames,
+    const std::unordered_map<std::string, std::string> options) {
+  VELOX_USER_CHECK_NOT_NULL(node_, "Table write node cannot be a leaf node");
+
+  node_ = std::make_shared<TableWriteNode>(
+      nextId(), node_, connectorId, tableName, kind, columnNames, options);
```
mbasmanova, line 1085: What kind of names are in "columnNames"? Are these names in table schema? Do we expect the input to this node to match columnNames? Let's add checks.

axiom/logical_plan/PlanPrinter.cpp
```diff
@@ -153,6 +153,11 @@ class ToTextVisitor : public PlanNodeVisitor {
     appendNode("Unnest", node, context);
   }
 
+  void visit(const TableWriteNode& node, PlanNodeVisitorContext& context)
+      const override {
+    appendNode("TableWrite", node, context);
```
mbasmanova, line 158: Should we include details about the table write in the output?

PR #239, opened 2025-08-21
feat: Integrate Presto SQL Parser

mbasmanova:
Summary: Use PrestoParser in HiveQueriesTestBase.cpp by default. Switch TpchPlanTest to parser TPC-H SQL using PrestoParser.

Differential Revision: D80723272


xiaoxmeng (approved):
@mbasmanova thanks!

PR #256, opened 2025-08-23
refactor: Avoid unnessary static strings and update velox

MBkkt:
Also https://github.com/facebookexperimental/verax/issues/257 fix this


mbasmanova:
Overlaps with #253

MBkkt:
@mbasmanova where? I already updated PR to the current main

MBkkt:
@mbasmanova Can you look?

PR #265, opened 2025-08-25
fix: Fix SQL parsing of CROSS JOIN UNNEST

mbasmanova:
Summary: Fix SQL parsing of `SELECT * FROM t, unnest(a)` into Scan -> Unnest vs. Join(Scan, Unnest).

Differential Revision: D80940365


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #275, opened 2025-08-26
refactor: Replace JoinCandidate::sideOf with joinSides

mbasmanova:
Differential Revision: D81038379


pansatadru (approved):
Thanks, nice simplification.

PR #282, opened 2025-08-26
feat: Make --data_path optional in CLI

mbasmanova:
Summary: Use Hive connector when --data_path is specified and TPC-H connector otherwise.

Differential Revision: D81076503


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #283, opened 2025-08-26
Copy Enums.h file to axiom/logical_plan

pedroerp:
Summary:
Copying file over from Velox since this isn't part of the external API
and not meant to be used by separate projects.

Differential Revision: D79216422

### DIFF
--- a/axiom/logical_plan/Enums.h
+++ b/axiom/logical_plan/Enums.h
@@ -1,0 +1,149 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


MBkkt (commented):
Can you explain please what a point to have such duplication?
Maybe it make sense to do this things part of some public headers in velox?

pedroerp:
> Can you explain please what a point to have such duplication? Maybe it make sense to do this things part of some public headers in velox?

@MBkkt strictly speaking, these headers are not part of Velox's external API, so they shouldn't be used outside of Velox. Having an external API that is too wide makes it harder for us to evolve Velox, as too many of the changes end up breaking client code.

For example, some of your recent changes to Velox APIs (baseVector variant, type enums, etc) ended up being a headache to merge because they end up breaking a large number of client code. In general, we should be more intentional about the APIs we expose as public in Velox (and in Verax).

PR #290, opened 2025-08-27
fix: Redundant Projections detection isn't correct

MBkkt:
Tests will be added in Unnest PR


axiom/optimizer/ToVelox.cpp
```diff
@@ -1081,37 +1081,50 @@ velox::core::PlanNodePtr ToVelox::makeProject(
     }
   }
 
-  bool redundant = true;
-  for (auto i = 0; i < project.exprs().size(); ++i) {
-    auto expr = project.exprs()[i];
-    if (expr->type() != PlanType::kColumnExpr) {
-      redundant = false;
-      break;
+  const auto& inputType = *input->outputType();
```
mbasmanova, line 1084: I'm not able to see what is changing here. Would you clarify? A test case would be helpful.

axiom/optimizer/ToVelox.cpp
```diff
@@ -1081,37 +1081,50 @@ velox::core::PlanNodePtr ToVelox::makeProject(
...
+  VELOX_DCHECK_EQ(project.columns().size(), outputTypeSize);
+
+  // redundant == false is possible because multiple reasons:
+  // 1. Unnest should refererence all their columns
+  //    if at least one column referenced.
+  // 2. Join should reference at least one column from each side,
+  //    if at least one column referenced.
+  // 3. some project expressions are not input column references.
+  const bool redundant = [&] {
+    if (inputType.size() != outputTypeSize) {
```
MBkkt, line 1095: @mbasmanova this line is main change.

Before my pr Projection was redundant even in case when `inputType.size() > outputTypeSize`
And it's wrong for last projection node.

axiom/optimizer/ToVelox.cpp
```diff
@@ -1081,37 +1081,50 @@ velox::core::PlanNodePtr ToVelox::makeProject(
...
-  for (auto i = 0; i < project.exprs().size(); ++i) {
-    auto expr = project.exprs()[i];
-    if (expr->type() != PlanType::kColumnExpr) {
-      redundant = false;
-      break;
+  const auto& inputType = *input->outputType();
+  const auto outputTypeSize = project.exprs().size();
+  VELOX_DCHECK_EQ(project.columns().size(), outputTypeSize);
+
+  // redundant == false is possible because multiple reasons:
```
mbasmanova, line 1088: Why is this comment relevant here?

MBkkt: I probably should remove it

axiom/optimizer/ToVelox.cpp
MBkkt: I fixed this because added tests

This query leads to invalid behavior, because unnest *should* produce "a_z".
But I don't know how to write test for this without unnest, probably something else with control subfields also should work

```
    auto logicalPlanUnnest =
        lp::PlanBuilder{}
            .values({rowVector})
            .filter("x % 2 = 0")
            .unnest({
                lp::Sql("array_distinct(a_a_y)").unnestAs("a_y"),
                lp::Sql("array_distinct(a_a_z)").unnestAs("a_z"),
            })
            .project({"x", "a_y"})
            .build();
```

mbasmanova: @MBkkt I see. I think this can be reproduced with

```
TEST_F(PlanTest, xxx) {
  testConnector_->createTable(
      "numbers", ROW({"a", "b", "c"}, {BIGINT(), DOUBLE(), VARCHAR()}));

  auto logicalPlan = lp::PlanBuilder()
                         .tableScan(kTestConnectorId, "numbers", {"a", "b"})
                         .aggregate({"a", "b"}, {"count(1)"})
                         .map({"a"})
                         .build();

  auto plan = toSingleNodePlan(logicalPlan);

  LOG(ERROR) << plan->toString(true, true) << std::endl;
}

-- Aggregation[3][FINAL [a, b] ] -> a:BIGINT, b:DOUBLE
  -- LocalPartition[2][REPARTITION <Verax partition function spec>] -> a:BIGINT, b:DOUBLE
    -- Aggregation[1][PARTIAL [a, b] ] -> a:BIGINT, b:DOUBLE
      -- TableScan[0][numbers] -> a:BIGINT, b:DOUBLE
```

The output should be one column (a), but it is 2 columns (a, b).

axiom/optimizer/ToVelox.cpp
```diff
@@ -1128,37 +1128,46 @@ velox::core::PlanNodePtr ToVelox::makeProject(
...
   }
 
-  bool redundant = true;
-  for (auto i = 0; i < project.exprs().size(); ++i) {
-    auto expr = project.exprs()[i];
-    if (expr->type() != PlanType::kColumnExpr) {
-      redundant = false;
-      break;
+  const auto& inputType = *input->outputType();
+  const auto outputTypeSize = project.exprs().size();
```
mbasmanova, line 1132: nit: a conventional name is numOutputs

axiom/optimizer/ToVelox.cpp
```diff
@@ -1128,37 +1128,46 @@ velox::core::PlanNodePtr ToVelox::makeProject(
...
-    if (input->outputType()->nameOf(i) != outputName(column)) {
-      redundant = false;
-      break;
+    for (size_t i = 0; i < outputTypeSize; ++i) {
+      const auto* expr = project.exprs()[i];
+      if (expr->type() != PlanType::kColumnExpr) {
+        return false;
+      }
+      const auto* column = project.columns()[i];
+      // TODO Why expr->sameOrEquals(*column) doesn't work?
```
mbasmanova, line 1147: I know the answer now. Let's remove this.

sameOrEquals is confusing name. It only tests for explicit equivalence, e.g. join keys are equal.

mbasmanova (approved):
Thank you for the fix.

axiom/optimizer/ToVelox.cpp
```diff
@@ -1128,37 +1128,46 @@ velox::core::PlanNodePtr ToVelox::makeProject(
...
-      redundant = false;
-      break;
+  const auto& inputType = *input->outputType();
+  const auto outputTypeSize = project.exprs().size();
+  VELOX_DCHECK_EQ(project.columns().size(), outputTypeSize);
+
+  const bool redundant = [&] {
+    if (inputType.size() != outputTypeSize) {
+      // TODO Maybe we don't want always return false here, see
+      // https://github.com/facebookexperimental/verax/issues/311
```
mbasmanova, line 1138: I'm not following the description of this issue. Let's remove it from here along with a TODO and add a pointer to the code to the issue instead.

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1307,6 +1307,31 @@ TEST_F(PlanTest, parallelCse) {
...
+      "numbers", ROW({"a", "b", "c"}, {BIGINT(), DOUBLE(), VARCHAR()}));
+
+  auto logicalPlan = lp::PlanBuilder{}
+                         .tableScan(kTestConnectorId, "numbers", {"a", "b"})
+                         .aggregate({"a", "b"}, {"count(1)"})
+                         .project({"a"})
+                         .build();
+
+  auto plan = toSingleNodePlan(logicalPlan);
+  ASSERT_TRUE(*plan->outputType() == *ROW("a", BIGINT()))
```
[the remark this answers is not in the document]
MBkkt: For me plan matcher output is quite useless, when I write tests and it fails I always print current plan before matcher to understand what differs

axiom/optimizer/ToVelox.cpp
MBkkt: The code isn't merged yet, but ok

axiom/optimizer/tests/PlanTest.cpp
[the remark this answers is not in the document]
mbasmanova: > but output isn't good in such case.

Are you saying the error message from the Matcher needs improvement? Let's do that then. It would be nice to consolidate plan verification in a single place (matcher) and do that in a unified way across all tests if possible.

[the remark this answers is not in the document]
MBkkt: Ok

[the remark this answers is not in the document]
mbasmanova: > For me plan matcher output is quite useless, when I write tests and it fails I always print current plan before matcher to understand what differs

Sounds like we need to add SCOPED_TRACE(plan->toString(true, true)) at the start of 'match'.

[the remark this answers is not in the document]
MBkkt: I have better suggestion (at least I think it's better) https://github.com/facebookexperimental/verax/issues/312

PR #291, opened 2025-08-27
refactor: Reduce number of arguments in markFieldAccessed

mbasmanova:
Differential Revision: D81168742


MBkkt (commented):
Thanks for applying review suggestion!

mbasmanova:
@MBkkt Appreciate detailed review. All comments should be addressed now. Let me know if you have any further suggestions.

PR #292, opened 2025-08-27
feat: Add parsing for TPC-H q2,17

mbasmanova:
Summary: These queries require subquery support in the optimizer. At this point we can parse, but not optimize them.

Differential Revision: D81176406

### DIFF
--- a/axiom/optimizer/tests/PrestoParser.cpp
+++ b/axiom/optimizer/tests/PrestoParser.cpp
@@ -454,7 +454,10 @@ class RelationPlanner : public sql::AstVisitor {
         if (query->is(sql::NodeType::kQuery)) {
           auto builder = std::move(builder_);
 
-          builder_ = newBuilder();
+          lp::PlanBuilder::Scope scope;
+          builder->captureScope(scope);
+
+          builder_ = newBuilder(scope);
           processQuery(query->as<sql::Query>());
           auto subqueryBuider = builder_;
 
--- a/axiom/optimizer/tests/TpchPlanTest.cpp
+++ b/axiom/optimizer/tests/TpchPlanTest.cpp
@@ -88,8 +88,22 @@ class TpchPlanTest : public virtual test::HiveQueriesTestBase {
     return sql;
   }
 
+  std::string readTpchSql(int32_t query) {
+    return readSqlFromFile(fmt::format("tpch.queries/q{}.sql", query));
+  }
+
+  void parseTpchSql(int32_t query) {
+    auto sql = readTpchSql(query);
+
+    auto statement = prestoParser_->parse(sql);
+
+    ASSERT_TRUE(statement->isSelect());
+    ASSERT_TRUE(
+        statement->asUnchecked<test::SelectStatement>()->plan() != nullptr);
+  }
+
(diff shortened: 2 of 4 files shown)


axiom/optimizer/ToGraph.h
```diff
@@ -53,12 +53,20 @@ struct BuiltinNames {
...
   const logical_plan::CallExpr* call{nullptr};
   int32_t lambdaOrdinal{-1};
 };
 
+struct MarkFieldsAccessedContext {
+  // 1:1 with 'sources'. Either output type of the plan node or signature of a
+  // lambda.
+  std::vector<const RowType*> rowTypes;
+  std::vector<LogicalContextSource> sources;
+};
```
MBkkt, line 68: Why does this class needed? It's not changed readability a lot but introduce back dynamic allocations on every recursive call of markFieldAccessed/etc

axiom/optimizer/Subfields.cpp
```diff
@@ -474,19 +448,14 @@ void ToGraph::markControl(const lp::LogicalPlanNode& node) {
...
 
-  std::vector<const RowType*> context;
-  std::vector<LogicalContextSource> sources;
-  for (auto i = 0; i < type->size(); ++i) {
-    std::vector<Step> steps;
-    markFieldAccessed(
-        source, i, steps, /* isControl */ false, context, sources);
+  for (auto i = 0; i < node.outputType()->size(); ++i) {
+    std::vector<Step> empty;
+    markFieldAccessed(source, i, empty, /* isControl */ false, {});
```
MBkkt, line 458:
```suggestion
  std::vector<Step> steps;
  for (size_t i = 0; i < node.outputType()->size(); ++i) {
    markFieldAccessed(source, i, empty, /* isControl */ false, {});
    VELOX_DCHECK(steps.empty());
```
In such case steps allocation will be reused between different ordinals

axiom/optimizer/ToGraph.h
mbasmanova: I think this is easier to read. I still expect more refactoring to simply this pattern:

``` 
    steps.push_back(...;
    markSubfields(input, steps, isControl, context);
    steps.pop_back();
```

mbasmanova: That's a nice trick, but would about the join case?

```
      markSubfields(
          condition,
          empty,
          /* isControl */ true,
          fromNodes(join->inputs()));
```

MBkkt: https://godbolt.org/z/EPW4fcjWc

MBkkt: > Why do you believe this matters?

I dunno but allocation on almost every path step doesn't look good to me.
Especially when it's easy to write without it.

> How are you thinking about it? 

Well, I think main code is ok, not very readable but not that bad to refactor it this way.

> I can make that change, but this is a lot of extra code. 

Not really a lot, it's like 10-20 additional lines?

mbasmanova: @MBkkt Makes sense. Updated: https://github.com/facebookexperimental/verax/pull/291/

xiaoxmeng (approved):

PR #300, opened 2025-08-28
test

feilong-liu:


assignUser:
I found the issue, actions/checkout cleans out the workspace by default:
```yaml
# Whether to execute `git clean -ffdx && git reset --hard HEAD` before fetching
    # Default: true
    clean: ''
```

So as we can see the cache gets downloaded correctly but is then deleted by `actions/checkout`. The solution is to just set `CCACHE_DIR=/tmp/ccache`

assignUser:
Ah, that's probably because you already had a stash were that path was logged. You could probably make it work by changing the stash key to get a clean start but this way works too :)

kgpai:
Thanks @assignUser for root causing the bug and fix.

Issue #307, opened 2025-08-29
Incorrect tests ORDER BY and subfields


MBkkt:
https://github.com/facebookexperimental/verax/blob/f41140842edca612c9cb9ad5f209c4f3d78d0a40/axiom/optimizer/tests/HiveQueriesTest.cpp#L131-L133

project in this query is wrong

mbasmanova:
Any chance you could show the full query and the full plan?

mbasmanova:
Project[2] node changes the order of columns. Are you saying we should make TopN produce columns in the right order?

PR #308, opened 2025-08-29
fix: Don't produce unnessary projections

MBkkt:
Eliminate unnecessary projections that restore the order of columns.

- usedChannels now return sorted list (It's the root of issue)
- skylineStruct have similar issue, it change order of columns compared to implementation where control and payload is single map. So I also refactored it (as a bonus it should works better because doesn't have unordered set of strings)

Fixes https://github.com/facebookexperimental/verax/issues/307

### DIFF
--- a/axiom/optimizer/ToVelox.cpp
+++ b/axiom/optimizer/ToVelox.cpp
@@ -898,34 +898,30 @@ bool hasSubfieldPushdown(const TableScan& scan) {
 // 'baseTable'. This is the type to return from the table reader
 // for the map column.
 RowTypePtr skylineStruct(BaseTableCP baseTable, ColumnCP column) {
+  BitSet allFields;
+  if (auto fields = baseTable->controlSubfields.findSubfields(column->id())) {
+    allFields = *fields;
+  }
+  if (auto fields = baseTable->payloadSubfields.findSubfields(column->id())) {
+    allFields.unionSet(*fields);
+  }
+
+  const auto numOutputs = allFields.size();
   std::vector<std::string> names;
   std::vector<TypePtr> types;
-  std::unordered_set<std::string> distinct;
-  auto valueType = column->value().type->childAt(1);
+  names.reserve(numOutputs);
+  types.reserve(numOutputs);
 
   auto* ctx = queryCtx();
-  auto addTopFields = [&](const BitSet& paths) {
-    paths.forEach([&](int32_t id) {
-      auto path = ctx->pathById(id);
-      auto& first = path->steps()[0];
-      std::string name =
-          first.field ? std::string(first.field) : fmt::format("{}", first.id);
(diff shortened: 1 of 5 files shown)


MBkkt:
@mbasmanova I fixed it

MBkkt:
@oerling
Did you mean control/payload separation is needed for some reason?
Can you explain this reason for me please? In such case I probably just need to fix bug without refactoring.

I'm external contributor, so it's unclear to me what did you mean by reference to some internal PR/issue.

mbasmanova:
> Did you mean control/payload separation is needed for some reason?

I'll let Orri elaborate on this, but I believe this is needed for so-called late materialization. 

Google's AI explanation for "late materialization in query engines" is good.

mbasmanova:
@MBkkt Would it be possible to implement the improvement from #307 while keeping separate tracking for control and payload columns?

MBkkt:
> The idea is that we trim the query to keep only control columns (the ones that affect filtering, joins, etc.). Run this simplified query to get row IDs for final rows, then load the payload columns. To do that we need to know what "control" columns are and if they are complex type, what subfields from them are used to "control" the query flow.

Get it, thanks!

Although I thought about slightly different approach to late materialization, where columns loaded lazy on first access.
I also saw some code like this in velox. Is this approach supported?
I interested because we used such approach and clickhouse recently also implemented it.
I think it should works better for "single node" queries (and probably doesn't work for something like distributed join).

MBkkt:
@mbasmanova I reverted refactor part about remove control/payload separation.
Now it's only fix tests

axiom/optimizer/ToVelox.cpp
```diff
@@ -893,34 +893,30 @@ bool hasSubfieldPushdown(const TableScan& scan) {
...
-  auto fields = baseTable->controlSubfields.findSubfields(column->id());
-  if (fields.has_value()) {
-    addTopFields(fields.value());
-  }
-  fields = baseTable->payloadSubfields.findSubfields(column->id());
-  if (fields.has_value()) {
-    addTopFields(fields.value());
-  }
+  auto valueType = column->value().type->childAt(1);
+  allFields.forEach([&](int32_t id) {
```
mbasmanova, line 912: Would you clarify what is different now? Is it that the order of allFields decides the order of output columns?

MBkkt: There're two changes
1. usedChannels now return sorted list (It's the root of issue)
2. skylineStruct have similar issue, it change order of columns compared to implementation where control and payload is single map. So I also refactored it (as a bonus it should works better because doesn't have unordered set of strings)

mbasmanova (approved):
Thank you for the fix.

In the future, please, provide more details in the PR description to make review easier.

mbasmanova:
Would you rebase?

PR #313, opened 2025-08-29
feat: Support rewrite of complex type functions when they are accesse…

oerling:
…d as a whole

We have a case of f(make_row_from_map(c, ids, names)). f is unknown. When accessing the map as a row and projecting the columns of the row as top level columns, we expect make_row_from_map to become a row constructor.

This is different from the case of getters or oyther known functions over make_row_from_map, where the getter is replaced by the top level column and there is no make_row_from_map.

We add a hook to FunctionMetadata for this.

### DIFF
--- a/axiom/optimizer/Subfields.cpp
+++ b/axiom/optimizer/Subfields.cpp
@@ -340,7 +340,16 @@ void ToGraph::markSubfields(
     // If the function is some kind of constructor, like
     // make_row_from_map or make_named_row, then a path over it
     // selects one argument. If there is no path, all arguments are
-    // implicitly accessed.
+    // implicitly accessed. If the whole value of make_row_from_map is accessed,
+    // this does not yet access the whole map but just the keys listed in
+    // make_row_from_map.
+    if (metadata->expandFunction && steps.empty()) {
+      auto newExpr = metadata->expandFunction(call);
+      if (newExpr) {
+        markSubfields(newExpr, steps, isControl, context);
+        return;
+      }
+    }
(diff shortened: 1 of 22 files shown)


axiom/optimizer/Subfields.cpp
```diff
@@ -338,7 +338,16 @@ void ToGraph::markSubfields(
     // If the function is some kind of constructor, like
     // make_row_from_map or make_named_row, then a path over it
     // selects one argument. If there is no path, all arguments are
-    // implicitly accessed.
+    // implicitly accessed. If the whole value of make_row_from_map is accessed,
+    // this does not yet access the whole map but just the keys listed in
+    // make_row_from_map.
+    if (metadata->expandFunction && steps.empty()) {
```
MBkkt, line 344: I understand the idea of expandFunction, but I don't see use-cases except `make_row_from_some_args`?
Can you explain it for me?

axiom/optimizer/Subfields.cpp
```diff
@@ -338,7 +338,16 @@ void ToGraph::markSubfields(
...
+      auto newExpr = metadata->expandFunction(call);
+      if (newExpr) {
+        markSubfields(newExpr, steps, isControl, context);
+        return;
+      }
+    }
```
MBkkt, line 350: I understand how expandFunction use in this PR, but I don't see use-cases except some really unique.
The only one is access when access whole value isn't same as access all value members.

Can you explain for me when else it can be useful?
Otherwise maybe something more generic can be implemented?

PR #314, opened 2025-08-30
pass configuration from Connector to ConnectorMetadata for all connectors

hdikeman:
Summary:
We want to integrate existing configuration systems (TW and defined config attributes) with ConnectorMetadata

There are quite a few configuration values which we have deferred for lack of support

In this change I am ensuring that Hive (and by extension Prism and LocalHive), Impulse, TPCH, and the Test ConnectorMetadata store a reference to the corresponding Connector configuration

Differential Revision: D81365452


hdikeman:
Closing so I can split Verax and Velox changes

PR #316, opened 2025-09-01
refactor: Single source of truth for function names

MBkkt:
We're using presto functions with "presto_" suffix, so for us really important to have single place where all these functions defined.

Also I think it's important for future verax to be compatible with different velox functions.

### DIFF
--- a/axiom/optimizer/BuiltinNames.h
+++ b/axiom/optimizer/BuiltinNames.h
@@ -24,6 +24,13 @@ namespace facebook::velox::optimizer {
 /// Used for identifiers. Allows comparing strings by comparing pointers.
 using Name = const char*;
 
+/// This class should be created and access via queryCtx().
+/// It's needed for optimizer to have access to some hardcoded names.
+/// Right now it's names of functions with presto semantics.
+///
+/// Some members here are functions another variables: the reason is simple,
+/// it's function when it's accessed rarely in some specific code that can be
+/// disabled, e.g. join sample.
 struct BuiltinNames {
   BuiltinNames();
 
@@ -32,19 +39,26 @@ struct BuiltinNames {
   BuiltinNames& operator=(BuiltinNames&&) = delete;
   BuiltinNames& operator=(const BuiltinNames&) = delete;
 
+  // Returns names of "opposite" function if it's exists, e.g. gt for lt.
+  // Opposite means if arguments will be swapped it will be same semantics.
+  // If such name doesn't exist returns it's own input, e.g. eq for eq.
   Name reverse(Name op) const;
 
+  /// Returns true for function names that are commutative or
+  /// that have oppostite function.
   bool isCanonicalizable(Name name) const {
-    return canonicalizable.contains(name);
+    return canonicalizable_.contains(name);
   }
 
-  // Used for function registry
+  /// These function names used only for FunctionRegistry,
+  /// this is why they're not "Name" type but "std::string_view".
   static constexpr std::string_view kTransformValues = "transform_values";
   static constexpr std::string_view kTransform = "transform";
   static constexpr std::string_view kZip = "zip";
   static constexpr std::string_view kRowConstructor = "row_constructor";
 
-  // Used for join sample
+  /// These function names used only for JoinSample,
+  /// this is why they're not variables but functions.
   Name crc32() const;
   Name bitwiseAnd() const;
   Name bitwiseRightShift() const;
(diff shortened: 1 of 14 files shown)


axiom/optimizer/BuiltinNames.h
```diff
@@ -0,0 +1,83 @@
...
+
+#include <folly/container/F14Set.h>
+
+namespace facebook::velox::optimizer {
+
+/// Pointer to an arena allocated interned copy of a null terminated string.
+/// Used for identifiers. Allows comparing strings by comparing pointers.
+using Name = const char*;
+
+struct BuiltinNames {
```
mbasmanova, line 27: Let's document that these functions are expected to have the semantics of the Presto functions because Optimizer hard-codes these semantics. 

Let's also clarify that names here are function names, not any other names. Wondering if it makes sense to rename the class to something like BuiltinFunctionNames for clarity.

It would be helpful to also clarify that a single instance of this class is help in queryCtx() and must be accessed via queryCtx().builtinNames(). Perhaps, explain why these cannot be just constant strings.

axiom/optimizer/BuiltinNames.h
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /// These function names used only for FunctionRegistry,
+  /// this is why they're not "Name" type but "std::string_view".
+  static constexpr std::string_view kTransformValues = "transform_values";
+  static constexpr std::string_view kTransform = "transform";
+  static constexpr std::string_view kZip = "zip";
+  static constexpr std::string_view kRowConstructor = "row_constructor";
+
+  /// These function names used only for JoinSample,
```
mbasmanova, line 60: This class feels like a "dumping ground". I don't think it makes sense. 

If we need to maintain distinct sets of names for different purposes, then let's create separate classes for each set.

axiom/optimizer/FunctionRegistry.h
```diff
@@ -100,20 +100,18 @@ using FunctionMetadataCP = const FunctionMetadata*;
 
 class FunctionRegistry {
  public:
-  FunctionMetadataCP metadata(const std::string& name) const;
+  FunctionMetadataCP metadata(std::string_view name) const;
```
mbasmanova, line 103: This is a nice refactor. Would it be possible to extract it into a separate PR?

axiom/optimizer/BuiltinNames.h
MBkkt: I don't think we need separate sets of names.

These names quite same, I just tried to explain why I made them functions but not variables.

If you think it makes sense to make it more similar I'm ok with it and can change it to variables

axiom/optimizer/FunctionRegistry.h
MBkkt: Done https://github.com/facebookexperimental/verax/pull/332

axiom/optimizer/BuiltinNames.h
mbasmanova: I feel this change needs to be done differently and would be a much larger multi-PR refactoring.

You made a good point that we want to avoid hard-coded function names and rather allow applications to specify what function names they use.

Different names would require different handling.

FunctionRegistry might be simplest to start with. Currently it is populated by default with Presto functions (declareBuiltIn). Instead, we'd want to remove declareBuiltIn() call and require the application to explicitly populate the registry (or leave it empty). We'd want to provide an API to populate the registry with Presto functions (same as declareBuiltIn) and we'll use that in tests and velox_sql CLI.

JoinSample needs to be refactored to allow application to specify a hash function to use (either a Velox function or C++ function) and provide a default function (it shouldn't be composed from Presto functions, but rather written as a C++ function wrapped in Velox Simple Function).

Canonicalization logic needs to allow application to specify a list of binary operations that are reversible and a list of functions that can be flattened. The latter list is already available in Velox: https://facebookincubator.github.io/velox/develop/expression-evaluation.html#flatten-ands-and-ors

The optimizer needs to handle special forms specified in the logical plan (kAnd, kOr, etc.) without assuming they map to functions with fixed name. The application should be able to provide names for these special forms.

Hope this makes sense.

MBkkt: Yes, it absolutely makes sense to me.
But the issue, right now, in our fork to make current code works (for an example use hash join "eq"), we need to maintain patch with 10+ changed files.

BuiltinNames class already exists and used, I just make it a little more general and put in separate file.
It will be really helpful for us because it allows us to take a lot of less time to rebase.
Because only one file will be changed in our patch about function names.

As soon as suggested by you approach will be implemented I will be glad to use it.

Can we get more iterative approach on this, please?

PR #320, opened 2025-09-01
refactor: Remove unused members and remove logical prefix from members

MBkkt:
This was unused

```cpp
  std::function<std::unordered_map<PathCP, core::TypedExprPtr>(
      const core::CallTypedExpr* call,
      std::vector<PathCP>& paths)>
      explode;
```

### DIFF
--- a/axiom/optimizer/Optimization.h
+++ b/axiom/optimizer/Optimization.h
@@ -31,7 +31,7 @@ namespace facebook::velox::optimizer {
 class Optimization {
  public:
   Optimization(
-      const logical_plan::LogicalPlanNode& plan,
+      const logical_plan::LogicalPlanNode& logicalPlan,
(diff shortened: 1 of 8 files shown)


axiom/optimizer/Optimization.h
```diff
@@ -290,7 +290,7 @@ class Optimization {
   const bool isSingleWorker_;
 
   // Top level plan to optimize.
-  const logical_plan::LogicalPlanNode* const logicalPlan_;
+  const logical_plan::LogicalPlanNode* const plan_;
```
mbasmanova, line 293: same

mbasmanova (approved):
Thanks.

MBkkt:
@mbasmanova I applied your suggestions

PR #325, opened 2025-09-02
fix: Do not split aggregation into partial and file for single-thread-single-node execution

MBkkt:
Fix https://github.com/facebookexperimental/verax/issues/315


mbasmanova:
@MBkkt Would you rebase?

PR #326, opened 2025-09-02
feat(connectors): Store a reference to connector config in all ConnectorMetadata objects

hdikeman:
Summary:
Previously, only the LocalHive connector was storing a reference to config. I am moving this logic to the parent Hive class so that any child classes inherit the config passed via the ConnectorFactory

Also, making a similar change for the TPCH and the TestConnector metadata objects

Differential Revision: D81501093


axiom/optimizer/connectors/tests/TestConnector.cpp
```diff
@@ -72,6 +72,11 @@ std::shared_ptr<SplitSource> TestSplitManager::getSplitSource(
       tableHandle->connectorId(), partitions);
 }
 
+TestConnectorMetadata::TestConnectorMetadata(TestConnector* connector)
+    : connector_(connector),
+      config_(connector->connectorConfig()),
```
mbasmanova, line 77: Do we need to store it? Can we not get it on-demand?

axiom/optimizer/connectors/tpch/TpchConnectorMetadata.cpp
```diff
@@ -112,7 +112,9 @@ std::vector<SplitSource::SplitAndGroup> TpchSplitSource::getSplits(
 }
 
 TpchConnectorMetadata::TpchConnectorMetadata(TpchConnector* tpchConnector)
-    : tpchConnector_(tpchConnector), splitManager_(this) {}
+    : tpchConnector_(tpchConnector),
+      config_(tpchConnector->connectorConfig()),
```
mbasmanova, line 116: same question

axiom/optimizer/connectors/tests/TestConnector.cpp
hdikeman: Removed the TPCH changes so we can retrieve on-demand if needed

mbasmanova (approved):
Looks fine, but changes to TestConnector seem to be unrelated. In the future, let's make separate PRs for independent changes.

PR #335, opened 2025-09-03
fix: Disable some noisy .clang-tidy checks

MBkkt:
Disable checks that are not useful or requires change almost every code line.

### **altera-struct-pack-align**

Checks that packed data structures are used when cross-compiling to FPGAs. This check warns when a struct is not packed but could be packed for more efficient FPGA synthesis.

### **altera-unroll-loops**

Checks for opportunities to unroll loops to improve performance on FPGA targets. Suggests using pragma unroll or pragma unroll with a factor.

### **boost-use-ranges**

Suggests using Boost.Range algorithms instead of standard library algorithms where applicable for better performance and readability.

### **bugprone-easily-swappable-parameters**

Finds function declarations where parameters of the same type are adjacent, making it easy to accidentally swap them when calling the function. This can catch bugs where parameters are passed in the wrong order.

### **cppcoreguidelines-avoid-const-or-ref-data-members**

Warns about const or reference data members in classes, as they can make classes non-assignable and non-copyable, violating the C++ Core Guidelines' recommendations for regular types.

### **cppcoreguidelines-avoid-do-while**

Flags do-while loops as they can be error-prone compared to while or for loops, as recommended by the C++ Core Guidelines.

### **cppcoreguidelines-non-private-member-variables-in-classes**

Enforces that all data members in classes should be private, following object-oriented design principles from the C++ Core Guidelines.

### **cppcoreguidelines-pro-bounds-constant-array-index**

Flags array subscript operations that use non-constant indices, as they cannot be statically verified to be within bounds, violating the bounds safety guidelines.

### **llvm-qualified-auto**

Requires that `auto` variables have explicit qualifiers like `const` or `*` when their type is not immediately obvious from the initializer.

### **modernize-use-designated-initializers**

Suggests using C++20 designated initializers for aggregate initialization to make code more readable and maintainable.

### **performance-avoid-endl**

Warns about using `std::endl` instead of `'\n'` because `std::endl` flushes the stream buffer which can be slower, especially in loops.

### **performance-enum-size**

Suggests specifying explicit underlying types for enums to control their size and potentially improve performance and memory usage.

### **readability-function-cognitive-complexity**

Measures and flags functions that have high cognitive complexity (too many nested control structures, conditions, etc.) making them hard to understand and maintain.

### **readability-identifier-length**

Enforces minimum and maximum lengths for identifiers (variable names, function names, etc.) to improve code readability.

### **readability-math-missing-parentheses**

Suggests adding parentheses to mathematical expressions to clarify operator precedence and improve readability.

### **readability-qualified-auto**

Similar to `llvm-qualified-auto`, requires explicit qualifiers with `auto` to make the intended type more obvious to readers.


mbasmanova:
@MBkkt Wondering if we should keep the following checks:

- cppcoreguidelines-non-private-member-variables-in-classes
- cppcoreguidelines-pro-type-const-cast
- cppcoreguidelines-pro-type-static-cast-downcast
- readability-avoid-nested-conditional-operator
- readability-function-cognitive-complexity
- readability-math-missing-parentheses

mbasmanova:
Got it. Let keep these then. 

- cppcoreguidelines-pro-type-const-cast
- cppcoreguidelines-pro-type-static-cast-downcast
- readability-avoid-nested-conditional-operator

Thanks.

PR #339, opened 2025-09-03
feat: Allow application to control contents of the FunctionRegistry

mbasmanova:
Summary:
Do not register Presto functions by default. Allow applications to decide whether to register Presto functions or some other functions.

Part of https://github.com/facebookexperimental/verax/issues/336

Differential Revision: D81596283


xiaoxmeng (approved):
@mbasmanova thanks!

PR #349, opened 2025-09-04
feat: Allow to customize element_at, subscript and cardinality

mbasmanova:
Part of https://github.com/facebookexperimental/verax/issues/336

Differential Revision: D81676895

### DIFF
--- a/axiom/optimizer/FunctionRegistry.cpp
+++ b/axiom/optimizer/FunctionRegistry.cpp
@@ -15,6 +15,7 @@
  */
 
 #include "axiom/optimizer/FunctionRegistry.h"
+#include "velox/expression/ExprConstants.h"
 
 namespace facebook::velox::optimizer {
 namespace lp = facebook::velox::logical_plan;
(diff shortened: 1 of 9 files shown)


axiom/optimizer/FunctionRegistry.cpp
```diff
@@ -165,21 +193,27 @@ void FunctionRegistry::registerPrestoFunctions(std::string_view prefix) {
...
   registry->registerSpecialForm(lp::SpecialForm::kTryCast, "trycast");
   registry->registerSpecialForm(lp::SpecialForm::kTry, "try");
   registry->registerSpecialForm(lp::SpecialForm::kIf, "if");
-  registry->registerSpecialForm(lp::SpecialForm::kCoalesce, "coalesce");
-  registry->registerSpecialForm(lp::SpecialForm::kSwitch, "switch");
-  registry->registerSpecialForm(lp::SpecialForm::kIn, "in");
+  registry->registerSpecialForm(
+      lp::SpecialForm::kCoalesce, fullName("coalesce"));
+  registry->registerSpecialForm(lp::SpecialForm::kSwitch, fullName("switch"));
+  registry->registerSpecialForm(lp::SpecialForm::kIn, fullName("in"));
```
MBkkt, line 216: Why does `and`, `or`, etc without `fullName`  and `coalesce`, `switch`, `in` with `fullName`.

I thought only velox special form callable should be without prefix, so I understand why `and` and `or`, etc without fullname. But I don't understand why `coalesce` and `switch` also with fullname

```
constexpr const char* kAnd = "and";
constexpr const char* kOr = "or";
constexpr const char* kSwitch = "switch";
constexpr const char* kIn = "in";
constexpr const char* kIf = "if";
constexpr const char* kFail = "fail";
constexpr const char* kCoalesce = "coalesce";
constexpr const char* kCast = "cast";
constexpr const char* kTryCast = "try_cast";
constexpr const char* kTry = "try";
```

MBkkt (commented):
Just wanted to say, thanks a lot for fixing this!

axiom/optimizer/FunctionRegistry.cpp
MBkkt: Maybe it's better to fix this in velox and allow to have prefix for any functions on registration

mbasmanova: @MBkkt Good catch. Let me switch to using constants from velox/expression/ExprConstants.h

PR #352, opened 2025-09-04
refactor: Expr and LogicalPlanNode should accept parameters by value and move

MBkkt:
In such case we avoid extra copies when we build logical plan.

Also I fixed some code to avoid unnecessary dynamic_cast

### DIFF
--- a/axiom/logical_plan/LogicalPlanNode.cpp
+++ b/axiom/logical_plan/LogicalPlanNode.cpp
@@ -240,13 +240,6 @@ const auto& setOperationNames() {
 
 VELOX_DEFINE_ENUM_NAME(SetOperation, setOperationNames)
 
-SetNode::SetNode(
-    std::string id,
-    const std::vector<LogicalPlanNodePtr>& inputs,
-    SetOperation operation)
-    : LogicalPlanNode{NodeKind::kSet, std::move(id), inputs, makeOutputType(inputs)},
-      operation_{operation} {}
-
 // static
 RowTypePtr SetNode::makeOutputType(
     const std::vector<LogicalPlanNodePtr>& inputs) {
--- a/axiom/logical_plan/LogicalPlanNode.h
+++ b/axiom/logical_plan/LogicalPlanNode.h
@@ -544,7 +544,9 @@ class SetNode : public LogicalPlanNode {
   SetNode(
       std::string id,
       const std::vector<LogicalPlanNodePtr>& inputs,
-      SetOperation operation);
+      SetOperation operation)
+      : LogicalPlanNode{NodeKind::kSet, std::move(id), inputs, makeOutputType(inputs)},
+        operation_{operation} {}
 
   SetOperation operation() const {
     return operation_;


axiom/logical_plan/Expr.h
```diff
@@ -109,15 +109,18 @@ class Expr {
     return kind_ == ExprKind::kSubquery;
   }
 
+  // TODO: better to return reference to avoid ability to check result
```
mbasmanova, line 112: I'm not sure about this. I think it is helpful to allow calling this without checking type kind first. Would it be possible to remove this from this PR?

(I would rename this to 'as' though.)

axiom/logical_plan/LogicalPlanNode.cpp
```diff
@@ -244,26 +241,26 @@ const auto& setOperationNames() {
...
+    const auto& rowType = inputs_[i]->outputType();
 
     // The names are different, but types must be the same.
     VELOX_USER_CHECK(
         firstRowType->equivalent(*rowType),
         "Output schemas of all inputs to a Set operation must match");
 
     // Individual column types must match exactly.
-    for (auto j = 0; j < firstRowType->size(); ++j) {
+    for (uint32_t j = 0; j < firstRowType->size(); ++j) {
```
mbasmanova, line 263: Why can't we use auto?

axiom/logical_plan/LogicalPlanNode.h
```diff
@@ -72,9 +89,12 @@ class LogicalPlanNode {
     return kind_ == kind;
   }
 
+  // TODO: better to return reference to avoid ability to check result
```
mbasmanova, line 92: same as above

mbasmanova (commented):
Thanks. Some comments.

axiom/logical_plan/Expr.h
MBkkt: But dynamic cast performance can be painful. It's already avoided

mbasmanova: > But dynamic cast performance can be painful.

True... when it matters. In most cases it doesn't.

MBkkt: >  I think it is helpful to allow calling this without checking type kind first.

In such case users will use dynamic_cast, even when it's not needed.
And still needs to check result of this dynamic_cast, so I don't think it's more convinient.

In best we compare such scenarios

```cpp
if (some.isCall()) {
  doSome(*some.asUnchecked<Call>());
}
```

vs

```cpp
if (auto* call = some.as<Call>()) {
  doSome(*call);
}
```

and if we will change interface to produce reference instead of pointer (it will be explicit without "Unchecked".
```cpp
if (some.isCall()) {
  doSome(some.as<Call>());
}
```

axiom/logical_plan/LogicalPlanNode.cpp
mbasmanova: (I feel 'auto' is better because it is easy to write and read. Specific type is hard to get right.)

axiom/logical_plan/LogicalPlanNode.h
```diff
@@ -52,11 +52,28 @@ class LogicalPlanNode {
...
+        id_{std::move(id)},
+        inputs_{std::move(inputs)},
+        outputType_{std::move(outputType)} {
+    VELOX_USER_CHECK_NOT_NULL(outputType_);
+    for (const auto& input : inputs_) {
+      VELOX_USER_CHECK_NOT_NULL(input);
+    }
+  }
+
+  LogicalPlanNode(
```
mbasmanova, line 68: Why did you add this ctor? Can we remove this? It would be helpful to keep that PR limited to the changes description in title + description.

MBkkt: Without this ctor for some other ctor move wasn't possible

MBkkt: Example: SetNode

axiom/logical_plan/LogicalPlanNode.cpp
MBkkt: But the with `auto` type is kind of incorrect, I can write `auto i = 0U` if you think it's better

To create `size_t` with literal we can write `0UZ`, but only starting from C++23

https://en.cppreference.com/w/cpp/language/integer_literal.html

axiom/logical_plan/LogicalPlanNode.h
mbasmanova: Yess.. I now remember why I didn't pass by value + move. I wanted to avoid troubles like this and I don't think the extra copies would show up anywhere. Do they?

mbasmanova: This new ctor is confusing. Would it be possible to remove it and not use std::move in SetNode?

MBkkt: I noticed because our transformer code as it supposed to be written fails (vector of strings was copied instead of move so string_view in some other place doesn't work).

I didn't measure this, but I don't think it's needed we can just have clear "by value" API, why not?
Is it something in implementation that you didn't like? I can try to figure out how to change it.

I also think then it comes to fast queries (less than 1-5ms) it's desirable to have cheap API in different places.
After we moved to axiom we noticed significant (in a few times for simple queries) slowdown in our sqllogic tests compared to our old implementation (that was more OLTP like rule based optimizer), this slowdown doesn't depend on query, it's just large "setup cost".
I didn't investigate reason yet, plan to do it in next few months.
But IMO it's nothing bad if we will make everything around core algorithms cheap if it's still allow us to have clear API.

MBkkt: > This new ctor is confusing. Would it be possible to remove it and not use std::move in SetNode?

Ok, I figure out it without new ctor

axiom/logical_plan/LogicalPlanNode.cpp
```diff
@@ -244,26 +241,26 @@ const auto& setOperationNames() {
 VELOX_DEFINE_ENUM_NAME(SetOperation, setOperationNames)
 
 SetNode::SetNode(
-    const std::string& id,
+    std::string id,
     const std::vector<LogicalPlanNodePtr>& inputs,
     SetOperation operation)
-    : LogicalPlanNode(NodeKind::kSet, id, inputs, inputs.at(0)->outputType()),
+    : LogicalPlanNode{NodeKind::kSet, std::move(id), inputs, inputs.empty() ? nullptr : inputs[0]->outputType()},
```
mbasmanova, line 247: > inputs.empty() ? nullptr

inputs cannot be empty; perhaps, use a helper function to check that inputs is not empty and return first input's output type

mbasmanova: smth like

```
LogicalPlanNode{NodeKind::kSet, std::move(id), inputs, toSetOutputType(inputs))
```

MBkkt: Done

axiom/logical_plan/LogicalPlanNode.cpp
```diff
@@ -244,26 +241,30 @@ const auto& setOperationNames() {
...
+    std::string id,
     const std::vector<LogicalPlanNodePtr>& inputs,
     SetOperation operation)
-    : LogicalPlanNode(NodeKind::kSet, id, inputs, inputs.at(0)->outputType()),
-      operation_{operation} {
+    : LogicalPlanNode{NodeKind::kSet, std::move(id), inputs, makeOutputType(inputs)},
+      operation_{operation} {}
+
+// static
+RowTypePtr SetNode::makeOutputType(
```
mbasmanova, line 251: nit: this can be a free-standing function; doesn't need to be a method of the SetNode class.

mbasmanova (approved):
Thanks.

axiom/logical_plan/LogicalPlanNode.cpp
MBkkt: I made it same way as join, unnest, project, aggregate.

To make it required and more similar to them, I moved constructor in the header
https://github.com/facebookexperimental/verax/pull/352/commits/227c9dd1914ae2903fd107fa417272fba2ebd320

MBkkt:
@mbasmanova done

PR #354, opened 2025-09-04
fix: Some presto functions should be without prefix

MBkkt:


axiom/optimizer/FunctionRegistry.cpp
```diff
@@ -164,31 +164,36 @@ void FunctionRegistry::registerPrestoFunctions(std::string_view prefix) {
...
     metadata->cost = 20;
-    registerFunction("zip", std::move(metadata));
+    registerFunction(fullName("zip"), std::move(metadata));
   }
 
   {
     auto metadata = std::make_unique<FunctionMetadata>();
     metadata->valuePathToArgPath = rowConstructorSubfield;
     metadata->explode = rowConstructorExplode;
+    // presto row_constructor created without prefix
```
mbasmanova, line 196: Make sure all comments are full sentences. Start with a capital letter and end with a period. Presto should always be spelled with a capital P.

(Would be nice to add expression::kRowConstructor to Velox.)

axiom/optimizer/FunctionRegistry.cpp
```diff
@@ -215,7 +221,7 @@ void FunctionRegistry::registerPrestoFunctions(std::string_view prefix) {
   registry->registerSpecialForm(
       lp::SpecialForm::kCoalesce, expression::kCoalesce);
   registry->registerSpecialForm(lp::SpecialForm::kSwitch, expression::kSwitch);
-  registry->registerSpecialForm(lp::SpecialForm::kIn, fullName("in"));
+  registry->registerSpecialForm(lp::SpecialForm::kIn, expression::kIn);
```
mbasmanova, line 224: I think "in" should have a prefix.

axiom/optimizer/ToVelox.cpp
```diff
@@ -167,18 +167,12 @@ void ToVelox::filterUpdated(BaseTableCP table, bool updateSelectivity) {
...
-    } else {
-      remainingFilter = std::make_shared<core::CallTypedExpr>(
-          BOOLEAN(),
-          specialForm(lp::SpecialForm::kAnd),
-          remainingFilter,
-          conjunct);
-    }
-  }
+  auto remainingFilter = remainingConjuncts.empty() ? nullptr
+      : remainingConjuncts.size() == 1              ? remainingConjuncts[0]
```
mbasmanova, line 171: I find this hard to read. Let's use if-else.

axiom/optimizer/ToVelox.cpp
```diff
@@ -431,34 +421,33 @@ core::TypedExprPtr ToVelox::toTypedExpr(ExprCP expr) {
...
+        return std::make_shared<core::CastTypedExpr>(
+            toTypePtr(expr->value().type), std::move(inputs), true);
+      }
 
-        return std::make_shared<core::CallTypedExpr>(
-            toTypePtr(expr->value().type),
-            std::move(inputs),
-            specialForm(form.value()));
+      std::string name;
+      if (auto form = SpecialFormCallNames::tryFromCallName(call->name())) {
```
mbasmanova, line 443: I specifically called tryFromCallName before checking for cast and try_cast to avoid extra checks.

mbasmanova (commented):
Thanks.

axiom/optimizer/ToVelox.cpp
MBkkt: There's no extra check. I think from compiler perspective it's same code
So if you like previous version more I will rollback this

MBkkt:
@mbasmanova I applied your suggestions

MBkkt:
@mbasmanova Do I need to fix something here?

MBkkt:
@mbasmanova done

PR #356, opened 2025-09-05
fix: Push-down limit physical node if limit small enough

MBkkt:


axiom/optimizer/Optimization.cpp
```diff
@@ -698,17 +698,23 @@ void Optimization::addPostprocess(
...
+    auto limit = make<Limit>(plan, dt->limit, dt->offset);
+    state.addCost(*limit);
+    plan = limit;
   }
   if (!dt->columns.empty()) {
     auto* project = make<Project>(plan, dt->exprs, dt->columns);
     plan = project;
   }
-  if (!dt->hasOrderBy() && dt->hasLimit()) {
+  if (dt->limit > kMaxLimitBeforeProject) {
```
mbasmanova, line 717: Should we keep "!dt->hasOrderBy()" check here? Looks like we might be missing test coverage.

MBkkt: Yes, I tried to write test and find this.

> If arrives after limit, then starts a new DT.

Is it actually corrects?
It prevents us from getting TopN for

```
    auto logicalPlan = lp::PlanBuilder{}
                           .tableScan(kTestConnectorId, "t", {"a", "b"})
                           .limit(10)
                           .orderBy({"a"})
                           .build();
```

```
    case lp::NodeKind::kSort:
      // Multiple orderBys are allowed before a limit. Last one wins. Previous
      // are dropped. If arrives after limit, then starts a new DT.

      makeQueryGraph(*node.onlyInput(), allowedInDt);

      if (currentDt_->hasLimit()) {
        finalizeDt(*node.onlyInput());
      }

      return addOrderBy(*node.asUnchecked<lp::SortNode>());
```

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1273,6 +1273,84 @@ TEST_F(PlanTest, values) {
...
+                           .tableScan(kTestConnectorId, "t", {"a", "b"})
+                           .project({"a + b as c"})
+                           .orderBy({"c"})
+                           .limit(10)
+                           .build();
+
+    auto plan = toSingleNodePlan(logicalPlan);
+
+    // Extra projection is a bug:
+    // https://github.com/facebookexperimental/verax/issues/357
```
mbasmanova, line 1321: Thank you for creating an issue for this.

mbasmanova (approved):
Thanks.

PR #370, opened 2025-09-08
feat(ExprPrinter): Implement AggregateExpr printer

pashandor789:


mbasmanova (commented):
@pashandor789 Looks good. Thank you for adding this logic. Would you add a test to axiom/logical_plan/tests/PlanPrinterTest.cpp?

pashandor789:
@mbasmanova can you take a look ?

PR #375, opened 2025-09-09
fix: Add AXIOM_MONO_LIBRARY

MBkkt:
Adding AxiomUtils to make possible to configure
1. multiple velox targets and multiple axiom targets
2. single velox target and multiple axiom targets
3. multiple velox targets and single axiom target
4. single velox target and single axiom targets

This PR also contains fixes from https://github.com/facebookincubator/axiom/pull/460


Makefile
```diff
@@ -17,7 +17,7 @@ BUILD_BASE_DIR=_build
 BUILD_DIR=release
 BUILD_TYPE=Release
 
-CMAKE_FLAGS := -DCMAKE_BUILD_TYPE=$(BUILD_TYPE)
+CMAKE_FLAGS := -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) -DAXIOM_BUILD_TESTING=ON -DVELOX_MONO_LIBRARY=ON -DAXIOM_MONO_LIBRARY=ON
```
MBkkt, line 20: I can remove it after PR will pass tests, if you think it's better without it
```suggestion
CMAKE_FLAGS := -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) -DAXIOM_BUILD_TESTING=ON -DVELOX_MONO_LIBRARY=ON
```
or
```suggestion
CMAKE_FLAGS := -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) -DAXIOM_BUILD_TESTING=ON
```

mbasmanova:
@assignUser @kgpai Jacob, Krishna, would you help review this change?

@MBkkt Would you update PR description to explain what is the effect of this change and why is it beneficial? Thanks.

axiom/optimizer/connectors/hive/CMakeLists.txt
```diff
@@ -38,6 +38,6 @@ velox_link_libraries(
   velox_gcs
   velox_abfs)
 
-if(${VELOX_BUILD_TESTING})
```
MBkkt, line 41: this was false in github CI

MBkkt:
> Why would it be helpful to use VELOX_MONO_LIBRARY? What does that do?

It's already exist in velox, it's used to produce single `libvelox.a` or `libvelox.so` instead of hundreds small `.so` and `.a` targets. It's useful to allow compiler/linker/lto to better optimize code and speedup their work.
It also decrease sizeof of build directory.

It mainly used by users of velox.

> Would you provide an example? Would you like to introduce a naming convention for CMake targets? That would be nice, but let's document it in https://github.com/facebookexperimental/verax/blob/main/CONTRIBUTING.md

`velox_verax`, `velox_fe_*`, but `axiom_logical_plan`, `axiom_sql_parser_*`, etc

I'm not sure what to write contributing, and honestly don't think it's really necessary.
New targets commonly "done" same way as previously existed targets.
So I think we just need to make current state consistent.

> It sounds there are 3 separate changes in this PR. Would you split these into 3 separate PRs?

The naming is deeply related to `axiom_add_library`, so it cannot be separated.

But I think I can try to separate `if (AXIOM_ENABLE_TESTING)`.

assignUser (changes requested):
I really appreciate your efforts here but just vendoring the Velox CMake  is not the right way to do this.

The wrapper functions in Velox are a workaround because the codebase was already massive and had 100+ targets. Reducing/combining those targets is a longterm goal but wasn't feasible in the short-term, hence the workaround with the mono library.

This is not the case for verax/axiom (btw what is the right name here for what? :grin:), we can just create fewer more logical targets from the start and keep think organized. I am not familiar with the code base so I don't really know what that amount is, it depends on intended usage etc. (e.g. could be libverax-optimizer, libverax-runner,  or something like that or just a single library if the use case would always require everything)

MBkkt:
@assignUser 

> This is not the case for verax/axiom (btw what is the right name here for what? 😁), we can just create fewer more logical targets from the start and keep think organized. I am not familiar with the code base so I don't really know what that amount is, it depends on intended usage etc. (e.g. could be libverax-optimizer, libverax-runner, or something like that or just a single library if the use case would always require everything)

The name is correct, as far as I know (see https://github.com/facebookexperimental/verax/issues/211).

I don't see why do we need multiple targets for this. Maybe someone need, in such case they can build with AXIOM_MONO=OFF.
Right now already a lot of targets. 

Also without this PR using verax is kind of painful, because VELOX_MONO just doesn't work.


Can we just have solution that 
1. works with VELOX_MONO
2. can produce single target
3. can produce multiple target

Later we can refactor this.

MBkkt:
With static build from source we don't need at all separate targets.
Because linker can drop unnecessary symbols/code itself.

At the same time, build + link time with single target (as dependency) is reduced compared to multiple dependent targets

assignUser:
> It's worse solution, it produces a lot of unnecessary targets

That's why I said ***short-term*** solution. The real solution is, as I explained above, to reduce the number of targets that are created. This codebase is small enough to just make that change instead of making use of a workaround from another codebase. 

> Also with separate targets build directory in CI takes a lot of more space.

I don't know why you are trying to convince me that less targets are better. I know, I was the one to write `velox_add_library` after all ;). I am just saying that using the velox workaround is not the way to go here.

MBkkt:
@assignUser I created this discussion for velox https://github.com/facebookincubator/velox/discussions/14807

MBkkt:
>  I am just saying that using the velox workaround is not the way to go here.

Why? What is bad about it, except it doesn't work for PUBLIC/PRIVATE/INTERFACE (can be fixed later, same in velox)

MBkkt:
Well in theory we can make add_library( OBJECT) for separated targets, and when reference them in separated STATIC/SHARED libraries + in single(common) library.

But it's complicated plus cmake doesn't work for OBJECT with cyclic dependecies

assignUser:
> It doesn't solve **my** problem, it doesn't produce single axiom target

Ok? The solution for that is to create a config for axiom like we just started adding in velox which will give you one `Axiom::axiom` main target. (We could also add an interface target like that for use with FetchContent).

> So why don't you want to allow us (in serenedb) to build with single axiom target?

I don't. I just want to do it in an idiomatic CMake way instead of vendoring tech-debt from another project. (even if I created said tech-debt ;))

> Well in theory we can make add_library( OBJECT) 

I agree OBJECT libraries are absolutely not the solution.

I understand your need for a single target you can link against in your downstream project and I agree, that is much better DX than having to check which of the n targets is needed (well maybe outside from high-level components like libvelox-core, libvelox-gpu, ... but that's not relevant here). But let's solve the actual problem in the right way.

MBkkt:
Maybe I can try to explain it in a different way.
I think this PR solves all current issues (in terms of desired binaries produced and it doesn't fail to build), maybe not the best way from cmake perspective, but still.

If someone want and can refactor it in idiomatic cmake, ofc it will be good.
But the issue even if it's possible (I'm not sure), it takes a lot of time.
Who will spend time on this?

Idiomatic cmake unfortunately doesn't get anything to our project.
So I don't have time to make cmake ideal. 
I made this cmake PR only because it's needed to solve the concrete issues.
Not because I want refactor something.

And I think it makes current state better than in main branch from any perspective (even from cmake).

If it's not possible to merge this, ok, well I probably can do anything else about it, and will maintain it as a patch

MBkkt:
> No, interface targets are virtual targets that are not actually build so the only influence they have is on configure/generate.

With interface target we will link against N small libraries, and we will build N small libraries.

Interface is just convenient way to link against all libraries.

It's not the same as single library

MBkkt:
> I will but won't have time for a bit to do it

This the whole point, no one, at least for now, have time for this.

I just want to solve build issues and make it is a way better than in current main.

PR #376, opened 2025-09-09
fix: Introduce AXIOM_BUILD_TESTING CMake variable

MBkkt:
Allow building Axiom with and without tests.

### DIFF
--- a/.cmake-format.yaml
+++ b/.cmake-format.yaml
@@ -31,7 +31,7 @@ format:
     - target_sources
     - target_link_libraries
 parse:
-  # No custom functions defined - axiom uses standard CMake functions
+  # No custom functions defined - Axiom uses standard CMake functions
   additional_commands: {}
 markup:
   first_comment_is_literal: true
--- a/.cmake-format.yaml
+++ b/.cmake-format.yaml
@@ -31,7 +31,7 @@ format:
     - target_sources
     - target_link_libraries
 parse:
-  # No custom functions defined - verax uses standard CMake functions
+  # No custom functions defined - axiom uses standard CMake functions
   additional_commands: {}
 markup:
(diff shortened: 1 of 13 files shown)


axiom/optimizer/connectors/hive/tests/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ target_link_libraries(
   velox_hive_connector_metadata_test
   velox_hive_connector_metadata
   velox_hive_connector
-  velox_exec_runner_test_util
+  axiom_runner_tests_utils
```
MBkkt, line 22: it was bug, this cmake doesn't work, but it wasn't revealed because different bug

axiom/optimizer/connectors/hive/CMakeLists.txt
```diff
@@ -38,6 +38,6 @@ velox_link_libraries(
   velox_gcs
   velox_abfs)
 
-if(${VELOX_BUILD_TESTING})
+if(AXIOM_BUILD_TESTING)
```
MBkkt, line 41: ditto

.cmake-format.yaml
```diff
@@ -31,7 +31,7 @@ format:
     - target_sources
     - target_link_libraries
 parse:
-  # No custom functions defined - verax uses standard CMake functions
+  # No custom functions defined - axiom uses standard CMake functions
```
mbasmanova, line 34: nit: axiom -> Axiom

mbasmanova (approved):
Thanks!

.cmake-format.yaml
MBkkt: fixed

feilong-liu (approved):

PR #378, opened 2025-09-09
refactor: Reduce dependency on velox::connector::Connector::metadata() API

mbasmanova:
Summary: First step towards removing velox::connector::Connector::metadata() API

Differential Revision: D82032893

### DIFF
--- a/axiom/optimizer/connectors/hive/LocalHiveConnectorMetadata.cpp
+++ b/axiom/optimizer/connectors/hive/LocalHiveConnectorMetadata.cpp
@@ -47,7 +47,7 @@ std::shared_ptr<SplitSource> LocalHiveSplitManager::getSplitSource(
   // Since there are only unpartitioned tables now, always makes a SplitSource
   // that goes over all the files in the handle's layout.
   auto tableName = tableHandle->name();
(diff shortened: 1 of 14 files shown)


axiom/optimizer/connectors/hive/LocalHiveConnectorMetadata.cpp
```diff
@@ -277,9 +277,9 @@ std::pair<int64_t, int64_t> LocalHiveTableLayout::sample(
 
   const auto outputType = ROW(std::move(names), std::move(types));
 
-  auto connectorQueryCtx =
-      reinterpret_cast<LocalHiveConnectorMetadata*>(connector()->metadata())
-          ->connectorQueryCtx();
+  auto connectorQueryCtx = reinterpret_cast<LocalHiveConnectorMetadata*>(
```
MBkkt, line 280:
```suggestion
  auto connectorQueryCtx = static_cast<LocalHiveConnectorMetadata*>(
```
It's wrong logically and can be wrong in practice if sometime `ConnectorMetadata` won't be first base class of `LocalHiveConnectorMetadata`

PR #379, opened 2025-09-09
refactor: Rename namespace

MBkkt:
fix https://github.com/facebookexperimental/verax/issues/377


mbasmanova:
> Also axiom::sql::presto => facebook::axiom::sql::presto, to make it consistent

We don't want 'facebook::' prefix. Let's remove it.

MBkkt:
> We don't want 'facebook::' prefix. Let's remove it.

Do you sure? In such case everywhere where used `velox::` needs to be written `facebook::velox::`.
Using in header also doesn't look like a good solution.

I think ideally will be replace `facebook::velox` with `velox` in velox.
But I doubt it's possible :(

MBkkt:
@mbasmanova So how I should handle this?
1. Write everywhere `facebook::velox` instead of `velox`
2. Put in some common header `using velox = facebook:velox`
3. Keep it `facebook::axiom` instead of just `axiom`

MBkkt:
@mbasmanova I rebased + applied your suggestion.

But I wanted to notice, with such solution using together `axiom::sql` and `facebook::axiom::runner/optimizer/logical_plan` can be inconvenient, because it will be different namespaces with same name.

I want only notice this, because to me there's no real difference (we don't plan to use `axiom::sql`), so it's as you requested.

mbasmanova:
Thank you, @MBkkt. I do hope we'll change all namespaces in Axiom to drop facebook:: prefix.

mbasmanova (approved):
Thanks.

PR #385, opened 2025-09-10
fix: Allow building with VELOX_MONO_LIBRARY

MBkkt:
@mbasmanova I copied PR https://github.com/facebookexperimental/verax/pull/375 and applied @assignUser suggestion to make cmake "modern".

Unfortunately such solution is worse in terms of produced binaries and usability for us in serenedb, but even in such case it's still better than current main.


MBkkt:
I don't need this in this way, so close it

PR #387, opened 2025-09-10
fix: Make cmake self consistent

MBkkt:
* Use velox_add_library (will be replaced with axiom_add_library) for all public libraries
* Use object library for connector (same as for other connectors)
* Remove unnecessary `add_dependencies`

### DIFF
--- a/axiom/optimizer/tests/CMakeLists.txt
+++ b/axiom/optimizer/tests/CMakeLists.txt
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-velox_add_library(axiom_optimizer_tests_parquet_tpch ParquetTpchTest.cpp)
+add_library(axiom_optimizer_tests_parquet_tpch ParquetTpchTest.cpp)
 
-velox_link_libraries(
+target_link_libraries(
   axiom_optimizer_tests_parquet_tpch
   axiom_logical_plan_builder
   axiom_optimizer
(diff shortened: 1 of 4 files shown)


axiom/optimizer/tests/CMakeLists.txt
```diff
@@ -100,7 +101,6 @@ add_executable(
   HiveQueriesTest.cpp
   PlanTest.cpp
   ParquetTpchTest.cpp
-  QueryTestBase.cpp
```
MBkkt, line 103: this was unnecessary, we're linked against library that contains this code

mbasmanova (approved):
@MBkkt Thanks.

assignUser (changes requested):
This moves into the wrong direction in my opinion as the build maintainer. As discussed in #375 at length, even if @MBkkt disagrees.

I still don't understand what the issue is with regards to multiple libraries and targets. Sure Abseil goes a bit far but having an interface target and multiple libraries is fine.

What issue do you have with multiple libraries (other than size, which again should be mostly addressed by using VELOX_MONO_LIBRARY, which is not possible at the moment.)? It is common practice to split a library into multiple libraries along components with different scope or additional dependencies e.g. libarrow, libarrow-filesystem, libarrow-dataset, etc.

Can you please articulate what exactly your issue is that you are trying to solve and *WHY* you need to solve it other then taste.

MBkkt:
@assignUser I think my main point is although VeloxUtils is workaround, it doesn't do anything really scary.
And at the same time it allows developers to write not best cmake (bad target separations, too granular targets, etc) and still have good metrics about build time, build directory size, etc

So I think it makes sense to have same workaround in Axiom.

PR #399, opened 2025-09-15
fix: Use function parameters instead of gflags for numDrivers and numWorkers in unittests

MBkkt:
Also
* test more cases and have better output in case of failure
* remove unnecessary namespaces


axiom/optimizer/tests/QueryTestBase.cpp
```diff
@@ -191,15 +191,6 @@ std::shared_ptr<core::QueryCtx> QueryTestBase::getQueryCtx() {
   return queryCtx_;
 }
 
-optimizer::PlanAndStats QueryTestBase::planVelox(
```
mbasmanova, line 194: I assume some call sites are affected by this change. Should we identify these and make sure they continue to work as before?

axiom/optimizer/tests/SubfieldTest.cpp
```diff
@@ -28,7 +28,7 @@
 DEFINE_string(subfield_data_path, "", "Data directory for subfield test data");
 DECLARE_uint32(optimizer_trace);
 
-DECLARE_int32(num_workers);
+DECLARE_uint32(num_workers);
```
mbasmanova, line 31: Shouldn't we delete this and change the code in toSingleNodePlan to pass 'options'?

```
core::PlanNodePtr toSingleNodePlan(
      const lp::LogicalPlanNodePtr& logicalPlan) {
    gflags::FlagSaver saver;
    FLAGS_num_workers = 1;

    auto plan = planVelox(logicalPlan).plan;
```

axiom/optimizer/tests/TpchPlanTest.cpp
```diff
@@ -25,7 +25,7 @@
 DEFINE_int32(num_repeats, 1, "Number of repeats for optimization timing");
 
 DECLARE_uint32(optimizer_trace);
-DECLARE_int32(num_workers);
+DECLARE_uint32(num_workers);
```
mbasmanova, line 28: Wondering if we can simply delete this.

MBkkt:
@mbasmanova I applied your suggestions and rebased

axiom/optimizer/tests/QueryTestBase.cpp
MBkkt: fixed, (default value is same as it was before my patch)

axiom/optimizer/tests/TpchPlanTest.cpp
MBkkt: Fixed

MBkkt:
@mbasmanova I removed from this PR anything that relates to this issue https://github.com/facebookexperimental/verax/issues/413 and updated PR description

Is it ok now?

axiom/optimizer/tests/SubfieldTest.cpp
```diff
@@ -364,11 +362,11 @@ class SubfieldTest : public QueryTestBase,
   }
 
   core::PlanNodePtr toSingleNodePlan(
-      const lp::LogicalPlanNodePtr& logicalPlan) {
-    gflags::FlagSaver saver;
-    FLAGS_num_workers = 1;
-
-    auto plan = planVelox(logicalPlan).plan;
+      const lp::LogicalPlanNodePtr& logicalPlan,
+      int32_t numDrivers = 1) {
```
mbasmanova, line 366: Do we have any callers that specify this argument? If not, perhaps, we can remove it.

axiom/optimizer/tests/PlanTest.h
```diff
@@ -76,25 +76,43 @@ class PlanTest : public test::QueryTestBase {
     VELOX_CHECK_NOT_NULL(planNode);
     VELOX_CHECK_NOT_NULL(referencePlan);
 
-    auto fragmentedPlan = planVelox(planNode, options);
-    auto referenceResult = assertSame(referencePlan, fragmentedPlan);
-
-    if (options.numWorkers != 1) {
-      auto singleNodePlan = planVelox(
+    auto referenceResult = runVelox(referencePlan);
```
mbasmanova, line 79: I didn't realize we have PlanTest.h. Would it be possible to delete this file and move the code to .cpp?

MBkkt: Ok.

PR #400, opened 2025-09-15
feat(connectors): Allow TPCH ConnectorMetadata to resolve any scale factor

hdikeman:
Summary:
Previously, since ConnectorMetadata returned a naked pointer to callers of findTable, we had to create TPCH tables up front and keep the references in a table array

Now that we return a shared pointer from findTable, we can create TPCH tables on demand so long as the scale factor requested is a valid one

The motivation is two-fold: (1) this brings us into spec with the Presto frontend, which permits any integer-valued scale factor, and (2) we would like to benchmark scale factors in the intermediate range (10, 25, 30). sf1 is too small and sf100 is too large for local benchmarking

Differential Revision: D82475141

### DIFF
--- a/axiom/connectors/tpch/TpchConnectorMetadata.cpp
+++ b/axiom/connectors/tpch/TpchConnectorMetadata.cpp
@@ -24,22 +24,44 @@
 
 namespace facebook::axiom::connector::tpch {
 
-const folly::F14FastMap<std::string, double>& scaleFactorNamespaceMap() {
-  const static folly::F14FastMap<std::string, double> sfmap = {
-      {"tiny", 0.01},
-      {"sf1", 1},
-      {"sf100", 100},
-      {"sf1000", 1000},
-      {"sf10000", 10000},
-      {"sf100000", 100000},
-      {"sf300", 300},
-      {"sf3000", 3000},
-      {"sf30000", 30000}};
-  return sfmap;
+namespace {
+
+bool isValidTpchTableName(const std::string& tableName) {
+  for (auto tpchTable : velox::tpch::tables) {
+    if (tableName == velox::tpch::toTableName(tpchTable)) {
+      return true;
+    }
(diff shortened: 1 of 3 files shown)


axiom/connectors/tpch/TpchConnectorMetadata.cpp
```diff
@@ -16,6 +16,7 @@
 
 #include "axiom/connectors/tpch/TpchConnectorMetadata.h"
 
+// #include <algorithm>
```
mbasmanova, line 19: please, remove

axiom/connectors/tpch/TpchConnectorMetadata.cpp
```diff
@@ -24,22 +25,39 @@
...
+    return true;
+  }
+  if (schema.length() <= 2 || !schema.starts_with("sf")) {
+    return false;
+  }
+  bool nonzero = false;
+  for (char c : schema.substr(2)) {
+    if (!std::isdigit(c)) {
+      return false;
+    } else if (c != '0') {
```
mbasmanova, line 48: nit: drop 'else' after 'return'

axiom/connectors/tpch/TpchConnectorMetadata.cpp
```diff
@@ -24,22 +25,39 @@
...
-      {"sf1", 1},
-      {"sf100", 100},
-      {"sf1000", 1000},
-      {"sf10000", 10000},
-      {"sf100000", 100000},
-      {"sf300", 300},
-      {"sf3000", 3000},
-      {"sf30000", 30000}};
-  return sfmap;
+bool isValidTpchTableName(const std::string& tableName) {
```
mbasmanova, line 28: put these into anonymous namespace

mbasmanova (approved):
Thanks.

PR #402, opened 2025-09-16
refactor: Use folly F14 maps and sets instead of std::unordered

MBkkt:


axiom/optimizer/Schema.h
```diff
@@ -27,6 +27,8 @@
 /// though, so that a schema cache can have its own lifetime.
 namespace facebook::axiom::optimizer {
 
+// TODO: It seems like QGAllocator doesn't work for folly F14 containers.
+// Investigate and fix.
```
MBkkt, line 31: https://github.com/facebookexperimental/verax/issues/403

PR #405, opened 2025-09-16
refactor: Use string_view for lookups/etc instead of std::string

MBkkt:
First commit can be ignored, it's separate PR: https://github.com/facebookexperimental/verax/pull/402


axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1107,25 +1107,24 @@ TEST_F(PlanTest, values) {
...
-      plan.values({rowVector});
-    }
-    return plan.filter(filter).project({
-        fmt::format("n_nationkey AS {}1", alias),
-        fmt::format("n_regionkey AS {}2", alias),
-        fmt::format("n_comment AS {}3", alias),
-    });
-  };
+  auto makeLogicalPlan =
+      [&](uint8_t leafType, const std::string& filter, std::string_view alias) {
```
mbasmanova, line 1111: Why std::string for filter and view for alias?

MBkkt: filter used in plan matcher directly, I didn't change plan matcher ctors (I'm not sure what is better string_view or string by value).
alias used only for fmt::format here, so it can be string_view (and strings isn't created here).

I can change it if you want, it is tests, so I'm ok with any option here

MBkkt:
@mbasmanova rebased

mbasmanova:
Would you rebase?

PR #408, opened 2025-09-16
feat: Add logical plan node TableWriteNode

MBkkt:
Only last commit make sense to review.

Other commits here are parts of other PRs

### DIFF
--- a/axiom/logical_plan/LogicalPlanNode.h
+++ b/axiom/logical_plan/LogicalPlanNode.h
@@ -705,14 +705,7 @@ class TableWriteNode : public LogicalPlanNode {
       std::vector<std::string> columnNames,
       std::vector<ExprPtr> columnExpressions,
       velox::RowTypePtr outputType,
-      folly::F14FastMap<std::string, std::string> options = {})
-      : LogicalPlanNode{NodeKind::kTableWrite, std::move(id), {std::move(input)}, std::move(outputType)},
-        connectorId_{std::move(connectorId)},
-        tableName_{std::move(tableName)},
-        kind_{kind},
-        columnNames_{std::move(columnNames)},
-        columnExpressions_{std::move(columnExpressions)},
-        options_{std::move(options)} {}
+      folly::F14FastMap<std::string, std::string> options = {});
 
   const std::string& connectorId() const {
     return connectorId_;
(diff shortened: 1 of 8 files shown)


axiom/logical_plan/LogicalPlanNode.h
```diff
@@ -656,4 +657,102 @@ class UnnestNode : public LogicalPlanNode {
...
 using UnnestNodePtr = std::shared_ptr<const UnnestNode>;
 
+/// Values of enum corresponds to axiom::connector::WriteKind.
+/// Specifies what type of write is intended when initiating or concluding a
+/// write operation.
+enum class WriteKind {
+  // Rows are added and all columns must be specified for the TableWriter. This
+  // covers insert, create table and replacing a Hive partition and any other
+  // use that adds whole rows.
+  kInsert = 1,
```
mbasmanova, line 667: I noticed you removed =1, =2, =3 from the other enum; why have them here?

axiom/logical_plan/LogicalPlanNode.h
```diff
@@ -656,4 +657,102 @@ class UnnestNode : public LogicalPlanNode {
...
+      std::vector<ExprPtr> columnExpressions,
+      velox::RowTypePtr outputType,
+      folly::F14FastMap<std::string, std::string> options = {})
+      : LogicalPlanNode{NodeKind::kTableWrite, std::move(id), {std::move(input)}, std::move(outputType)},
+        connectorId_{std::move(connectorId)},
+        tableName_{std::move(tableName)},
+        kind_{kind},
+        columnNames_{std::move(columnNames)},
+        columnExpressions_{std::move(columnExpressions)},
+        options_{std::move(options)} {}
```
mbasmanova, line 716: Would you add some sanity checks? 

- columnNames.size() == columnExpressions.size()
- columnNames are unique and non-empty
- table name is not empty

axiom/logical_plan/LogicalPlanNode.h
MBkkt: > I noticed you removed =1, =2, =3 from the other enum

It's wrong, thanks to notice, I will fix it

MBkkt: The idea is that we want to have same numbers for connector WriteKind and logical_plan WriteKind.
And we probably want this numbers be written explicitly

MBkkt: > I noticed you removed =1, =2, =3 from the other enum; 

I don't sure, about what enums you're written, can you clarify it please?

mbasmanova: Maybe I got confused. I see both enums having integer values now.

PR #414, opened 2025-09-17
feat: ConnectorMetadata preparation for TableWrite

MBkkt:
## New description 

This PR is introduce PartitionType (the way how data shuffled), replacement for DistributionType mode.
It's also 
1. remove not really used and incorrect numPartitions from DistributionType.
2. make isBroadcast same as isGather property of DistributionType, that decide what type of shuffle will be done.
3. Implement HivePartitionType (copied from Orri write table PR and apply Maria's review suggestions)
4. Remove unused and unpractical interface writePartitionInfo (copied from Orri write table PR)
5. Extend finishWrite interface with bool (commit vs abort) (copied from Orri write table PR)
6. Make createTable/dropTable hive specific abstractions

The idea of this PR taken from Orri changes about TableWrite.
Plus Maria suggestions about hive buckets vs partitions.

## Old description 

@mbasmanova I'm not sure about hive part, can you look?

I'm planning to write tests for `dropTable`, if everything else is ok.

Other code is partially tested by current code and difficult to test fully without whole TableWrite implementation.
As an option, you can review this part, and then I will add to this PR next part.
If you have some idea about tests for this, I will glad to implement it.

WDYT?

### DIFF
--- a/axiom/optimizer/Schema.cpp
+++ b/axiom/optimizer/Schema.cpp
@@ -368,8 +368,9 @@ std::optional<bool> Distribution::canBeSamePartition(
     // Different locus cannot be copartitioned.
     return false;
   }
-  if (lhsType.isGather != rhsType.isGather) {
-    // If one is gather and the other is not, they cannot be copartitioned.
+  if (lhsType.isGather || rhsType.isGather) {
+    // If either is gather, they cannot be copartitioned,
+    // because we don't know where it will be.
     return false;
   }
   if (lhsType.isBroadcast || rhsType.isBroadcast) {
--- a/axiom/optimizer/Schema.h
+++ b/axiom/optimizer/Schema.h
@@ -118,19 +118,12 @@ using LocusCP = const Locus*;
 
 /// Distribution of data.
 /// 'partitionType' is nullptr if the data is not partitioned
-/// by some specific functions.
+/// by some connector specific function.
 struct DistributionType {
   LocusCP locus{nullptr};
   const connector::PartitionType* partitionType{nullptr};
   bool isGather{false};
   bool isBroadcast{false};
-
-  static DistributionType gather() {
-    static constexpr DistributionType kGather = {
-        .isGather = true,
-    };
-    return kGather;
-  }
 };
 
 // Describes output of relational operator. If base table, cardinality is
@@ -167,8 +160,11 @@ struct Distribution {
   static Distribution gather(
       ExprVector orderKeys = {},
       OrderTypeVector orderTypes = {}) {
+    static constexpr DistributionType kGather = {
+        .isGather = true,
+    };
     return {
-        DistributionType::gather(),
+        kGather,
         {},
         std::move(orderKeys),
         std::move(orderTypes),
--- a/axiom/optimizer/RelationOp.cpp
+++ b/axiom/optimizer/RelationOp.cpp
@@ -724,31 +724,13 @@ std::string Project::toString(bool recursive, bool detail) const {
   return out.str();
 }
 
-namespace {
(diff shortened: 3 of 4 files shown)


axiom/optimizer/Schema.h
```diff
@@ -116,24 +116,13 @@ class Locus {
...
-/// and both sides have an equal number of 1:1 type matched partitioning keys.
+/// Distribution of data.
+/// 'partitionType' is nullptr if the data is not partitioned
+/// by some specific functions.
 struct DistributionType {
-  bool operator==(const DistributionType& other) const = default;
-
   LocusCP locus{nullptr};
-  int32_t numPartitions{1};
+  const connector::PartitionType* partitionType{nullptr};
```
MBkkt, line 124:
```cpp
  // TODO: We probably should have separate partition types
  // for gather and Velox hash partitioning, instead of nullptr.
  // The main issue, for gather we don't know there it will be gathered.
  // For Velox hash partitioning, we know it will be partitioned
  // by hash of partitioning columns, but we don't know the number of
  // partitions before it's actually planned.
```

axiom/optimizer/Optimization.cpp
```diff
@@ -1368,8 +1351,6 @@ Distribution somePartition(const RelationOpPtrVector& inputs) {
   }
 
   DistributionType distributionType;
-  distributionType.numPartitions =
-      queryCtx()->optimization()->runnerOptions().numWorkers;
```
MBkkt, line 1372: This was useless, `numPartitions` actually didn't control number of partitions in PartitionedOutputNode
Instead this expression used `keys.empty() ? 1 : fragment.width` (actually correct one)

```
  source.fragment.planNode =
      std::make_shared<velox::core::PartitionedOutputNode>(
          nextId(),
          distribution.isBroadcast
              ? velox::core::PartitionedOutputNode::Kind::kBroadcast
              : velox::core::PartitionedOutputNode::Kind::kPartitioned,
          keys,
          keys.empty() ? 1 : fragment.width,
```

axiom/optimizer/Schema.cpp
```diff
@@ -351,22 +360,43 @@ ColumnCP IndexInfo::schemaColumn(ColumnCP keyValue) const {
...
+    // If either is gather, they cannot be copartitioned,
+    // because we don't know where it will be.
     return false;
   }
-  if (isBroadcast || other.isBroadcast) {
+  if (lhsType.isBroadcast || rhsType.isBroadcast) {
+    // If either is broadcast, they are copartitioned,
+    // because broadcast data is available everywhere.
     return true;
   }
```
MBkkt, line 380: I don't understand why, but gather and broadcast (or opposite) can't be same partition.

axiom/connectors/ConnectorMetadata.h
```diff
@@ -167,6 +182,42 @@ struct SortOrder {
...
+  /// Otherwise returns nullptr.
+  /// Partitions are compatible if data in one partitioned dataset can only
+  /// match data in the same partition of another dataset if joined on equality
+  /// of partition keys.
+  /// Compatibility is not strict equality in the case of e.g. Hive where a
+  /// dataset partitioned 8 ways is compatible with one partitioned 16 ways if
+  /// the function is the same. In such a case the partition that will be
+  /// returned is the 8 way one. On the 16 side data from partitions 0 and 1
+  /// match 0 on the 8 side and 2, 3 match 1 and so on.
+  virtual const PartitionType* copartition(
```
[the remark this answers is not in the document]
MBkkt: It should used for hive tables with partitions.
Maybe there's no such tables, but in such case partition is also probably not used.

PR #419, opened 2025-09-19
fix: TempProjections doesn't work for some cases

MBkkt:
This bug I believe was presented even for old implementation.

There're two separate issues:
1. We cannot do something like `exprs_.push_back(fieldRef);`, in general we cannot have field access in projection that just access this projection output.
2. numUsedChannels_ wasn't computed correctly for case with name, I find this when saw plan for TableWrite, but test covers this too.

### DIFF
--- a/axiom/optimizer/tests/PlanTest.cpp
+++ b/axiom/optimizer/tests/PlanTest.cpp
@@ -1373,30 +1373,6 @@ TEST_F(PlanTest, lastProjection) {
   ASSERT_TRUE(matcher->match(plan));
 }
 
-TEST_F(PlanTest, tmpProjectionsWorksForAliases) {
-  testConnector_->createTable(
-      "numbers", ROW({"a", "b", "c"}, {BIGINT(), BIGINT(), VARCHAR()}));
-
-  auto logicalPlan =
-      lp::PlanBuilder{}
-          .tableScan(kTestConnectorId, "numbers", {"a", "b", "c"})
-          .project({"a + b AS ab1", "a + b AS ab2"})
-          .aggregate({"ab1", "ab2"}, {"count(1) AS c1"})
-          .project({"ab1 AS x", "ab2 AS y", "c1 AS z"})
-          .build();
-
-  auto plan = toSingleNodePlan(logicalPlan);
-
-  auto matcher = core::PlanMatcherBuilder()
-                     .tableScan()
-                     .project({"1", "plus(a, b)"})
-                     .singleAggregation()
-                     .project({"ab1", "ab1", "c1"})
-                     .build();
-
-  ASSERT_TRUE(matcher->match(plan));
-}
-
 } // namespace
 } // namespace facebook::axiom::optimizer
 


axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1323,6 +1323,30 @@ TEST_F(PlanTest, lastProjection) {
   ASSERT_TRUE(matcher->match(plan));
 }
 
+TEST_F(PlanTest, tmpProjectionsWorksForAliases) {
+  testConnector_->createTable(
+      "numbers", ROW({"a", "b", "c"}, {BIGINT(), BIGINT(), VARCHAR()}));
+
+  auto logicalPlan =
```
mbasmanova, line 1330: I'm seeing that even simpler plan triggers a failure. Do we need 2nd project? What is it for?

```
  auto logicalPlan =
      lp::PlanBuilder{}
          .tableScan(kTestConnectorId, "numbers", {"a", "b", "c"})
          .project({"a + b AS ab1", "a + b AS ab2"})
          .aggregate({"ab1", "ab2"}, {"count(1) AS c1"})

-- Aggregation[2][SINGLE [ab1, ab2] c1 := count("__r2")] -> ab1:BIGINT, ab2:BIGINT, c1:BIGINT
  -- Project[1][expressions: (__r2:BIGINT, 1), (ab1:BIGINT, plus("a","b")), (ab2:BIGINT, "ab1")] -> __r2:BIGINT, ab1:BIGINT, ab2:BIGINT
    -- TableScan[0][numbers] -> a:BIGINT, b:BIGINT  

Reason: Field not found: ab1. Available fields are: a, b.
```

I'm also seeing suboptimal plan without aggregation. a + b is computed twice. I assume this is a separate problem not being addressed in the PR.

```
  auto logicalPlan =
      lp::PlanBuilder{}
          .tableScan(kTestConnectorId, "numbers", {"a", "b", "c"})
          .project({"a + b AS ab1", "a + b AS ab2"})

-- Project[1][expressions: (ab1:BIGINT, plus("a","b")), (ab2:BIGINT, plus("a","b"))] -> ab1:BIGINT, ab2:BIGINT
  -- TableScan[0][numbers] -> a:BIGINT, b:BIGINT
```

axiom/optimizer/ToVelox.cpp
```diff
@@ -530,51 +530,54 @@ class TempProjections {
...
       exprs_.push_back(queryCtx()->optimization()->toTypedExpr(expr));
-      names_.push_back(
-          optName ? *optName : fmt::format("__r{}", nextChannel_ - 1));
-      fieldRefs_.push_back(std::make_shared<velox::core::FieldAccessTypedExpr>(
-          toTypePtr(expr->value().type), names_.back()));
-      return fieldRefs_.back();
+      return addFieldRef(optName ? *optName : fmt::format("__r{}", nextField_));
+    }
+
+    auto& fieldRef = fieldRefs_[it->second];
```
mbasmanova, line 559: I find it difficult to follow this code. We discussed that all this logic doesn't belong to ToVelox and needs to be removed. Perhaps, we should do that instead of trying to patch up this logic. Alternatively, we may need to extract it into a separate file, add documentation and unit tests.

MBkkt: We need to move this code to physical plan, we have issue about this. 
And I think I will do it when I will finish table write story (axiom part + integration with axiom in serenedb).

I don't want to do this now, it's non trivial. I just want to fix bug, to avoid incorrect behavior.
I can write some comments in this method to make it more easy to follow.
WDYT?

axiom/optimizer/tests/PlanTest.cpp
MBkkt: I also created issue about aggregation group keys dedup https://github.com/facebookexperimental/verax/issues/431

PR #421, opened 2025-09-22
feat: Add support for window functions 

pashandor789:
This pull request adds support for window functions to the logical plan, introducing a new `WindowNode` that enables computation of windowed aggregates and functions.

**Window Function Support:**

Added a new `WindowNode` class which will be grouped by window specs (orderby + partitionby as gkeys) in QueryGraph.


MBkkt (commented):
We discussed and decided to remove logical_plan::WindowNode (and corresponding node in optimizer).
Instead project with window expressions will be used and ToVelox for window expressions in project nodes will be implemented

PR #433, opened 2025-09-25
feat: Do not create projections for constant args to aggregates

mbasmanova:
Summary:
Before: 

> SELECT k, count(1) FROM t GROUP BY 1

was planned as Scan -> Project(__r2: 1) -> Agg(k, count(__r2))

After:

There is no Project: Scan -> Agg(k, count(1)).

Differential Revision: D83271286

### DIFF
--- a/axiom/optimizer/QueryGraph.h
+++ b/axiom/optimizer/QueryGraph.h
@@ -886,8 +886,9 @@ class AggregationPlan : public PlanObject {
         aggregates_(std::move(aggregates)),
         columns_(std::move(columns)),
         intermediateColumns_(std::move(intermediateColumns)) {
-    VELOX_CHECK_EQ(groupingKeys.size() + aggregates.size(), columns.size());
-    VELOX_CHECK_EQ(aggregates.size(), intermediateColumns.size());
+    VELOX_CHECK(!groupingKeys_.empty() || !aggregates_.empty());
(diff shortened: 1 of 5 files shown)


axiom/optimizer/QueryGraph.h
```diff
@@ -885,7 +885,10 @@ class AggregationPlan : public PlanObject {
         groupingKeys_(std::move(groupingKeys)),
         aggregates_(std::move(aggregates)),
         columns_(std::move(columns)),
-        intermediateColumns_(std::move(intermediateColumns)) {}
+        intermediateColumns_(std::move(intermediateColumns)) {
+    VELOX_CHECK_EQ(groupingKeys.size() + aggregates.size(), columns.size());
```
xiaoxmeng, line 889: nit: do we want to check on the class member sizes directly? Thanks!

xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

axiom/optimizer/QueryGraph.h
MBkkt: I think it's not nit, right now it's kind of bug, it's always compare zero with zero.

mbasmanova: Nice catch. Let me fix this.

PR #434, opened 2025-09-26
fix: Deduplicate order by expressions

MBkkt:
fix https://github.com/facebookexperimental/verax/issues/423


axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -1373,6 +1373,26 @@ TEST_F(PlanTest, lastProjection) {
   ASSERT_TRUE(matcher->match(plan));
 }
 
+TEST_F(PlanTest, orderByDuplicateKeys) {
+  testConnector_->createTable("t", ROW({"a", "b"}, {BIGINT(), BIGINT()}));
+
+  auto logicalPlan = lp::PlanBuilder{}
+                         .tableScan(kTestConnectorId, "t", {"a", "b"})
```
mbasmanova, line 1380: nit: you can drop , {"a", "b"}

mbasmanova (approved):
Thanks.

PR #436, opened 2025-09-26
refactor: Optimize expression deduplication logic

MBkkt:


mbasmanova:
> refactor: Expression deduplication works suboptimal way and contains unnecessary code

Notice that this title describes the problem, not the changes or fixes contained in this PR. It sounds like this PR adds suboptimal logic and unnecessary code. Perhaps, consider 

> refactor: Streamline expression deduplication logic

or 

> refactor: Optimize expression deduplication logic

axiom/optimizer/ToGraph.cpp
```diff
@@ -951,58 +951,52 @@ struct AggregateDedupHasher {
...
-  ExprVector groupingKeys = translateColumns(agg.groupingKeys());
-
   ColumnVector columns;
 
   ExprVector deduppedGroupingKeys;
-  deduppedGroupingKeys.reserve(groupingKeys.size());
+  deduppedGroupingKeys.reserve(agg.groupingKeys().size());
 
   folly::F14FastMap<ExprCP, ColumnCP> uniqueGroupingKeys;
+  auto newRenames = renames_;
```
mbasmanova, line 976: This is a logic change, no?

axiom/optimizer/ToVelox.cpp
```diff
@@ -1338,6 +1338,14 @@ velox::core::PlanNodePtr ToVelox::makeAggregation(
   const auto numKeys = op.groupingKeys.size();
 
   TempProjections projections{*this, *op.input(), false};
+
+  std::vector<std::string> keyNames;
+  keyNames.reserve(op.groupingKeys.size());
+  for (auto i = 0; i < op.groupingKeys.size(); ++i) {
+    keyNames.push_back(outputName(op.columns()[i]));
+  }
+  auto keys = projections.toFieldRefs(op.groupingKeys, &keyNames);
```
mbasmanova, line 1347: Is this also a change in logic? In https://github.com/facebookexperimental/verax/pull/435#discussion_r2380700929 you wrote "It's also required because otherwise TempProjections fails to work"

axiom/optimizer/ToGraph.cpp
MBkkt: Yes, in this branch two commits

PR #441, opened 2025-09-27
fix: De-flake TpchPlanTest.q09

mbasmanova:
Summary:
Before this change, 1000 runs of q09 would reliably produce a few failures. After this change all 1000 runs succeed.

> $ buck test axiom/optimizer/... -- q09 --stress-runs 1000

TpchPlanTest used to compare query runs between Velox w/ Hive Connector and Axiom w/ TPC-H Connector. It seems that there is a bug in TPC-H Connector which needs further investigation. Using Hive Connector for both runs produces reliable results.

Differential Revision: D83408674


MBkkt:
@mbasmanova shouldn't we instead use this PR https://github.com/facebookincubator/velox/pull/14984 ?

Also can we add address, undefined, thread sanitizers in CI?
(ideally also memory, but it's build libc++ from source, we do this in serenedb, but it's quite specific).

xiaoxmeng (approved):
@mbasmanova thanks!

Issue #357, opened 2025-09-05
Projection and OrderBy don't interact well


MBkkt:
```cpp
TEST_F(PlanTest, orderByAfterProject) {
  const auto tableType =
      ROW({"a", "b", "c"}, {INTEGER(), INTEGER(), INTEGER()});
  testConnector_->createTable("t", tableType);

  auto logicalPlan = lp::PlanBuilder{}
                         .tableScan(kTestConnectorId, "t", {"a", "b", "c"})
                         .project({"c", "a + b as x1"})
                         .orderBy({"x1"})
                         .project({"x1", "c"})
                         .build();

  auto plan = toSingleNodePlan(logicalPlan);

  std::cerr << plan->toString(true, true) << std::endl;

  auto matcher = core::PlanMatcherBuilder{}
                     .tableScan("t", tableType)
                     .project({"c", "a + b as x1"})
                     .orderBy({"x1"})
                     .project({"x1", "c"})
                     .build();

  EXPECT_TRUE(matcher->match(plan));
}
```

Current output

```
-- Project[3][expressions: (x1:INTEGER, plus("a","b")), (c:INTEGER, "c")] -> x1:INTEGER, c:INTEGER
  -- OrderBy[2][__r3 ASC NULLS LAST] -> a:INTEGER, b:INTEGER, c:INTEGER, __r3:INTEGER
    -- Project[1][expressions: (a:INTEGER, "a"), (b:INTEGER, "b"), (c:INTEGER, "c"), (__r3:INTEGER, plus("a","b"))] -> a:INTEGER, b:INTEGER, c:INTEGER, __r3:INTEGER
      -- TableScan[0][t] -> a:INTEGER, b:INTEGER, c:INTEGER
```

Last projection shouldn't compute anything in such example

> (x1:INTEGER, plus("a","b")) 

this doesn't look good

MBkkt:
@mbasmanova @oerling I think it's about any projection vs other node that can compute expressions.

In general it feels like current to graph, just do projection this undesired way.

From one side it's good, because it allows us to use expressions before actual projection.
But from other, if projection (after some reorder/etc) happens before node with expression, projection columns isn't used.
And if projection happens after node with expression, it's also don't reuse it.

PR #451, opened 2025-10-01
feat: Eliminate duplicate projections

mbasmanova:
Summary: Fix  https://github.com/facebookincubator/axiom/issues/357

Extend PlanState to store pre-computed expressions as a map from expression to a column produced by PrecomputeProjection.

Populate this map when processing OrderBy and use it to eliminate expressions when generating "final" Project for a DT.

Follow-ups need to apply similar logic to all other nodes that use PrecomputeProjection.

Differential Revision: D83653189


xiaoxmeng (approved):
@mbasmanova thanks!

PR #453, opened 2025-10-01
build: Do not use velox_add_library in cmake

mbasmanova:
Differential Revision: D83676849


MBkkt (commented):
This PR introduce multiple targets for users that used VELOX_MONO, that doesn't good in case of static libraries.
Can we consider different approach please.
Different approach allows both single target and multiple targets
This PR requires as manually patch whole axiom cmake

axiom/connectors/tests/CMakeLists.txt
```diff
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-velox_add_library(axiom_test_connector TestConnector.cpp)
+add_library(axiom_test_connector TestConnector.cpp)
```
MBkkt, line 15: Why does `axiom_hive_connector_metadata OBJECT` but `add_library(axiom_test_connector`?

mbasmanova:
@MBkkt This PR doesn't introduce any new targets. All it does is replace velox_add_library with add_library and velox_link_library with link_library to fix build error: https://github.com/facebookincubator/axiom/actions/runs/18162165352/job/51695246328?pr=451

MBkkt:
@mbasmanova This PR replace 1 targets with N+1 targets for users that uses VELOX_MONO.
Because before this PR all changed targets was part of velox library for us. Now it's multiple different targets.

axiom/connectors/tpch/CMakeLists.txt
```diff
@@ -12,10 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-velox_add_library(axiom_tpch_connector_metadata OBJECT
-                  TpchConnectorMetadata.cpp)
+add_library(axiom_tpch_connector_metadata OBJECT TpchConnectorMetadata.cpp)
```
kgpai, line 15: Curious why this is object for tpchconnectormetadata .

kgpai (approved):

mbasmanova:
> Because before this PR all changed targets was part of velox library (with VELOX_MONO).

This doesn't seem right. Velox is a dependency of Axiom. Having "velox" library include Axiom is ... strange.

I think we need to figure out a cleaner solution, but for now the priority is to get CI working.

MBkkt:
@mbasmanova 
> This doesn't seem right.

I agree, this is why in different PR, that fixes build, I introduced new variable `AXIOM_MONO_LIBRARY`, so it's separated from velox library, but still can be single library

axiom/connectors/tpch/CMakeLists.txt
MBkkt: It's also `axiom_hive_connector_metadata`

MBkkt:
@kgpai 

>  - But I am afraid I dont understand why that is a problem.

I think, I explained why is it important for us.
Can you explain why is it important for you/fb/? to write `add_library` instead of `axiom_add_library` and `target_add_libraries` instead of `axiom_add_libraries`?

mbasmanova:
> This PR replaces 1 library with N+1 libraries for users that uses VELOX_MONO.
> Because before this PR all changed targets was part of velox library (with VELOX_MONO).
> Now it's multiple different targets.

Do you know what is the value of N ? Can I get that to count instances of add_library under axiom? While I believe you'd like N to be zero, is there a non-zero value that would also work? Can we have 1 library per top-level directory under axiom? That would be N = 6? 

> common  connectors  logical_plan  optimizer  runner  sql

mbasmanova:
> This PR requires us in SereneDB manually patch whole Axiom cmake

Would you share that patch? I'd like to better understand what is missing.

MBkkt:
@mbasmanova 

> Do you know what is the value of N ? 

<= 10 I believe for now

> Can I get that by counting instances of add_library under axiom? 

Yes, but ignore test only code.

> While I believe you'd like N to be zero, is there a non-zero value that would also work? 

N == 1, sounds also completely ok
N <= 10 potentially are also ok, but I don't understand how this will be controlled 

> Can we have 1 library per top-level directory under axiom?

I think this works, but it's already not true, because it's difficult to control

>  That would be N = 6?

I think right now, it's not big number yes, but I think it can change in future.
Change such things in future is complicated.
For an example, we already have not really needed separation:
Do someone really need separated library for axiom_model or axiom_schema_resolver?

> Would you share that patch? I'd like to better understand what is missing.

You probably saw my PR with AxiomUtils, do I need to rebase it?

mbasmanova:
>  I don't understand how this will be controlled

Assuming 1 library per top-level directory (roughly) works, we can write guidelines for how to structure cmake files. We'll have to rely on code review and occasional refactoring to enforce these. Maybe there is a way to write a pre-commit hook that fails if total number of libraries exceeds 12 (or some other magic number).

> Do someone really need separated library for axiom_model or axiom_schema_resolver?

axiom_model might be better separate because it is pure experimental code which is expected to be used in a standalone utility, not during runtime. I merged it into optimizer library, but I think that's a mistake.

axiom_schema_resolver, as you pointed out, can be merged with the connector library. Let's do that.

> You probably saw my PR with AxiomUtils

I didn't realize that's the actual patch you are maintaining. Got it.

MBkkt:
> I didn't realize that's the actual patch you are maintaining. Got it.

I maintained different smaller patch (that just adjusted few lines in VeloxUtils), but now when `velox_add_library` not used I think we need to use whole AxiomUtils PR

PR #455, opened 2025-10-01
build: Merge 3 libraries under optimizer into one

mbasmanova:
Summary: Also, sort .cpp files under add_library alphabetically.

Differential Revision: D83694821


mbasmanova:
CC: @assignUser

mbasmanova:
> About PR in general, I still don't understand how to control this in future, and why this works better than mono approach for you

Would you elaborate what you refer to by "how to control this"? What is "this"?

MBkkt:
@mbasmanova How to control that we don't have more libraries when directories in root?

MBkkt:
@mbasmanova I still don't have understanding why axiom mono considering as something bad.
But in any case what you suggest is make sense for me, thanks

mbasmanova:
> I still don't have understanding why axiom mono considering as something bad.

That's a good question. I don't understand this either.

In my mind, it would be helpful to create many small libraries to allow for faster incremental builds and also have an option to create one mono binary for deploy to prod.

@kgpai @assignUser Would this make sense? Is this possible with cmake?

mbasmanova:
@assignUser Jacob, thank you for clarifying. This makes sense to me.

> So 'header hygiene' (e.g. forward decl where possible) and proper separation of concerns will help a lot.

This is a great point. Wondering if we could add a section about this to Velox coding guidelines.

mbasmanova:
@MBkkt 

> Some headers can be avoided with forward declaration, but it doesn't worth it

Would you elaborate what you mean by "it doesn't worth it"?

> google forbid them entirely

Do you have a link to this guidance from Google?

Curious if you know about some C++ projects that do this "right"? Perhaps, we can take a look at what / how they do and see if it makes sense to copy that approach.

MBkkt:
> Would you elaborate what you mean by "it doesn't worth it"?

I think I agree with google codestyle here, link bellow

> Do you have a link to this guidance from Google?

https://google.github.io/styleguide/cppguide.html#Forward_Declarations

> Try to avoid forward declarations of entities defined in another project.

> Curious if you know about some C++ projects that do this "right"? Perhaps, we can take a look at what / how they do and see if it makes sense to copy that approach.

I think there's no best way to do this, every project has own consideration.

But for an example we in serenedb, use "fwd.hpp" inside each directory where we decided to use forward declaration.
This allow us to have smaller compile time and at the same time avoid 
1. situation when some class forward declared 100 times (not convenient to change/refactor/etc).
2. creating additional header with forward declarations for each common header

PR #456, opened 2025-10-01
build: Use gersemi instead of the no longer maintained cmake-format

mbasmanova:
Summary: See https://github.com/facebookincubator/velox/pull/14237/files

Differential Revision: D83697286


mbasmanova:
CC: @assignUser

PR #460, opened 2025-10-02
build: Enable VELOX_MONO_LIBRARY

MBkkt:
* Don't use `OBJECT` unless required
* Don't use `velox_add_library` for targets that don't start with `velox_` prefix
* Remove unnecessary `add_dependencies`
* Remove `QueryTestBase.cpp` duplication (it's referenced in two libraries, and one library linked to another)
* Use `VELOX_MONO_LIBRARY` in Axiom CI to make build faster and ensure it works

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ BUILD_BASE_DIR=_build
 BUILD_DIR=release
 BUILD_TYPE=Release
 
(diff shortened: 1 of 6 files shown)


mbasmanova (approved):
Thanks.

Issue #462, opened 2025-10-02
Adjust Makefile format-fix/etc to use pre-commit instead of cmake/clang formats directly


MBkkt:

mbasmanova:
@MBkkt I'm not sure I understand. Would you elaborate?

mbasmanova:
Fixed in #466

PR #466, opened 2025-10-02
refactor: Delete no longer used scripts/format-check.py

mbasmanova:
Summary: We switched to using pre-commit hooks.

Fixes https://github.com/facebookincubator/axiom/issues/462

Differential Revision: D83751750


MBkkt:
> We switched to using pre-commit hooks.

In first I was a little confused that pre-commit isn't about git hooks but about some modern python package (script)

kgpai (approved):

PR #468, opened 2025-10-02
build: Add axiom_mono cmake target

mbasmanova:
Differential Revision: D83780645


mbasmanova:
CC: @assignUser

@MBkkt Valery, would something like this be useful?

kgpai (approved):

Issue #474, opened 2025-10-06
Incorrect dedup of complex type constants


mbasmanova:
ToGraph::makeConstant dedups constants using Variant. This is incorrect because empty arrays of different types use the same Variant. The dedup needs to consider both Variant and Type.

Repro:

```
TEST_F(PlanTest, dedupEmptyArrays) {
  testConnector_->addTable(
      "t", ROW({"a", "b"}, {ARRAY(REAL()), MAP(INTEGER(), REAL())}));

  auto logicalPlan =
      lp::PlanBuilder()
          .values(ROW({}), {variant::row({})})
          .map({
              lp::Lit(variant::array({}), ARRAY(ARRAY(REAL()))).as("a"),
              lp::Lit(variant::array({}), ARRAY(MAP(INTEGER(), VARCHAR())))
                  .as("b"),
          })
          .map({
              "typeof(a)",
              "typeof(b)",
          })
          .build();

  auto referencePlan =
      exec::test::PlanBuilder()
          .values({makeRowVector(ROW({}), 1)})
          .project({"'array(array(real))'", "'array(map(integer, varchar))'"})
          .planNode();

  checkSame(logicalPlan, referencePlan, {.numDrivers = 1, .numWorkers = 1});
}

Expected 1, got 1
1 extra rows, 1 missing rows
1 of extra rows:
"array(array(real))" | "array(array(real))"

1 of missing rows:
"array(array(real))" | "array(map(integer, varchar))"
```

CC: @oerling

PR #476, opened 2025-10-07
fix: Dedup of complex type constants

mbasmanova:
Summary:
Make sure to include both type and Variant value in dedup key.

Remove unused APIs:
- QueryGraphContext::toVector and toVectorPtr.
- Literal(BaseVector) ctor.

Fixes https://github.com/facebookincubator/axiom/issues/474

Differential Revision: D84024098


xiaoxmeng (approved):

axiom/optimizer/ToGraph.h
```diff
@@ -61,17 +61,20 @@ struct ExprDedupHasher {
 using FunctionDedupMap =
     folly::F14FastMap<ExprDedupKey, ExprCP, ExprDedupHasher>;
 
-struct VariantPtrHasher {
-  size_t operator()(const std::shared_ptr<const velox::Variant>& value) const {
-    return value->hash();
+struct TypedVariant {
+  const velox::Type* type;
```
MBkkt, line 65: I think this isn't needed, but if you use it, you can 
1. use `pointer == pointer` in `TypedVariantComparer`
2. use pointer hash instead of `hashKind` in `TypedVariantHasher`

axiom/optimizer/ToGraph.h
```diff
@@ -61,17 +61,20 @@ struct ExprDedupHasher {
...
 };
 
-struct VariantPtrComparer {
-  bool operator()(
-      const std::shared_ptr<const velox::Variant>& left,
-      const std::shared_ptr<const velox::Variant>& right) const {
-    return *left == *right;
+struct TypedVariantComparer {
+  bool operator()(const TypedVariant& left, const TypedVariant& right) const {
+    return *left.type == *right.type && *left.value == *right.value;
```
MBkkt, line 77:
```suggestion
    return left.type == right.type && *left.value == *right.value;
```
They're deduplicated

mbasmanova: Good points. Updated. Thanks.

PR #479, opened 2025-10-07
feat: Add support for TableWrite

MBkkt:
Extend the Optimizer to support logical plans with TableWriteNode as a root. Allows to create plans that append data to existing tables.

The input to TableWriteNode may produce only a subset of the columns in a table. The rest of the columns will be populated with default values. The connector is responsible for specifying default values. By default, these are assumed to be NULL.

ToGraph creates a separate DT for TableWrite. This allows to simply the rest of the code as DT with TableWrite doesn't contain anything else.

The optimizer doesn't compute the cost of TableWrite. Since every candidate plan must include TableWrite in the end, the cost of write doesn't affect cost comparison for the candidates.

Add `PlanAndWrite` instead of `PlanAndStats` to be able pass `finishWrite` callback that should invokes `ConnectorMetadata::finishWrite` or `abortWrite`.

Change `ConnectorMetadata::finishWrite` to return number of rows.
And actually compute number of rows for LocalHiveConnectorMetadata::finishWrite.

Change `ConnectorMetadata::abortWrite` to make it noexcept.
And actually make `LocalHiveConnectorMetadata::abortWrite` noexcept.

Extend LocalRunner to run finishWrite callback and construct vector with number of written rows.

Added `shared_from_this` to `axiom::connector::Table` because it's tricky to pass `shared_ptr<Table>` from `ToGraph` to `ToVelox`. TODO: Consider other way of doing this without `shared_from_this`, but also without additional hashmap/vector/etc.

Follow-ups:
1. Add tests for default column values with non-null default value.
2. Implement WriteKind::kCreate, kUpdate, kDelete.
4. Add support for writing partitioned tables.
5. Update LocalHiveConnectorMetadata to use Velox FileSystem instead of Linux-specific APIs.

### DIFF
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -1512,7 +1512,6 @@ PlanObjectP ToGraph::addWrite(const lp::TableWriteNode& tableWrite) {
   columnExprs.reserve(tableSchema.size());
   for (uint32_t i = 0; i < tableSchema.size(); ++i) {
     const auto& columnName = tableSchema.nameOf(i);
-    const auto& columnType = tableSchema.childAt(i);
 
     auto it = std::ranges::find(tableWrite.columnNames(), columnName);
     if (it != tableWrite.columnNames().end()) {
@@ -1525,7 +1524,7 @@ PlanObjectP ToGraph::addWrite(const lp::TableWriteNode& tableWrite) {
       columnExprs.push_back(make<Literal>(
           Value{toType(tableColumn->type()), 1}, &tableColumn->defaultValue()));
     }
-    VELOX_DCHECK(*columnType == *columnExprs.back()->value().type);
+    VELOX_DCHECK(*tableSchema.childAt(i) == *columnExprs.back()->value().type);
   }
 
   renames_.clear();


axiom/optimizer/DerivedTable.h
```diff
@@ -33,6 +33,9 @@ using AggregationPlanCP = const AggregationPlan*;
 enum class OrderType;
 using OrderTypeVector = QGVector<OrderType>;
 
+class WritePlan;
+using WritePlanCP = const WritePlan*;
```
mbasmanova, line 37: I wonder if we have checks in place to ensure that Write node appears only once in logical plan it is located at the root.

axiom/optimizer/Optimization.cpp
```diff
@@ -675,6 +675,18 @@ void Optimization::addPostprocess(
     DerivedTableCP dt,
     RelationOpPtr& plan,
     PlanState& state) const {
+  if (dt->write) {
```
mbasmanova, line 678: Would you update PR description to explain how we model writes in the query graph, how we process these during optimization and how we convert these to Velox? It would be nice to clarify why the checks below are valid, e.g. why a DT with a write cannot have an agg, order by or limit.

axiom/optimizer/Plan.cpp
```diff
@@ -214,6 +214,11 @@ const PlanObjectSet& PlanState::downstreamColumns() const {
     }
   }
 
+  // Write.
+  if (dt->write && !placed.contains(dt->write)) {
```
mbasmanova, line 218: > placed.contains(dt->write)

Wouldn't this always be false? Is there a use case for calling this API after placing a write?

axiom/optimizer/QueryGraph.h
```diff
@@ -932,4 +932,42 @@ class AggregationPlan : public PlanObject {
 
 using AggregationPlanCP = const AggregationPlan*;
 
+class WritePlan : public PlanObject {
+ public:
+  WritePlan(
+      const connector::Table& table,
+      connector::WriteKind kind,
+      ExprVector columnExpressions,
```
mbasmanova, line 940: columnExpressions -> columnExprs

we have a guideline that asks to not abbreviate, but "expression" is a long word that is used a lot; hence, we use "expr"; would be nice to be consistent

axiom/optimizer/QueryGraph.h
```diff
@@ -932,4 +932,42 @@ class AggregationPlan : public PlanObject {
 
 using AggregationPlanCP = const AggregationPlan*;
 
+class WritePlan : public PlanObject {
+ public:
+  WritePlan(
```
mbasmanova, line 937: Would you document the arguments? Do we expect columnExprs and output be aligned 1:1 and be non-empty? Let's add sanity checks.

MBkkt: No, output is something like (count of written rows, etc)
expressions is expressions used for write into the table

axiom/connectors/hive/LocalHiveConnectorMetadata.cpp
```diff
@@ -783,14 +791,56 @@ void deleteDirectoryRecursive(const std::string& path) {
...
+      tempDirPath = parentDir / ("temp_" + std::to_string(randomNumber));
+    } while (fs::exists(tempDirPath));
+    if (velox::common::generateFileDirectory(tempDirPath.c_str())) {
+      return tempDirPath;
+    }
+  }
+}
+
+// Move all files from sourceDir to targetDir recursively.
+void moveFilesRecursively(
```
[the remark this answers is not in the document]
MBkkt: I think it should be better to use it, but I think I prefer to don't change most of hive code and keep it linux only as it was before my PR.

I can create issue about this.

axiom/optimizer/DerivedTable.h
MBkkt: No we don't have such checks because I think it's not required by our current code.

In general right now nothing prevent us from logical plan where insert isn't root.
I think it can make sense, because it allows as to put some compute after table write on table write dependent column output.

axiom/connectors/hive/LocalHiveConnectorMetadata.cpp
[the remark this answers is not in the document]
mbasmanova: Sounds good. Thanks.

axiom/optimizer/DerivedTable.h
mbasmanova: Given that we are not supporting such plans, it would be helpful to add checks to prevent these plans from being created in the first place. We can relax these check if we ever decide to support shapes where write is not a root node.

axiom/optimizer/QueryGraph.h
MBkkt: Ok, changed

axiom/optimizer/Plan.cpp
MBkkt: Changed to assert

axiom/connectors/hive/HiveConnectorMetadata.h
```diff
@@ -202,6 +202,10 @@ class HiveConnectorMetadata : public ConnectorMetadata {
   /// Return the filesystem path for the storage of the specified table.
   virtual std::string tablePath(std::string_view table) const = 0;
 
+  virtual std::string makeStagingDirectory() {
```
mbasmanova, line 205: Would you document this API?

Should it be pure virtual?

axiom/connectors/hive/LocalHiveConnectorMetadata.cpp
```diff
@@ -799,6 +799,43 @@ void deleteDirectoryContents(const std::string& path) {
...
+      templatePath);
+  return resultPath;
+}
+
+// Move all files from sourceDir to targetDir recursively.
+void moveFilesRecursively(
+    const fs::path& sourceDir,
+    const fs::path& targetDir) {
+  if (!fs::is_directory(sourceDir)) {
+    throw std::runtime_error(
```
mbasmanova, line 819: VELOX_FAIL for consistency

axiom/connectors/hive/LocalHiveConnectorMetadata.cpp
```diff
@@ -799,6 +799,43 @@ void deleteDirectoryContents(const std::string& path) {
...
+        "Source directory does not exist or is not a directory: " +
+        sourceDir.string());
+  }
+  // Create the target directory if it doesn't exist
+  fs::create_directories(targetDir);
```
mbasmanova, line 824: Do we need to check if this was successful? What happens on failure?

axiom/optimizer/Optimization.cpp
```diff
@@ -675,6 +675,19 @@ void Optimization::addPostprocess(
...
+    VELOX_DCHECK(!dt->hasAggregation());
+    VELOX_DCHECK(!dt->hasOrderBy());
+    VELOX_DCHECK(!dt->hasLimit());
+    VELOX_DCHECK(dt->columns == dt->write->output());
+    PrecomputeProjection precompute{plan, dt, /*projectAllInputs=*/false};
+    auto writeColumns = precompute.toColumns(dt->write->columnExprs());
+    plan = std::move(precompute).maybeProject();
+    state.addCost(*plan);
+    plan = make<TableWrite>(plan, std::move(writeColumns), dt->write);
```
mbasmanova, line 687: Should we add the cost of TableWrite as well?

axiom/optimizer/QueryGraph.h
```diff
@@ -932,4 +932,46 @@ class AggregationPlan : public PlanObject {
 
 using AggregationPlanCP = const AggregationPlan*;
 
+class WritePlan : public PlanObject {
+ public:
+  /// @param table The table to write to.
+  /// @param kind Indicates the type of write (create/insert/delete/update)
+  /// @param columnExpressions Expressions producing the values to write.
```
mbasmanova, line 939: columnExpressions -> columnExprs

Do we expect these to match table schema? E.g. same number of columns and same order? Would be nice to clarify.

axiom/optimizer/ToGraph.cpp
```diff
@@ -1487,6 +1482,70 @@ PlanObjectP ToGraph::addLimit(const lp::LimitNode& limitNode) {
   return currentDt_;
 }
 
+PlanObjectP ToGraph::addWrite(const lp::TableWriteNode& tableWrite) {
+  if (tableWrite.writeKind() != lp::WriteKind::kInsert) {
+    VELOX_NYI("Only INSERT supported for TableWrite");
```
mbasmanova, line 1487: Why did you add kCreate if only kInsert is supported? I'm confused.

axiom/optimizer/ToGraph.cpp
```diff
@@ -1487,6 +1482,70 @@ PlanObjectP ToGraph::addLimit(const lp::LimitNode& limitNode) {
...
+    if (it != tableWrite.columnNames().end()) {
+      const auto nth = it - tableWrite.columnNames().begin();
+      const auto& columnExpr = tableWrite.columnExpressions()[nth];
+      columnExprs.push_back(translateExpr(columnExpr));
+    } else {
+      const auto* tableColumn = connectorTable->findColumn(columnName);
+      VELOX_DCHECK_NOT_NULL(tableColumn);
+      auto& defaultValue = tableColumn->defaultValue();
+      columnExprs.push_back(make<Literal>(
+          Value{toType(defaultValue.inferType()), 1}, &defaultValue));
```
mbasmanova, line 1518: > defaultValue.inferType()

This is not right. We should not use Variant::inferType() in prod code.

axiom/optimizer/ToVelox.cpp
```diff
@@ -228,12 +228,24 @@ PlanAndStats ToVelox::toVeloxPlan(
   top.fragment.planNode = makeFragment(plan, top, stages);
   stages.push_back(std::move(top));
 
+  runner::FinishWrite finishWrite;
+  if (!finishWrites_.empty()) {
```
mbasmanova, line 232: Why finishWrites_ is a list? Is this a left-over from multi-table-layout prototype? If so, let's remove.

axiom/optimizer/QueryGraph.h
MBkkt: > Do we expect these to match table schema? 

Yes
> E.g. same number of columns and same order? 
We expect that number of columns will be same.

We're also know that order will be same, but it doesn't matter, because
1. TableWrite allow different order of input
2. This is used only to get columns from `PrecomputeProjection`, and if they're already presented in the input, they won't be reordered.

> Would be nice to clarify.

I will add assert about count

axiom/runner/LocalRunner.cpp
```diff
@@ -143,12 +143,18 @@ LocalRunner::LocalRunner(
...
+      finishWrite_{plan_->finishWrite()},
       splitSourceFactory_{std::move(splitSourceFactory)} {
   params_.queryCtx = std::move(queryCtx);
   params_.outputPool = std::move(outputPool);
 }
 
 velox::RowVectorPtr LocalRunner::next() {
+  if (finishWrite_) {
+    runWrite();
+    return nullptr;
```
mbasmanova, line 155: Would you update documentation for next() API? Typically, it would be nice to at least return the number of rows written.

axiom/optimizer/ToGraph.cpp
MBkkt: > Why did you add kCreate if only kInsert is supported? I'm confused.

You mean in previous PR, because I want to use static_cast from `lp::WriteKind` to `connector::WriteKind`, and have clear errors about kCreate isn't supported

axiom/optimizer/ToGraph.cpp
```diff
@@ -1487,6 +1482,70 @@ PlanObjectP ToGraph::addLimit(const lp::LimitNode& limitNode) {
...
+
+    auto it = std::ranges::find(tableWrite.columnNames(), columnName);
+    if (it != tableWrite.columnNames().end()) {
+      const auto nth = it - tableWrite.columnNames().begin();
+      const auto& columnExpr = tableWrite.columnExpressions()[nth];
+      columnExprs.push_back(translateExpr(columnExpr));
+    } else {
+      const auto* tableColumn = connectorTable->findColumn(columnName);
+      VELOX_DCHECK_NOT_NULL(tableColumn);
+      auto& defaultValue = tableColumn->defaultValue();
```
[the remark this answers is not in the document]
MBkkt: What do you mean?
`decltype(defaultValue)` is `Variant` not `some_ptr<Variant>`.

If it's sql null, it's completely ok, because it's default default column value.

axiom/optimizer/tests/WritePartitionTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+    test::HiveQueriesTestBase::TearDownTestCase();
+  }
+
+  void SetUp() override {
+    HiveQueriesTestBase::SetUp();
+    connector_ = velox::connector::getConnector(exec::test::kHiveConnectorId);
+    metadata_ = dynamic_cast<connector::hive::LocalHiveConnectorMetadata*>(
+        connector::ConnectorMetadata::metadata(exec::test::kHiveConnectorId));
+    parquet::registerParquetReaderFactory();
+    parquet::registerParquetWriterFactory();
```
mbasmanova, line 47: Do we need these?

At least, registerParquetReaderFactory is already called in one of the base classes.

axiom/optimizer/tests/WritePartitionTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  void TearDown() override {
+    connector_.reset();
+    HiveQueriesTestBase::TearDown();
+    parquet::unregisterParquetReaderFactory();
+    parquet::unregisterParquetWriterFactory();
```
mbasmanova, line 54: Let's ensure the order is the opposite of the SetUp.

axiom/optimizer/tests/WritePartitionTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+  auto write1 = lp::PlanBuilder(context)
+                    .values({data})
+                    .tableWrite(
+                        exec::test::kHiveConnectorId,
+                        "test",
+                        lp::WriteKind::kInsert,
+                        {"key1", "key2", "data", "ds"},
+                        exprs({"key1", "key2", "data", "ds"}),
+                        writeOutputType_)
```
mbasmanova, line 135: I think the write output type needs to come from the connector. Otherwise, how would we produce this plan from SQL? Are we missing some Connector APIs?

axiom/optimizer/tests/WritePartitionTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+                    .build();
+  runVelox(write1);
+
+  auto countPlan =
+      lp::PlanBuilder(context)
+          .tableScan(exec::test::kHiveConnectorId, "test", {"key1"})
```
mbasmanova, line 141: {"key1"} -> {} or just drop this argument altogether; we don't need to read any columns to count rows

axiom/optimizer/tests/WritePartitionTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+          .aggregate({}, {"count(1)"})
+          .build();
+
+  {
+    auto result = runVelox(countPlan);
+    EXPECT_EQ(
+        kTestBatchSize * 10,
+        result.results[0]->childAt(0)->as<FlatVector<int64_t>>()->valueAt(0));
+  }
```
mbasmanova, line 150: This code repeats. Consider adding a helper function.

axiom/optimizer/tests/WritePartitionTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+  // the copy runs distributed.
+  folly::F14FastMap<std::string, std::string> options2 = {
+      {"file_format", "parquet"},
+      {"compression_kind", "snappy"},
+  };
+  {
+    auto table = metadata_->createTable("test2", tableType, options, {});
+    auto handle =
+        metadata_->beginWrite(table, connector::WriteKind::kCreate, {});
+    metadata_->finishWrite(handle, {}, {}).get();
```
mbasmanova, line 196: this code repeats as well

axiom/optimizer/tests/WritePartitionTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+  void TearDown() override {
+    connector_.reset();
+    HiveQueriesTestBase::TearDown();
+    parquet::unregisterParquetReaderFactory();
+    parquet::unregisterParquetWriterFactory();
+  }
+
+  std::vector<RowVectorPtr>
+  makeTestData(int32_t numBatches, int32_t batchSize, int32_t dayOffset = 0) {
```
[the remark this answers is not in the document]
MBkkt: > Is this a copy-paste? 

Yes, from Orri TableWrite PR, I will try to cleanup this and other parts of test

axiom/optimizer/tests/WritePartitionTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+                      .filter("data2 is null")
+                      .project({"key1", "key2", "data", "ds"})
+                      .build();
+
+  {
+    auto result = runVelox(readPlan);
+    exec::test::assertEqualResults(data, result.results);
+  }
+
+  // Create a second table to copy the first one into. Values runs single node,
```
mbasmanova, line 186: > Values runs single node, the copy runs distributed.

What does this refer to? I can't connect this comment to code.

axiom/optimizer/tests/WritePartitionTest.cpp
MBkkt: > Otherwise, how would we produce this plan from SQL?

I think plan builder/something like this should can have this logic.

If you think some connector API is needed for fb usecases, please say how this API should look and I will add it

axiom/connectors/hive/HiveConnectorMetadata.h
MBkkt: Added docs, made pure virtual

axiom/connectors/hive/LocalHiveConnectorMetadata.cpp
MBkkt: Done

axiom/optimizer/ToGraph.cpp
MBkkt: I think I wrote that I want it same as connector.
I will try be more clear next time

axiom/optimizer/ToGraph.cpp
```diff
@@ -1487,6 +1482,70 @@ PlanObjectP ToGraph::addLimit(const lp::LimitNode& limitNode) {
...
+  const auto* schemaTable =
+      schema_.findTable(tableWrite.connectorId(), tableWrite.tableName());
+  VELOX_CHECK_NOT_NULL(
+      schemaTable,
+      "Table not found: {} via connector {}",
+      tableWrite.tableName(),
+      tableWrite.connectorId());
+  const auto* connectorTable = schemaTable->connectorTable;
+  VELOX_DCHECK_NOT_NULL(connectorTable);
+  const auto& tableRow = *connectorTable->type();
```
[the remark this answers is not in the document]
MBkkt: Done, I choose tableSchema

axiom/optimizer/tests/WritePartitionTest.cpp
mbasmanova: I don't think this is specific to FB-internal use cases, but I think we would want the Connector to specify the output type of the write in ConnectorWriteHandle. 

Something like this might work:

```
class ConnectorWriteHandle {
 public:
  explicit ConnectorWriteHandle(
      velox::connector::ConnectorInsertTableHandlePtr veloxHandle)
      : veloxHandle_(std::move(veloxHandle)) {}

  virtual ~ConnectorWriteHandle() = default;

  const velox::connector::ConnectorInsertTableHandlePtr& veloxHandle() const {
    return veloxHandle_;
  }

// TODO Add an API to return output type for the write. That type should match the writeResult passed to finishWrite.

  velox::RowTypePtr writeResultType() const;

 private:
  const velox::connector::ConnectorInsertTableHandlePtr veloxHandle_;
};
```

axiom/optimizer/ToVelox.cpp
```diff
@@ -1412,6 +1424,67 @@ velox::core::PlanNodePtr ToVelox::makeValues(
...
+          metadata->abortWrite(handle, {});
+        }
+        handle.reset();
+      });
+
+  return std::make_shared<velox::core::TableWriteNode>(
+      nextId(),
+      ROW(std::move(inputNames), std::move(inputTypes)),
+      std::move(columnNames),
+      std::nullopt,
```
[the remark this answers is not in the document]
MBkkt: Fixed

axiom/runner/LocalRunner.cpp
```diff
@@ -161,6 +167,28 @@ velox::RowVectorPtr LocalRunner::next() {
...
 
+void LocalRunner::runWrite() {
+  std::vector<velox::RowVectorPtr> result;
+  try {
+    start();
+    while (cursor_->moveNext()) {
+      result.push_back(cursor_->current());
+    }
+    finishWrite_(true, result);
+    state_ = State::kFinished;
```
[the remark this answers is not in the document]
MBkkt: No, fixed

axiom/optimizer/Optimization.cpp
mbasmanova: Makes sense. Let's add a short comment here and clarify in the PR description.

axiom/connectors/hive/LocalHiveConnectorMetadata.cpp
```diff
@@ -799,6 +799,43 @@ void deleteDirectoryContents(const std::string& path) {
   closedir(dir);
 }
 
+// Create a temporary directory.
+// Its path contains two parts 'path' as prefix and unique id as suffix.
+std::string createTemporaryDirectory(const std::string& path) {
```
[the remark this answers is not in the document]
mbasmanova: Makes sense. I remember we discussed with Henry that this is a gap in the current implementation. Thank you for fixing this.

CC: @hdikeman

axiom/optimizer/tests/WritePartitionTest.cpp
MBkkt: Fixed

axiom/optimizer/tests/WritePartitionTest.cpp
MBkkt: Deduplicated

axiom/optimizer/tests/WritePartitionTest.cpp
MBkkt: Deduplicated

axiom/optimizer/tests/WriteTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+  auto write1 = lp::PlanBuilder(context)
+                    .values({data})
+                    .tableWrite(
+                        exec::test::kHiveConnectorId,
+                        "test",
+                        lp::WriteKind::kInsert,
+                        {"key1", "key2", "data", "ds"},
+                        exprs({"key1", "key2", "data", "ds"}),
+                        writeOutputType_)
```
MBkkt: This suggestion make sense to me, I will adjust code

axiom/optimizer/tests/WritePartitionTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+class WritePartitionTest : public test::HiveQueriesTestBase {
+ protected:
+  static void SetUpTestCase() {
+    test::HiveQueriesTestBase::SetUpTestCase();
+  }
+
+  static void TearDownTestCase() {
+    test::HiveQueriesTestBase::TearDownTestCase();
+  }
+
```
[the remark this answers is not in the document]
MBkkt: Removed

axiom/optimizer/tests/WritePartitionTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+#include "velox/dwio/parquet/RegisterParquetWriter.h"
+#include "velox/exec/tests/utils/PlanBuilder.h"
+
+namespace facebook::axiom::optimizer {
+namespace {
+
+using namespace velox;
+namespace lp = facebook::axiom::logical_plan;
+
+class WritePartitionTest : public test::HiveQueriesTestBase {
```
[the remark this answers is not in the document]
MBkkt: Renamed

axiom/optimizer/tests/WriteTest.cpp
MBkkt: I think I cannot do this.
There's problem, now code rely on TableWriteNode output type and in general I don't think this can and should be avoided.
But to make it correct I need
1. call beginWrite a lot of earlier, this isn't nice, and I hope we can avoid this.
2. make logical plan depends on connector, you wanted to avoid this?

axiom/optimizer/tests/WriteTest.cpp
```diff
@@ -0,0 +1,230 @@
...
+    test::HiveQueriesTestBase::TearDownTestCase();
+  }
+
+  void SetUp() override {
+    HiveQueriesTestBase::SetUp();
+    connector_ = velox::connector::getConnector(exec::test::kHiveConnectorId);
+    metadata_ = dynamic_cast<connector::hive::LocalHiveConnectorMetadata*>(
+        connector::ConnectorMetadata::metadata(exec::test::kHiveConnectorId));
+    parquet::registerParquetReaderFactory();
+    parquet::registerParquetWriterFactory();
```
MBkkt: I see this test inherit from LocalRunnerTestBase.
I removed (un)registerParquetReaderFactory

axiom/optimizer/tests/WriteTest.cpp
mbasmanova: I think we need to remove outputType from lp::TableWriteNode. It doesn't belong there. Logical plan is effectively a different representation of SQL.

MBkkt: @mbasmanova 

I made commit to make it connector specific
https://github.com/facebookincubator/axiom/pull/479/commits/d1d1b1b8fa2867d4e39da532798c684ff7041dab

But how such queries will look in LogicalPlan?
```sql
UPDATE products SET price = price * 1.10
WHERE price <= 99.99
RETURNING name, old.price AS old_price, new.price AS new_price,
            new.price - old.price AS price_change
```

MBkkt: In general in postgres DML isn't always root node.

Something like this is also valid and possible :(
But it's even more specific and probably non-standard, so less important than common statements with `RETURNING`
```sql
WITH new_customer AS (
    INSERT INTO customers (name, email)
    VALUES ('John Doe', 'john.doe@example.com')
    RETURNING 'customers', customer_id
),
new_address AS (
    INSERT INTO addresses (customer_id, street, city, country)
    SELECT customer_id, '123 Main St', 'New York', 'USA'
    FROM new_customer
    RETURNING 'addresses', customer_id
)
SELECT * FROM new_customer 
UNION ALL 
SELECT * FROM new_address;
```

hdikeman (changes requested):
Some initial comments, thanks for adding this @MBkkt !

I will take another look tomorrow morning as well

axiom/optimizer/ToGraph.cpp
```diff
@@ -1487,6 +1482,73 @@ PlanObjectP ToGraph::addLimit(const lp::LimitNode& limitNode) {
   return currentDt_;
 }
 
+PlanObjectP ToGraph::addWrite(const lp::TableWriteNode& tableWrite) {
+  const auto writeKind =
+      static_cast<connector::WriteKind>(tableWrite.writeKind());
+  if (writeKind != connector::WriteKind::kInsert) {
```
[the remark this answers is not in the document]
MBkkt: Just because I didn't want to make this PR additionally complex.

I also think implementation can be quite straightforward, as you mentioned

axiom/optimizer/ToVelox.cpp
```diff
@@ -1412,6 +1416,65 @@ velox::core::PlanNodePtr ToVelox::makeValues(
...
+  };
+
+  return std::make_shared<velox::core::TableWriteNode>(
+      nextId(),
+      ROW(std::move(inputNames), std::move(inputTypes)),
+      std::move(columnNames),
+      /*columnStatsSpec=*/std::nullopt,
+      std::make_shared<const velox::core::InsertTableHandle>(
+          connector->connectorId(), handle->veloxHandle()),
+      /*hasPartitioningScheme*/ false,
```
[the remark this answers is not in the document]
MBkkt: Fixed

axiom/connectors/hive/HiveConnectorMetadata.cpp
```diff
@@ -175,7 +175,11 @@ ConnectorWriteHandlePtr HiveConnectorMetadata::beginWrite(
   auto insertHandle =
       std::make_shared<velox::connector::hive::HiveInsertTableHandle>(
           inputColumns,
-          makeLocationHandle(tablePath(table->name()), std::nullopt),
+          makeLocationHandle(
+              tablePath(table->name()),
+              kind == WriteKind::kCreate
```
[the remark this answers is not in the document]
MBkkt: Fixed? https://github.com/facebookincubator/axiom/pull/479/commits/88ef9c68c13144cc425acacb55e772b2384e77c0

I'm not sure about abortWrite in case kCreate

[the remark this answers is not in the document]
hdikeman: What you have now looks good, thanks

axiom/optimizer/ToVelox.cpp
```diff
@@ -1412,6 +1416,65 @@ velox::core::PlanNodePtr ToVelox::makeValues(
...
+  outputTypes.reserve(outputColumns.size());
+  for (const auto* column : outputColumns) {
+    outputNames.push_back(column->outputName());
+    outputTypes.push_back(toTypePtr(column->value().type));
+  }
+
+  auto* connector = table.layouts().front()->connector();
+  auto* metadata = connector::ConnectorMetadata::metadata(connector);
+  auto handle =
+      metadata->beginWrite(table.shared_from_this(), write.kind(), {});
```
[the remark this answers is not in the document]
hdikeman: Fair enough, I don't have a strong preference

Thanks

axiom/runner/LocalRunner.cpp
hdikeman: Realize that the finishWrite interface does not permit this

Perhaps should have been an int64 future, which yields the number of rows

PR #485, opened 2025-10-07
fix: Use inner expression Value for DerivedTable columns

MBkkt:
Also,
* Do not create extra Columns for identity projections.
* Rename `pString` to `lpString` to be consistent with `leString`, these functions are used only with the debugger.

### DIFF
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -108,40 +108,35 @@ ToGraph::ToGraph(
   }
 }
 
-void ToGraph::setDtOutput(
-    DerivedTableP dt,
-    const lp::LogicalPlanNode& logicalPlan) {
-  const auto& outputType = logicalPlan.outputType();
-  for (auto i = 0; i < outputType->size(); ++i) {
-    const auto& type = outputType->childAt(i);
-    const auto& name = outputType->nameOf(i);
+void ToGraph::addDtColumn(DerivedTableP dt, std::string_view name) {
+  const auto* inner = translateColumn(name);
+  dt->exprs.push_back(inner);
 
-    auto inner = translateColumn(name);
-    dt->exprs.push_back(inner);
-
-    Value value(toType(type), 0);
+  ColumnCP outer = nullptr;
+  if (inner->isColumn() && inner->as<Column>()->relation() == dt &&
+      inner->as<Column>()->outputName() == name) {
+    outer = inner->as<Column>();
+  } else {
     const auto* columnName = toName(name);
-    auto* outer = make<Column>(columnName, dt, value, columnName);
-    dt->columns.push_back(outer);
-    renames_[name] = outer;
+    outer = make<Column>(columnName, dt, inner->value(), columnName);
+  }
+  dt->columns.push_back(outer);
+  renames_[name] = outer;
+}
+
+void ToGraph::setDtOutput(DerivedTableP dt, const lp::LogicalPlanNode& node) {
+  const auto& type = *node.outputType();
+  for (const auto& name : type.names()) {
+    addDtColumn(dt, name);
   }
 }
 
 void ToGraph::setDtUsedOutput(
     DerivedTableP dt,
     const lp::LogicalPlanNode& node) {
-  const auto& type = node.outputType();
+  const auto& type = *node.outputType();
   for (auto i : usedChannels(node)) {
-    const auto& name = type->nameOf(i);
-
-    const auto* inner = translateColumn(name);
-    dt->exprs.push_back(inner);
-
-    const auto* columnName = toName(name);
-    const auto* outer =
-        make<Column>(columnName, dt, inner->value(), columnName);
-    dt->columns.push_back(outer);
-    renames_[name] = outer;
+    addDtColumn(dt, type.nameOf(i));
   }
 }
 
@@ -1850,7 +1845,7 @@ extern std::string leString(const lp::Expr* e) {
   return lp::ExprPrinter::toText(*e);
 }
 
-extern std::string pString(const lp::LogicalPlanNode* p) {
+extern std::string lpString(const lp::LogicalPlanNode* p) {
   return lp::PlanPrinter::toText(*p);
 }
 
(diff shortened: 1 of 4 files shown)


axiom/optimizer/ToGraph.h
```diff
@@ -415,6 +413,8 @@ class ToGraph {
       const logical_plan::LogicalPlanNode& node,
       DerivedTableP outerDt = nullptr);
 
+  void addDtColumn(DerivedTableP dt, const velox::RowType& type, uint32_t i);
```
mbasmanova, line 416: Would you document this method? What is 'i'?

axiom/optimizer/ToGraph.cpp
```diff
@@ -108,22 +108,31 @@ ToGraph::ToGraph(
...
+void ToGraph::addDtColumn(
     DerivedTableP dt,
-    const lp::LogicalPlanNode& logicalPlan) {
-  const auto& outputType = logicalPlan.outputType();
-  for (auto i = 0; i < outputType->size(); ++i) {
-    const auto& type = outputType->childAt(i);
-    const auto& name = outputType->nameOf(i);
+    const velox::RowType& type,
+    uint32_t i) {
+  const auto& name = type.nameOf(i);
```
mbasmanova, line 115: Do we use type and i just to get the name? Can we instead pass the name? and move type.nameOf(i) to the caller?

axiom/optimizer/ToGraph.cpp
```diff
@@ -108,22 +108,31 @@ ToGraph::ToGraph(
...
 
-    auto inner = translateColumn(name);
-    dt->exprs.push_back(inner);
+  const auto* inner = translateColumn(name);
+  dt->exprs.push_back(inner);
 
-    Value value(toType(type), 0);
```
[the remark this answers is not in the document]
mbasmanova: I'll need to spend some time to understand this. It would be helpful if you could share some more details.

axiom/optimizer/ToGraph.cpp
MBkkt: Yes. Yes. Done.

axiom/optimizer/ToGraph.cpp
```diff
@@ -108,40 +108,35 @@ ToGraph::ToGraph(
...
-
-    Value value(toType(type), 0);
+void ToGraph::addDtColumn(DerivedTableP dt, std::string_view name) {
+  const auto* inner = translateColumn(name);
+  dt->exprs.push_back(inner);
+
+  ColumnCP outer = nullptr;
+  if (inner->isColumn() && inner->as<Column>()->relation() == dt &&
+      inner->as<Column>()->outputName() == name) {
+    outer = inner->as<Column>();
```
mbasmanova, line 118: casts are not cheap; can we avoid casting 3 times here?

MBkkt: This is `reinterpret_cast<const T*>` so it should be noop for compiler

axiom/optimizer/ToGraph.cpp
```diff
@@ -108,40 +108,35 @@ ToGraph::ToGraph(
...
-    dt->exprs.push_back(inner);
-
-    Value value(toType(type), 0);
+void ToGraph::addDtColumn(DerivedTableP dt, std::string_view name) {
+  const auto* inner = translateColumn(name);
+  dt->exprs.push_back(inner);
+
+  ColumnCP outer = nullptr;
+  if (inner->isColumn() && inner->as<Column>()->relation() == dt &&
+      inner->as<Column>()->outputName() == name) {
```
mbasmanova, line 117: > inner->as<Column>()->outputName() == name

I'm not sure I understand why this is correct. Maybe this is Ok for ToGraph::setDtOutput, where dt is the root DT, but why would this be Ok for ToGraph::setDtUsedOutput which is called for all DTs, not just root one.

> // Adds a column 'name' from 'dt' to the root DerivedTable.

This comment doesn't seem correct as the function is called not only for root DT, but for all DTs.

Am I missing something?

axiom/optimizer/ToGraph.cpp
mbasmanova: > This is reinterpret_cast<const T*> so it should be noop for compiler

I didn't realize that... this seems not very safe. Should we add a DCHECK or something to ensure T is the right type?

```
  /// Returns 'this' as const T.
  template <typename T>
  const T* as() const {
    return reinterpret_cast<const T*>(this);
  }
```

axiom/optimizer/ToGraph.cpp
MBkkt: > I'm not sure I understand why this is correct. Maybe this is Ok for ToGraph::setDtOutput, where dt is the root DT, but why would this be Ok for ToGraph::setDtUsedOutput which is called for all DTs, not just root one.

Because of this condition `inner->as<Column>()->relation() == dt`
Without this condition it's incorrect.


> This comment doesn't seem correct as the function is called not only for root DT, but for all DTs.

Yes, should be `Adds a column 'name' from current DerivedTable to the 'dt'`.

axiom/optimizer/ToGraph.cpp
MBkkt: https://github.com/facebookincubator/axiom/pull/485/commits/efe24f42652f71aa9d9fe004cd471e5c9ab76536

MBkkt:
rebased and squashed (to be able cherry-pick in table write PR)

axiom/optimizer/ToGraph.cpp
```diff
@@ -108,40 +108,35 @@ ToGraph::ToGraph(
...
-    auto inner = translateColumn(name);
-    dt->exprs.push_back(inner);
-
-    Value value(toType(type), 0);
+void ToGraph::addDtColumn(DerivedTableP dt, std::string_view name) {
+  const auto* inner = translateColumn(name);
+  dt->exprs.push_back(inner);
+
+  ColumnCP outer = nullptr;
+  if (inner->isColumn() && inner->as<Column>()->relation() == dt &&
```
mbasmanova, line 116: Wondering, when would "inner->as<Column>()->relation() == dt" be false?

mbasmanova (approved):
Thanks.

PR #486, opened 2025-10-07
refactor: Use move in LocalRunner and MultiFragmentPlan ctors

MBkkt:


mbasmanova:
Got it. Thank you for clarifying.

PR #488, opened 2025-10-08
feat: Implement LocalHiveConnectorMetadata for WriteKind::kInsert

MBkkt:
Now all files in case of kInsert created in directory that is specific for this `beginWrite`.
This allows us to safely delete this directory if we will abort this write.

This also contains small fix: rowIdHandles should accept Table instead of TableLayout because we decided that TableLayout is read-only API


axiom/connectors/hive/HiveConnectorMetadata.h
```diff
@@ -183,6 +183,9 @@ class HiveConnectorMetadata : public ConnectorMetadata {
       WriteKind kind,
       const ConnectorSessionPtr& session) override;
 
+  velox::RowTypePtr writeResultType(const Table& table, WriteKind kind)
```
mbasmanova, line 186: Can we move this into ConnectorWriteHandle?

Alternatively, if this doesn't depend on table or kind, we can keep the API at this level and just remove arguments.

MBkkt: Done

axiom/connectors/hive/LocalHiveConnectorMetadata.cpp
```diff
@@ -799,6 +799,41 @@ void deleteDirectoryContents(const std::string& path) {
...
+  for (const auto& entry : fs::directory_iterator(sourceDir)) {
+    // Compute the relative path from the source directory
+    fs::path relPath = fs::relative(entry.path(), sourceDir);
+    fs::path destPath = targetDir / relPath;
+    // Create enclosing directories in the target if they don't exist
+    fs::create_directories(destPath.parent_path());
+    // Move the file/directory to the target directory
+    fs::rename(entry.path(), destPath);
+  }
+}
```
MBkkt, line 835: I'm not sure, is this really correct for hive?
Isn't possible that directory with same name already exist?

@hdikeman I blindly applied your suggestion here (about removing recursive iterator).

axiom/connectors/hive/HiveConnectorMetadata.cpp
```diff
@@ -172,17 +172,23 @@ ConnectorWriteHandlePtr HiveConnectorMetadata::beginWrite(
...
+              tablePath(table->name()), makeStagingDirectory(table->name())),
           storageFormat,
           bucketProperty,
           compressionKind,
           serdeParameters,
           writerOptions);
   return std::make_shared<HiveConnectorWriteHandle>(
-      std::move(insertHandle), table, kind);
+      velox::ROW(
+          {"numWrittenRows", "fragment", "tableCommitContext"},
```
mbasmanova, line 187: We may need to make this extensible (similar to makeLocationHandle) in a follow-up.

mbasmanova (approved):
Thanks.

PR #490, opened 2025-10-08
feat: Use aggregate function metadata to ignore DISTINCT/ORDER BY

pashandor789:


axiom/optimizer/tests/AggregationPlanTest.cpp
MBkkt, line 1: We need to check it works for distributed queries too

axiom/optimizer/ToGraph.cpp
```diff
@@ -1062,28 +1076,37 @@ AggregationPlanCP ToGraph::translateAggregation(const lp::AggregateNode& agg) {
...
+        }
+
+        std::stringstream error;
+        error << "Aggregate function signature is not supported: "
+              << velox::exec::toString(aggregate->name(), argTypes)
+              << ". Supported signatures: "
+              << velox::exec::toString(aggrEntry->signatures) << ".";
+        VELOX_USER_FAIL(error.str());
+      };
+      const auto* accumulatorType = resolveAccumulatorType();
```
MBkkt, line 1101: I don't think we want to duplicate this code with velox.
Can we use old code/make function in velox (that will be used in both cases)?

axiom/optimizer/ToGraph.cpp
```diff
@@ -1037,14 +1040,25 @@ AggregationPlanCP ToGraph::translateAggregation(const lp::AggregateNode& agg) {
...
+        aggrEntry, "Aggregate function not registered: {}", aggregate->name());
+    const auto& metadata = aggrEntry->metadata;
 
-    if (aggregate->isDistinct() && !orderKeys.empty()) {
+    ExprVector orderKeys;
+    OrderTypeVector orderTypes;
+    if (metadata.orderSensitive) {
+      std::tie(orderKeys, orderTypes) = dedupOrdering(aggregate->ordering());
+    }
+    bool isDistinct = aggregate->isDistinct() && !metadata.ignoreDuplicates;
```
MBkkt, line 1054: NIT
make it consistent
```suggestion
    const bool isDistinct = !metadata.ignoreDuplicates && aggregate->isDistinct();

    ExprVector orderKeys;
    OrderTypeVector orderTypes;
    if (metadata.orderSensitive) {
      std::tie(orderKeys, orderTypes) = dedupOrdering(aggregate->ordering());
    }
```

PR #498, opened 2025-10-12
fix: Fix distribution and plan enumeration

### LINKED ISSUE #549: NextJoin::isWorse logic is incorrect
We have ‘this’ plan and ‘other’. isWorse should return true if there is no way ‘this’ is cheaper than ‘other’. 

If both plans have same distribution, then we just compare cost(this) to cost(other) and return true if cost(this) is larger. 

However, if distributions are different, then we should check if cost(other) + cost(shuffle) is still less then cost(this). In other words, 'other' is so cheap that even after adding shuffle it is still cheaper than 'this' w/o shuffle. Obviously, other is much better plan and no point considering 'this'.

However, we check cost(this) + cost(shuffle) against cost(other). The fact that this plan is costlier than other after adding shuffle doesn’t mean this plan is no good. The shuffle may not be needed if other’s distribution is not interesting.

@MBkkt I believe you have discovered this issue earlier and commented about it somewhere. I cannot find this comment easily. If you have a link handy, please, add here. Do you already have a PR to fix this? If not, I'll make one.

MBkkt:
If any questions "why", please ask, I spent some time in trying to understand what current code wants and what is actually written.

In short I did following:
- Remove `DistributionType::numPartitions`, because it was unused and confusing
- Remove `DistributionType::mode`, because it was unused and confusing
- Remove `isSamePartition` because it was incorrect for broadcast at least
- Add `needsShuffle` instead of `isSamePartition`, it will work correct for broadcast and will allow more cases
- Replace `isSameOrder` with `needsSort`, now if something has sort over "a, b" we don't need to resort for sort "a"
- Fix plan enumeration and shuffle costs they're both was incorrect

Some Notes:

All current tests are passed except few with incorrect join order expected in the tests, such tests adjusted

In most places Cost::cost used as plan cost, plan input cardinality is always 1, so we can omit inputCardinality from computation for them. But in such case function will became invalid for relation. If needed we can have two functions 

Currently input/result cardinality doesn't handle something like `select count(*) from (select 1 limit 0)` 
Because cardinality stored as fanout and multiply doesn't handle this case in a good way.
I have idea how to fix this, but it will be done in the next PR (because requires some refactoring in how we process cost)
```cpp
// select count(*) from (select 1 limit 0)
// i -- input cardinality
// f -- fanout
// r -- result cardinality (i * f) 
// value scan  i=1 f=1 r=1
// project     i=1 f=1 r=1
// limit       i=1 f=0 r=0
// aggregation i=0 f=? r=1 (now f will be NaN, because divide by zero)
```

### Follow ups:
1. Now shuffle cost for broadcast, gather and partition is same, I'm not sure it's right, especially for broadcast
   - As workaround we can multiply broadcast shuffle cost part by number of workers.
2. I'm not sure how velox works, do we keep order of rows after broadcast? If yes, we can keep everything in Distribution::broadcast and only replace DistributionType. It can allow us to enumerate some interesting plans in future, but I think it can be bad in terms of choice of best plan before we fix 1.

### DIFF
--- a/axiom/optimizer/Optimization.cpp
+++ b/axiom/optimizer/Optimization.cpp
@@ -478,33 +478,33 @@ bool isIndexColocated(
     const IndexInfo& info,
     const ExprVector& lookupValues,
     const RelationOpPtr& input) {
-  const auto& distribution = info.index->distribution;
-  if (distribution.isBroadcast) {
-    return true;
+  const auto& current = input->distribution();
+  const auto& desired = info.index->distribution;
+  if (const auto needsShuffle = current.maybeNeedsShuffle(desired)) {
+    return !*needsShuffle;
   }
 
+  // TODO: Code in this function actually doesn't feel right.
+
+  if (!hasCopartition(current.partitionType(), desired.partitionType())) {
+    return false;
+  }
+
+  if (current.partition.empty()) {
+    return false;
+  }
+
+  if (current.partition.size() != desired.partition.size()) {
+    return false;
+  }
+
+  // We should check it when we will add indexes.
   // True if 'input' is partitioned so that each partitioning key is joined to
   // the corresponding partition key in 'info'.
-  if (input->distribution().distributionType != distribution.distributionType) {
-    return false;
-  }
-
-  if (input->distribution().partition.empty()) {
-    return false;
-  }
-
-  if (input->distribution().partition.size() != distribution.partition.size()) {
-    return false;
-  }
-
-  for (auto i = 0; i < input->distribution().partition.size(); ++i) {
-    auto nthKey = position(lookupValues, *input->distribution().partition[i]);
-    if (nthKey != kNotFound) {
-      if (info.schemaColumn(info.lookupKeys.at(nthKey)) !=
-          distribution.partition.at(i)) {
-        return false;
-      }
-    } else {
+  for (size_t i = 0; i < current.partition.size(); ++i) {
+    auto nthKey = position(lookupValues, *current.partition[i]);
+    if (nthKey == kNotFound ||
+        info.schemaColumn(info.lookupKeys[nthKey]) != desired.partition[i]) {
       return false;
     }
   }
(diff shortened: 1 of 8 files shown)


mbasmanova (commented):
This is a lot of changes. I'm not sure how to process these. Is there a way to break up the PR into smaller pieces?

MBkkt:
@mbasmanova 

> This is a lot of changes. I'm not sure how to process these. Is there a way to break up the PR into smaller pieces?

Not really a lof, it's 3 hundreds, not 3 thousands, unfortunately I'm not sure how to make it simpler.
I think there are few things that can be separated, like
- Locus, numPartitions, mode removal
- Use current code to add layouts as they supposed to be added (I'm not sure how wrong `isSamePartition`, etc will react on this)
- fix for missed unnest inputCardinality
- some refactoring lines, e.g. pop_back vs erase, etc (some of them requires "fix missed unnest inputCardinality")

But I also think main complexity of this PR, not in these small parts, but in -- replace `isSamePartition` and `isSameOrder` with `needsShuffle` and `needsSort` + change plan enumeration to be correct and use to these new methods in a correct way

Do you sure separation to such pieces will help you to process them?
If yes, I will do this.

In general I think there's still issues with PlanState (cost in PlanState isn't always corresponds to "current Plan"), but I plan to fix this separately, otherwise PR will be thousands loc...
For an example such asserts make sense logically, but fails with main (and this branch) code and tests.
I plan to fix this, but if I will do this in this PR it will be thousands lines.
image

mbasmanova:
> Remove Locus because it's unused now, and if partition will be connector dependent we won't need this

Can this be a separate PR?

> Use current connector interface to add all connector table layouts

Would you explain this a bit more? Can this be a separate PR?

MBkkt:
> Can this be a separate PR?

https://github.com/facebookincubator/axiom/pull/503

> Would you explain this a bit more? Can this be a separate PR?

https://github.com/facebookincubator/axiom/pull/504

MBkkt:
@mbasmanova tpch.q9 in this PR starts to fail because of this https://github.com/facebookincubator/axiom/pull/527
Can you look and give me advice what is to do?

PR #499, opened 2025-10-14
feat: Add support for INSERT INTO SQL

mbasmanova:
Summary: Add support for queries like INSERT INTO test SELECT ...

Earlier changes added support for INSERT INTO logical plans. This change adds missing pieces to support same queries expressed in SQL.

Differential Revision: D84600741


xiaoxmeng (approved):
@mbasmanova thanks!

PR #502, opened 2025-10-14
fix: Compute input cardinality for Unnest relation

MBkkt:


axiom/optimizer/RelationOp.cpp
```diff
@@ -498,10 +498,12 @@ Unnest::Unnest(
     ExprVector replicateColumns,
     ExprVector unnestExprs,
     ColumnVector unnestedColumns)
-    : RelationOp{RelType::kUnnest, input, input->distribution(), concatColumns(replicateColumns, unnestedColumns)},
+    : RelationOp{RelType::kUnnest, std::move(input), concatColumns(replicateColumns, unnestedColumns)},
```
mbasmanova: this change is a no-op?

axiom/optimizer/RelationOp.cpp
```diff
@@ -498,10 +498,12 @@ Unnest::Unnest(
...
       replicateColumns{std::move(replicateColumns)},
       unnestExprs{std::move(unnestExprs)},
-      unnestedColumns{std::move(unnestedColumns)} {}
+      unnestedColumns{std::move(unnestedColumns)} {
+  cost_.inputCardinality = inputCardinality();
```
mbasmanova: This is the fix, right?

mbasmanova (approved):
Thanks.

PR #503, opened 2025-10-14
refactor: Remove unused code

MBkkt:


axiom/optimizer/Optimization.cpp
```diff
@@ -427,10 +427,9 @@ RelationOpPtr repartitionForAgg(const RelationOpPtr& plan, PlanState& state) {
 
   const auto* agg = state.dt->aggregation;
 
-  // If no grouping and not yet gathered on a single node, add a gather before
-  // final agg.
-  if (agg->groupingKeys().empty() &&
-      !plan->distribution().distributionType.isGather) {
```
MBkkt: Just in case: few lines above written "if" with opposite condition + return

axiom/optimizer/Optimization.cpp
```diff
@@ -1568,9 +1558,12 @@ PlanP unionPlan(
   for (auto i = 1; i < states.size(); ++i) {
     const auto& otherCost = states[i].cost;
     fullyImported.intersect(inputPlans[i]->fullyImported);
-    firstState.cost.add(otherCost);
-    // The input cardinality is not additive, the fanout and other metrics are.
-    firstState.cost.inputCardinality -= otherCost.inputCardinality;
```
MBkkt: About this code.
add is incorrect and used only here.

In add we are also lost `totalBytes` and `transferBytes`

axiom/optimizer/Plan.h
```diff
@@ -81,9 +81,9 @@ struct PlanSet {
   /// Interesting equivalent plans.
   std::vector<std::unique_ptr<Plan>> plans;
 
-  /// Cost of lowest-cost plan plus shuffle. If a cutoff is applicable, nothing
-  /// more expensive than this should be tried.
-  float bestCostWithShuffle{0};
+  /// Cost of lowest-cost plan plus shuffle. If a cutoff is applicable,
+  /// nothing more expensive than this should be tried.
+  float bestCostWithShuffle{std::numeric_limits<float>::infinity()};
```
MBkkt: This is just to remove two conditions

mbasmanova (commented):
I see there are 2 types of changes related to (1) cost and (2) distribution. I understand the changes to 'cost', but not yet to 'distribution'. Would it be possible to extract 'cost' fixes / refactors into a separate PR?

axiom/optimizer/RelationOp.cpp
```diff
@@ -732,31 +732,13 @@ std::string Project::toString(bool recursive, bool detail) const {
...
-} // namespace
-
 OrderBy::OrderBy(
-    const RelationOpPtr& input,
+    RelationOpPtr input,
     ExprVector orderKeys,
     OrderTypeVector orderTypes,
     int64_t limit,
     int64_t offset)
-    : RelationOp{RelType::kOrderBy, input, makeOrderByDistribution(input, std::move(orderKeys), std::move(orderTypes))},
```
mbasmanova: This change doesn't seem right. The output of OrderBy is sorted, hence, we'd want 'distribution' to reflect that, no?

mbasmanova: > Distribution::gather(std::move(orderKeys), std::move(orderTypes)) will do this.

Sorry, I missed that you are passing sorting keys to Distribution::gather(). Read this as just Distribution::gather(). Thank you for clarifying.

MBkkt:
@mbasmanova I also updated this PR to avoid having same changes (removed "cost" changes from this PR)

mbasmanova (approved):
Thank you. I believe I understand the changes now. Splitting the PR really helped a lot.

PR #505, opened 2025-10-15
feat: Add support for table properties in CTAS SQL

mbasmanova:
Summary: Enable parsing of queries like CREATE TABLE t WITH (partitioned_by = ARRAY['ds']) AS SELECT ...

Differential Revision: D84667709


xiaoxmeng (approved):
@mbasmanova thanks!

PR #509, opened 2025-10-15
fix: Remove  Locus

MBkkt:
Locus was introduced to represent different storage systems being accessed in a single query. We believe we do not need a separate abstraction and can use Connector instead.


mbasmanova (approved):
Thanks. I think the original PR description was incorrect. I edited it. Please, take a look to see if it makes sense.

PR #511, opened 2025-10-15
feat: Add optimizer support for writing partitioned tables

mbasmanova:
Summary: In multi-node setup, when writing to partitioned table (bucketed Hive tables), check if input is already partitioned in a compatible way. If not, add a shuffle.

In single-node multi-threaded setup, add local exchange before partitioned write.

Check the table layout to determine whether write is partitioned or not. Tables with multiple layouts are not supported yet. To support these, we'll need to introduce layout-for-write or primary layout or similar.

Differential Revision: D84718419

### DIFF
--- a/axiom/connectors/ConnectorMetadata.h
+++ b/axiom/connectors/ConnectorMetadata.h
@@ -32,6 +32,8 @@ class ITypedExpr;
 using TypedExprPtr = std::shared_ptr<const ITypedExpr>;
 
 class PartitionFunctionSpec;
+using PartitionFunctionSpecPtr =
+    std::shared_ptr<const core::PartitionFunctionSpec>;
 } // namespace facebook::velox::core
 
 /// Base classes for schema elements used in execution. A ConnectorMetadata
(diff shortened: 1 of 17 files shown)


axiom/connectors/ConnectorMetadata.h
```diff
@@ -182,6 +184,46 @@ struct SortOrder {
...
+  /// keys. Compatibility is not strict equality in the case of e.g. Hive where
+  /// a dataset partitioned 8 ways is compatible with one partitioned 16 ways if
+  /// the function is the same. In such a case the partition to use is the 8 way
+  /// one. On the 16 side data from partitions 0 and 1 match 0 on the 8 side and
+  /// 2, 3 match 1 and so on.
+  virtual const PartitionType* copartition(
+      const PartitionType& other) const = 0;
+
+  /// Returns the types of the partitioning keys. Empty if no partitioning.
+  virtual const std::vector<velox::TypePtr>& partitionKeyTypes() const = 0;
```
MBkkt: It's unused (used only in one implementation, we don't need this method to be in connector interface), can you remove it?

axiom/optimizer/Optimization.cpp
```diff
@@ -699,6 +698,71 @@ const RelationOpPtr& maybeDropProject(const RelationOpPtr& plan) {
...
+  }
+
+  const auto& tableSchema = write->table().type();
+
+  // Find values for all partition columns.
+  ExprVector keyValues;
+  keyValues.reserve(partitionColumns.size());
+  for (const auto* column : partitionColumns) {
+    const auto index = tableSchema->getChildIdx(column->name());
+    keyValues.emplace_back(write->columnExprs().at(index));
```
MBkkt: This is bug
`columnExpr` not 1:1 with `write->table().type();` 
it's 1:1 with `columnNames` that can be subset and reordered of `write->table().type();` 

As simple solution you can just use linear search for columnNames, but in general i think we should create hashmap in logical plan name => expr

MBkkt: I'm wrong about this, I forgot that we reordered this

axiom/optimizer/Optimization.cpp
```diff
@@ -699,6 +698,71 @@ const RelationOpPtr& maybeDropProject(const RelationOpPtr& plan) {
...
+  const auto* planPartitionType =
+      plan->distribution().distributionType.partitionType();
+
+  auto copartition =
+      copartitionType(planPartitionType, layout->partitionType());
+
+  // Copartitioning is possible if PartitionTypes are compatible and the table
+  // has no fewer partitions than the plan.
+  bool shuffle = !copartition || copartition != planPartitionType;
+  if (!shuffle) {
```
MBkkt: This doesn't work for local hive

Local hive has such lines
```
 if (numPartitions_ % otherPartitionType->numPartitions_ == 0) {
        return otherPartitionType;
      }
```
so when plan and layout partitionType will have same number of buckets you will return right argument, not left.

So copartition != planPartitionType will be true.

axiom/connectors/ConnectorMetadata.h
```diff
@@ -182,6 +184,46 @@ struct SortOrder {
...
+  /// Returns 'this' or '&other' if the partitions are compatible. Partitions
+  /// are compatible if data in one partitioned dataset can only match data in
+  /// the same partition of another dataset if joined on equality of partition
+  /// keys. Compatibility is not strict equality in the case of e.g. Hive where
+  /// a dataset partitioned 8 ways is compatible with one partitioned 16 ways if
+  /// the function is the same. In such a case the partition to use is the 8 way
+  /// one. On the 16 side data from partitions 0 and 1 match 0 on the 8 side and
+  /// 2, 3 match 1 and so on.
+  virtual const PartitionType* copartition(
+      const PartitionType& other) const = 0;
```
MBkkt: I think we need to clarify this:
1. return nullptr => shuffle needed (example, different bucket keys for hive?)
2. return this => nothing needed (example, keys same and  `this->numBuckets % other->numBuckets == 0`)
3. return some other ptr => needs shuffle but can be optimized shuffle (e.g.  `other->numBuckets % this->numBuckets`, we can split each input partition to some other partitions)

instead of `other` name for argument, IMO it's better `desired`, because `this` is commonly current plan root plan, and `other` is commonly partition type needed by next relation operator.

axiom/optimizer/Optimization.cpp
mbasmanova: Fixed.

PR #512, opened 2025-10-16
fix: Replace Cost::add with correct logic

MBkkt:


axiom/optimizer/Optimization.cpp
```diff
@@ -1565,9 +1565,12 @@ PlanP unionPlan(
   for (auto i = 1; i < states.size(); ++i) {
     const auto& otherCost = states[i].cost;
     fullyImported.intersect(inputPlans[i]->fullyImported);
-    firstState.cost.add(otherCost);
-    // The input cardinality is not additive, the fanout and other metrics are.
-    firstState.cost.inputCardinality -= otherCost.inputCardinality;
+    // The input cardinality is not additive.
```
mbasmanova: This comment doesn't seem to refer to the following code. Is it intended to explain why there is no code to add up input cardinalities?

MBkkt: > Is it intended to explain why there is no code to add up input cardinalities?

Yes, I keep it because of this

mbasmanova:
@MBkkt No worries. I rebased and landed.

PR #527, opened 2025-10-19
fix: Fix bug that prevents us from explore plan space

MBkkt:
Same issue reproduced when isWorse always returns false.

`isWorse` used for following logic if `plan` worse than `other.plan` then we need to remove plan.

If they have same partition it makes sense that we compare just their cost.

But if they're not same we should compare plan.cost against other.plan.cost + shuffle. Because we want to know that even after reshuffle other.plan will be cheaper than plan

Is everything else ok? No all code is full of such bugs...
(for an example here we're also should multiply unitCost by inputCardinality)

Unfortunately when I fix this, I found that in tpch q9 we starts to explore more plans, and some of this plans generated invalid. I don't understand what code is responsible for generating such invalid plans, and how we can prevent this from happens.

ERROR:
```
E20251019 21:22:50.839110 3090096 Exceptions.h:53] Line: /home/mironov/projects/serenedb/axiom/velox/velox/core/PlanNode.cpp:1502, Function:validate, Expression: leftType->containsChild(key->name()) Left side join key not found in left side output: s_suppkey, Source: RUNTIME, ErrorCode: INVALID_STATE
unknown file: Failure
C++ exception with description "Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Reason: Left side join key not found in left side output: s_suppkey
Retriable: False
Expression: leftType->containsChild(key->name())
```

What is inside leftType, rigthType and outputType of this join
```
Left output column: p_partkey, ps_partkey, ps_suppkey, ps_supplycost
Right output column: l_orderkey, l_partkey, l_suppkey, l_quantity, l_extendedprice, l_discount
Join output column: p_partkey, l_orderkey, l_partkey, l_suppkey, l_quantity, l_extendedprice, l_discount, ps_partkey, ps_suppkey, ps_supplycost,
```

### DIFF
--- a/axiom/optimizer/Plan.cpp
+++ b/axiom/optimizer/Plan.cpp
@@ -491,9 +491,9 @@ bool NextJoin::isWorse(const NextJoin& other) const {
   float shuffle =
       plan->distribution().isSamePartition(other.plan->distribution())
       ? 0
-      : plan->cost().fanout * shuffleCost(plan->columns());
-  return cost.unitCost + cost.setupCost + shuffle >
-      other.cost.unitCost + other.cost.setupCost;
+      : other.cost.fanout * shuffleCost(other.plan->columns());
+  return cost.unitCost + cost.setupCost >
+      other.cost.unitCost + other.cost.setupCost + shuffle;
 }
 
 size_t MemoKey::hash() const {


mbasmanova:
@MBkkt I'm not able to follow the PR description. Would it be possible to explain what is the original problem you are trying to fix? I.e. is there a query that produces the wrong plan? Would be helpful to show the query, the plan being produces and the optimal plan that should be produced. Then, explain how the implementation of isWorse is preventing the Optimizer from choosing the optimal plan.

MBkkt:
@mbasmanova 

> I see that this causes an error: Reason: Left side join key not found in left side output: s_suppkey

Yes I know, I tried to debug it and understand why, but I didn't succeed.
Because I don't understand current join optimization code related to producing additional semi filter joins ("edt" derived tables, etc). So I'm unable to understand what author of code wanted here, some comments and asserts maybe can helps.

From debugging perspective what I see that we create join with leftKeys contains key that is not part of left subtree columns output (but left subtree output contains equivalent key `ps_suppkey` and `s_suppkey` somewhere deeper).
I don't know 
1. maybe it's bug in `downstreamColumns`
2. maybe in what keys we're adding to this join edge (this keys adding additionally later, because other optimizer code)
3. maybe in something else
 
I spend on debugging this some time, and unfortunately it looks pointless to me for now, I just don't have enough context what is author tried to do.

In serenedb I just set `noImportOfExists` to `true` as a temporary solution.
Sad thing about this, there's no single test for this behavior, see https://github.com/facebookincubator/axiom/pull/498/commits/79685d1cee5a5c25c494af9fe33b3dc61bdc8292

> We need to debug and fix that. Let's create a GitHub issue.

Created, https://github.com/facebookincubator/axiom/issues/530

PR #542, opened 2025-10-22
fix: Drop unnecessary OrderBy from QueryGraph

### LINKED ISSUE #571: Flatten Set Operations
- [ ] `UNION` should flatten all inner `UNION` and `UNION ALL`
- [ ] `UNION ALL` should flatten all inner `UNION ALL`
- [ ] `INTERSECT` should flatten all inner `INTERSECT` and `INTERSECT ALL`
- [ ] `INTERSECT ALL` should flatten all inner `INTERSECT ALL`
- [ ] `EXCEPT` should flatten **first** inner `EXCEPT` and `EXCEPT ALL`
- [ ] `EXCEPT ALL` should flatten **first** inner `EXCEPT ALL`

MBkkt:
Order by is dropped when it's not needed (before union/intersects/join/aggregation/order by/etc). So instead of old code when we reassign order by, now we just skip logical plan order by if it's not needed.

The one special case exists, if we found limit, we should re enable order by back.

Follow up:
Ideally we want to distinguish between Ordered TopN and just Unordered TopN, now it's kind of impossible

CC: @pashandor789

### DIFF
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -847,6 +847,9 @@ constexpr uint64_t deny(uint64_t mask, T... op) {
   return (mask & ... & ~allow(op));
 }
 
+constexpr uint64_t kUnorderedAllowedInDt =
+    deny(kAllAllowedInDt, lp::NodeKind::kSort);
+
 } // namespace
 
 std::optional<ExprCP> ToGraph::translateSubfieldFunction(
@@ -1209,6 +1212,9 @@ AggregationPlanCP ToGraph::translateAggregation(const lp::AggregateNode& agg) {
 }
 
 void ToGraph::addOrderBy(const lp::SortNode& order) {
+  VELOX_DCHECK(currentDt_->orderKeys.empty());
+  VELOX_DCHECK(currentDt_->orderTypes.empty());
+
   auto [deduppedOrderKeys, deduppedOrderTypes] =
       dedupOrdering(order.ordering());
 
@@ -1344,9 +1350,9 @@ DerivedTableP ToGraph::newDt() {
   return dt;
 }
 
-void ToGraph::wrapInDt(const lp::LogicalPlanNode& node) {
+void ToGraph::wrapInDt(const lp::LogicalPlanNode& node, bool unordered) {
   auto* outerDt = std::exchange(currentDt_, newDt());
-  makeQueryGraph(node, kAllAllowedInDt);
+  makeQueryGraph(node, unordered ? kUnorderedAllowedInDt : kAllAllowedInDt);
   finalizeDt(node, outerDt);
 }
 
@@ -1607,7 +1613,7 @@ DerivedTableP ToGraph::translateSubquery(
   VELOX_CHECK(correlatedConjuncts_.empty());
 
   auto* outerDt = std::exchange(currentDt_, newDt());
-  makeQueryGraph(node, kAllAllowedInDt);
+  makeQueryGraph(node, kUnorderedAllowedInDt);
   auto* subqueryDt = currentDt_;
   finalizeSubqueryDt(node, outerDt);
 
@@ -1891,7 +1897,7 @@ bool hasNondeterministic(const lp::ExprPtr& expr) {
 void ToGraph::translateSetJoin(const lp::SetNode& set) {
   auto* setDt = currentDt_;
   for (auto& input : set.inputs()) {
-    wrapInDt(*input);
+    wrapInDt(*input, /*unordered=*/true);
   }
 
   const bool exists = set.operation() == lp::SetOperation::kIntersect;
@@ -2000,7 +2006,7 @@ void ToGraph::translateUnion(const lp::SetNode& set) {
   auto translateUnionInput = [&](const lp::LogicalPlanNode& input) {
     renames_ = renames;
     currentDt_ = newDt();
-    makeQueryGraph(input, kAllAllowedInDt);
+    makeQueryGraph(input, kUnorderedAllowedInDt);
     auto* newDt = std::exchange(currentDt_, setDt);
 
     const auto& type = input.outputType();
@@ -2059,7 +2065,13 @@ void ToGraph::makeQueryGraph(
     const lp::LogicalPlanNode& node,
     uint64_t allowedInDt) {
   if (!contains(allowedInDt, node.kind())) {
-    wrapInDt(node);
+    if (node.kind() == lp::NodeKind::kSort) {
+      // Sort not allowed doesn't mean we need to wrap it in DT,
+      // instead we should skip it.
+      makeQueryGraph(*node.onlyInput(), allowedInDt);
+    } else {
+      wrapInDt(node, /*unordered=*/false);
+    }
     return;
   }
 
@@ -2077,7 +2089,10 @@ void ToGraph::makeQueryGraph(
       const auto& filter = *node.as<lp::FilterNode>();
       if (hasNondeterministic(filter.predicate())) {
         auto* outerDt = std::exchange(currentDt_, newDt());
-        makeQueryGraph(input, kAllAllowedInDt);
+        allowedInDt = contains(allowedInDt, lp::NodeKind::kSort)
+            ? kAllAllowedInDt
+            : kUnorderedAllowedInDt;
+        makeQueryGraph(input, allowedInDt);
         addFilter(filter);
         finalizeDt(node, outerDt);
         break;
@@ -2091,13 +2106,12 @@ void ToGraph::makeQueryGraph(
     } break;
     case lp::NodeKind::kAggregate: {
       const auto& input = *node.onlyInput();
-      makeQueryGraph(input, allowedInDt);
+      makeQueryGraph(input, deny(allowedInDt, lp::NodeKind::kSort));
       if (currentDt_->hasAggregation() || currentDt_->hasLimit()) {
         finalizeDt(input);
-      } else if (currentDt_->hasOrderBy()) {
-        currentDt_->orderKeys.clear();
-        currentDt_->orderTypes.clear();
       }
+      VELOX_DCHECK(currentDt_->orderKeys.empty());
+      VELOX_DCHECK(currentDt_->orderTypes.empty());
 
       auto* agg = translateAggregation(*node.as<lp::AggregateNode>());
 
@@ -2195,13 +2209,14 @@ void ToGraph::makeQueryGraph(
     } break;
     case lp::NodeKind::kSort: {
       const auto& input = *node.onlyInput();
-      makeQueryGraph(input, allowedInDt);
+      makeQueryGraph(input, deny(allowedInDt, lp::NodeKind::kSort));
       if (currentDt_->hasLimit()) {
         finalizeDt(input);
       }
       addOrderBy(*node.as<lp::SortNode>());
     } break;
     case lp::NodeKind::kLimit: {
+      allowedInDt |= allow(lp::NodeKind::kSort);
       makeQueryGraph(*node.onlyInput(), allowedInDt);
       addLimit(*node.as<lp::LimitNode>());
     } break;
@@ -2223,7 +2238,7 @@ void ToGraph::makeQueryGraph(
     } break;
     case lp::NodeKind::kTableWrite: {
       VELOX_DCHECK_EQ(allowedInDt, kAllAllowedInDt);
-      wrapInDt(*node.onlyInput());
+      wrapInDt(*node.onlyInput(), /*unordered=*/true);
       addWrite(*node.as<lp::TableWriteNode>());
     } break;
     default:
--- a/axiom/optimizer/ToGraph.h
+++ b/axiom/optimizer/ToGraph.h
@@ -292,7 +292,7 @@ class ToGraph {
   // DerivedTable. Done for joins to the right of non-inner joins,
   // group bys as non-top operators, whenever descendents of 'node'
   // are not freely reorderable with its parents' descendents.
-  void wrapInDt(const logical_plan::LogicalPlanNode& node);
+  void wrapInDt(const logical_plan::LogicalPlanNode& node, bool unordered);
 
   // Start new DT and add 'currentDt_' as a child.
   // Set 'currentDt_' to the new DT.
--- a/axiom/optimizer/tests/HiveQueriesTest.cpp
+++ b/axiom/optimizer/tests/HiveQueriesTest.cpp
@@ -246,18 +246,13 @@ TEST_F(HiveQueriesTest, joinWithLimitBothSides) {
 
   {
     auto plan = toSingleNodePlan(logicalPlan);
-    // TODO: We don't need orderBy before the join.
-    auto matcher = core::PlanMatcherBuilder()
-                       .tableScan("nation")
-                       .limit()
-                       .orderBy()
-                       .hashJoin(
-                           core::PlanMatcherBuilder()
-                               .tableScan("region")
-                               .limit()
-                               .orderBy()
-                               .build())
-                       .build();
+    auto matcher =
+        core::PlanMatcherBuilder()
+            .tableScan("nation")
+            .limit()
+            .hashJoin(
+                core::PlanMatcherBuilder().tableScan("region").limit().build())
+            .build();
     AXIOM_ASSERT_PLAN(plan, matcher);
   }
 


axiom/optimizer/ToGraph.h
```diff
@@ -200,10 +200,15 @@ class ToGraph {
   // to the left.
   void canonicalizeCall(Name& name, ExprVector& args);
 
-  // Converts 'plan' to PlanObjects and records join edges into
-  // 'currentDt_'. If 'node' does not match  allowedInDt, wraps 'node' in
-  // a new DerivedTable.
-  PlanObjectP makeQueryGraph(
+  DerivedTableP makeStream(
```
mbasmanova: Would it be possible to document these methods?

mbasmanova:
I'm not able to capture the gist of the new design. Would it be possible to describe how the new logic works to transform a tree of logical plan nodes into a tree of derived tables?

mbasmanova:
> But in some cases we need to keep current dt for our current operator (join, and potentially union/etc) in such cases it's possible to make transformation that instead of using current dt for input nodes will create new one and put this new dt (with inputs nodes) into old one (with current)

I'm not following. Would you clarify some more? 

Are we traversing the logical plan tree bottom up and fill up a DT? When DT is "full", we start a new one? How do we handle nodes with multiple inputs? Do we create separate DTs for each input, then combine these if possible?

axiom/optimizer/ToGraph.cpp
```diff
@@ -1667,45 +1625,52 @@ void ToGraph::makeUnionDistributionAndStats(
...
-    auto isUnionLike =
-        [](const lp::LogicalPlanNode& node) -> const lp::SetNode* {
-      if (node.kind() == lp::NodeKind::kSet) {
-        const auto* set = node.asUnchecked<lp::SetNode>();
-        if (set->operation() == lp::SetOperation::kUnion ||
-            set->operation() == lp::SetOperation::kUnionAll) {
-          return set;
-        }
-      }
+  // TODO: use deducing this lambda when C++23 is available.
```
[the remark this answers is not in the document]
MBkkt: ok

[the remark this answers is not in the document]
MBkkt: https://github.com/facebookincubator/axiom/pull/542/commits/15ec906b932b1517d0f283b08ea9cff9e63c4275

fixed

axiom/optimizer/ToGraph.h
MBkkt: Added this and other comments https://github.com/facebookincubator/axiom/pull/542/commits/15ec906b932b1517d0f283b08ea9cff9e63c4275

axiom/optimizer/tests/HiveQueriesTest.cpp
```diff
@@ -157,5 +157,117 @@ TEST_F(HiveQueriesTest, orderOfOperations) {
           .filter("n_nationkey > 70 AND n_regionkey < 7"));
 }
 
+TEST_F(HiveQueriesTest, joinWithTopNBothSides) {
+  lp::PlanBuilder::Context context(exec::test::kHiveConnectorId);
+  auto logicalPlan = lp::PlanBuilder(context)
+                         .tableScan("nation")
+                         .orderBy({"n_nationkey"})
+                         .limit(0, 10)
+                         .as("n1")
```
mbasmanova: Do we need these aliases? I believe not.

mbasmanova (commented):
PR description mentions a number of improvements. Would it be possible to add tests for these?

axiom/optimizer/ToGraph.cpp
```diff
@@ -1481,34 +1452,25 @@ PlanObjectP ToGraph::addFilter(const lp::FilterNode* filter) {
...
-  return currentDt_;
 }
 
-PlanObjectP ToGraph::addAggregation(const lp::AggregateNode& aggNode) {
-  currentDt_->aggregation = translateAggregation(aggNode);
-  return currentDt_;
-}
-
-PlanObjectP ToGraph::addLimit(const lp::LimitNode& limitNode) {
+void ToGraph::addLimit(const lp::LimitNode& limit) {
```
mbasmanova: This is a nice refactoring, but makes it hard to understand material changes. Would it be possible to extract it into a separate PR?

axiom/optimizer/ToGraph.h
```diff
@@ -200,10 +200,34 @@ class ToGraph {
...
+  // but removes order by from currentDt_.
+  // This is useful for cases where we want to ignore the order
+  // Example: set operations, join, write, aggregation inputs.
+  DerivedTableP makeUnordered(
+      const logical_plan::LogicalPlanNode& node,
+      uint64_t allowedInDt);
+
+  // Helper function to call makeQueryGraph but finishes currentDt_
+  // if it will have limit.
+  // This is useful for cases where we want to apply limit before our node.
```
mbasmanova: I believe limit must be the last op added to a DT. Hence, we always want to start a new DT after limit, no?

MBkkt:
@mbasmanova I added test for main changes 

About flatten union I think existing test cover this (they're adjusted, that's why I changed PlanMatcher a little)

About sropped order, It's tested in my tests (`(... limit order by) join ...`

If you have thoughts about some concrete tests, can you share them, I will try to add such tests

axiom/optimizer/ToGraph.h
```diff
@@ -200,10 +200,34 @@ class ToGraph {
...
 
-  // Converts 'plan' to PlanObjects and records join edges into
-  // 'currentDt_'. If 'node' does not match  allowedInDt, wraps 'node' in
-  // a new DerivedTable.
-  PlanObjectP makeQueryGraph(
+  // Helper function to call makeQueryGraph
+  // but removes order by from currentDt_.
+  // This is useful for cases where we want to ignore the order
+  // Example: set operations, join, write, aggregation inputs.
+  DerivedTableP makeUnordered(
```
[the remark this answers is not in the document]
MBkkt: It's same as result of makeQueryGraph.
As for makeQueryGraph I wrote both cases null and not null

axiom/optimizer/tests/HiveQueriesTest.cpp
MBkkt: Yes, I need this, otherwise I don't understand how to write this

`"n1.n_regionkey = r1.r_regionkey",`

axiom/optimizer/tests/HiveQueriesTest.cpp
```diff
@@ -157,5 +157,117 @@ TEST_F(HiveQueriesTest, orderOfOperations) {
...
+  {
+    auto plan = toSingleNodePlan(logicalPlan);
+    auto matcher =
+        core::PlanMatcherBuilder()
+            .tableScan("nation")
+            .topN()
+            .hashJoin(
+                core::PlanMatcherBuilder().tableScan("region").topN().build())
+            .build();
+    ASSERT_TRUE(matcher->match(plan));
```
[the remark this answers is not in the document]
MBkkt: ok

axiom/optimizer/ToGraph.h
```diff
@@ -200,10 +200,34 @@ class ToGraph {
...
+  // and call makeQueryGraph to fill it.
+  void wrapInDt(const logical_plan::LogicalPlanNode& node);
+
+  // Fill currentDt_ from 'node'.
+  // If 'allowedInDt' does not allow 'node', wraps in a new DerivedTable.
+  // Returns nullptr if 'node' is fully processed in currentDt_.
+  // Otherwise returns the "outer" DerivedTable, which needs to be
+  // finalized by the caller.
+  // This is useful for cases where we want to control from caller
+  // do we need to drop order by/limit/aggregation or not.
```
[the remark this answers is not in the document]
MBkkt: There's no such cases, comment adjusted, see https://github.com/facebookincubator/axiom/pull/542/commits/35b20d724979aeb5c8a6c82118acd859457c17b1

axiom/optimizer/tests/HiveQueriesTest.cpp
```diff
@@ -157,5 +157,117 @@ TEST_F(HiveQueriesTest, orderOfOperations) {
...
+  }
+
+  auto planNodeIdGenerator = std::make_shared<core::PlanNodeIdGenerator>();
+  auto referencePlan = exec::test::PlanBuilder(planNodeIdGenerator)
+                           .tableScan("nation", getSchema("nation"))
+                           .orderBy({"n_nationkey"}, false)
+                           .limit(0, 10, false)
```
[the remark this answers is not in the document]
MBkkt: https://github.com/facebookincubator/axiom/pull/542/commits/35b20d724979aeb5c8a6c82118acd859457c17b1

axiom/optimizer/tests/HiveQueriesTest.cpp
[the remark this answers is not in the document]
MBkkt: https://github.com/facebookincubator/axiom/pull/542/commits/35b20d724979aeb5c8a6c82118acd859457c17b1

mbasmanova:
Still trying to wrap my head around the new design. 

If we consider only linear trees (no branches / joins) then it makes sense to fill up a DT while traversing the plan bottom up. Before adding a node, check if it "fits" in current DT and if not, finalize current DT and start a new one.

This doesn't quite work for joins. Let's consider an example:

```
- 4: join
   - 3: agg
     -  2: scan
   - 1: scan
```

We start populating a DT with scan 1, then we add scan 2, then add agg 3, then add join 4. They all end up in the same DT, which is incorrect. The agg 3 must be in a separate DT. Is this the bug that triggered this PR? Just want to double check my understanding of the issue with existing code.

Now, it looks like Join introduces a restriction of what can be added to the current DT. This restriction is different from rules that decide whether an operation fits into a DT or not (bottom up). The join says that only tables and filters are allowed. Anything else must be wrapped in a separate DT. I believe the existing allow-in-dt mechanism is supposed to enforce that. Is this not working / buggy?

To summarize, currently there are 2 mechanism that are supposed to work together to cover all cases. On the way down the tree, join node updates allow-in-dt. If a node is not allowed, then it starts a new DT (wrapInDt logic). On the way up, each node checks whether it fits in a DT and, if not, starts a new DT (finalizeDt logic). The difference is wrapInDt creates an empty DT and fills it up using child nodes. FinalizeDT creates a new DT and places current DT into it as a 'table'.

axiom/optimizer/tests/HiveQueriesTest.cpp
MBkkt: You're right, I missed that they have different prefix

MBkkt:
@mbasmanova I can say it this way:

We have two approaches to produce DT

1. finalize then it's necessary (works for linear operators)
2. create new dt for input (needed for joins)

In first approach we cannot do finalize later.
In second approach with my code we also cannot do this later, but we can do it later with current main code.

So in short I tried to make behavior of two approaches consistent which gives some benefits:
1. correct having
2. remove unnecessary order by
3. maybe some other, I'm not sure, needs to think more about what cases we have.

pansatadru (approved):

mbasmanova (commented):
@MBkkt I'll try to understand this PR again. Is PR description up to date? It doesn't seem so. If there is anything you could clarify to help understanding, that would be much appreciated. Otherwise, I'll need to find bandwidth to run and debug the code to try to figure out how it works.

mbasmanova (commented):
If there is any way to break up this PR into smaller pieces, that would help a lot. Thanks.

axiom/optimizer/DerivedTable.cpp
```diff
@@ -595,17 +594,79 @@ void DerivedTable::importJoinsIntoFirstDt(const DerivedTable* firstDt) {
...
-  joins = dt->joins;
-  joinOrder = dt->joinOrder;
   columns = dt->columns;
   exprs = dt->exprs;
-  fullyImported = dt->fullyImported;
+  joinedBy = dt->joinedBy;
+  tables = dt->tables;
+  tableSet = dt->tableSet;
+
+  // TODO: Is setop possible here at all?
```
mbasmanova: What is flattenDt for? What does it do? Which code paths does it participate in? How are you thinking about this logic?

axiom/optimizer/ToGraph.cpp
```diff
@@ -1913,166 +1869,160 @@ void ToGraph::makeUnionDistributionAndStats(
...
+  wrapInDt(logicalPlan, /*unordered=*/false);
+  return currentDt_;
+}
+
+DerivedTableP ToGraph::makeUnordered(
+    const lp::LogicalPlanNode& node,
+    uint64_t allowedInDt) {
+  auto* outerDt = makeQueryGraph(node, allowedInDt);
+  if (currentDt_->hasOrderBy() && !currentDt_->hasLimit()) {
+    currentDt_->orderKeys.clear();
```
mbasmanova: Is there a way to avoid adding OrderBy in the first place? Removing it works, but it is a bit awkward and we also keep the expressions that may have been added specifically for OrderBy.

axiom/optimizer/DerivedTable.cpp
```diff
@@ -595,17 +594,79 @@ void DerivedTable::importJoinsIntoFirstDt(const DerivedTable* firstDt) {
...
+  VELOX_DCHECK_NOT_NULL(dt);
+  VELOX_DCHECK_EQ(tables.size(), 1);
+  VELOX_DCHECK(tables[0] == dt);
+  VELOX_DCHECK_NULL(dt->write);
+  if (hasAggregation() && dt->hasAggregation()) {
+    // Can't have two aggregations in single dt.
+    return;
+  }
+  if (hasLimit() || dt->hasLimit()) {
+    // TODO: Can we combine limits?
```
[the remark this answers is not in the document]
MBkkt: This isn't true for dt, because if one of them have aggregation, it won't work this way.
Same about cardinality estimation

axiom/optimizer/ToGraph.cpp
MBkkt: I'm not sure how, the first idea is handle allowInDt for order by differently.
But I think it doesn't always work.
My suggestion is to think about this separately

axiom/optimizer/DerivedTable.cpp
MBkkt: flattenDt called when we optimize DT and this DT have single DT in it's tables.

This is possible in two cases:
1) to graph produced such code
2) optimizer imported all joins in single dt

Right now we don't have enough tests for this code + setop, so I'm not sure will or not it called

axiom/optimizer/DerivedTable.h
```diff
@@ -240,11 +240,6 @@ struct DerivedTable : public PlanObject {
 
   // Sets 'columns' and 'exprs'.
   void makeProjection(const ExprVector& exprs);
-
-  /// Number of fully processed leading elements of 'conjuncts'. Set by
-  /// distributeConjuncts(). Used to avoid reprocessing conjuncts in repeated
-  /// calls to distributeConjuncts().
-  int32_t numCanonicalConjuncts{0};
```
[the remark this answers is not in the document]
MBkkt: I made PR: https://github.com/facebookincubator/axiom/pull/585

axiom/optimizer/ToGraph.cpp
```diff
@@ -179,7 +179,7 @@ ExprCP ToGraph::tryFoldConstant(
     auto typedExpr = queryCtx()->optimization()->toTypedExpr(call);
     auto exprSet = evaluator_.compile(typedExpr);
     const auto& first = *exprSet->exprs().front();
-    if (first.specialFormKind() != velox::exec::SpecialFormKind::kConstant) {
+    if (!first.isConstant()) {
```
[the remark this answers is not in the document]
MBkkt: I made PR: https://github.com/facebookincubator/axiom/pull/584

axiom/optimizer/DerivedTable.cpp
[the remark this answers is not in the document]
MBkkt: I removed these changes from this PR, they're not necessary here.
Lets continue this discussion in a separate PR: https://github.com/facebookincubator/axiom/pull/579

axiom/optimizer/ToGraph.cpp
MBkkt: I did this as you suggested, now we never reassign orderKeys and never clear them.
We just skip Sort node if it's not necessary

axiom/optimizer/ToGraph.cpp
```diff
@@ -2010,70 +1976,54 @@ DerivedTableP ToGraph::makeQueryGraph(const lp::LogicalPlanNode& logicalPlan) {
...
+    if (node.kind() == lp::NodeKind::kSort) {
+      // Sort not allowed doesn't mean we need to wrap it in DT,
+      // instead we should skip it.
+      makeQueryGraph(*node.onlyInput(), allowedInDt);
+    } else {
+      wrapInDt(node, /*unordered=*/false);
+    }
+    return;
+  }
+
```
MBkkt: This is core changes @mbasmanova

MBkkt:
@mbasmanova I rebased this PR

axiom/optimizer/ToGraph.cpp
```diff
@@ -1199,6 +1202,9 @@ AggregationPlanCP ToGraph::translateAggregation(const lp::AggregateNode& agg) {
 }
 
 void ToGraph::addOrderBy(const lp::SortNode& order) {
+  VELOX_DCHECK(currentDt_->orderKeys.empty());
```
mbasmanova: nit: perhaps, !currentDt_->hasOrderBy()

mbasmanova (commented):
@MBkkt Thank you for the optimization. Would you add tests to cover all the cases that are being optimized?

axiom/optimizer/ToGraph.h
```diff
@@ -292,7 +292,7 @@ class ToGraph {
   // DerivedTable. Done for joins to the right of non-inner joins,
   // group bys as non-top operators, whenever descendents of 'node'
   // are not freely reorderable with its parents' descendents.
-  void wrapInDt(const logical_plan::LogicalPlanNode& node);
+  void wrapInDt(const logical_plan::LogicalPlanNode& node, bool unordered);
```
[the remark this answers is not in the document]
MBkkt: I think unordered more clear, dropOrderBy just document the code (what is done in the code) but we can read code instead.
unordered document intention (we want unordered subquery), so I think it's better.

Also unordered common word in C++, unordered map/set/etc, so meaning of this really clear.


Although I can modify comment for this function, to clarify it's additionally

MBkkt:
> Would you clarify what is Ordered and Unordered TopN?

@mbasmanova 

`select x from order by x limit 10` => provide 10 smallest `x` elements in sorted order, this commonly done with topn (it's ordered topn)
`select * from (select * from order by x limit 10) => provide 10 smallest `x` elements in any order, this commonly done with topn but it doesn't do final sort for heap/partition result/tournament tree (it's unordered topn)

This is common optimization that exists in other query engines

PR #550, opened 2025-10-23
refactor: Introduce PlanCost

mbasmanova:
Summary: Cost struct is used to represent a cost of a single RelationOp as well as a cost of a Plan (a tree of RelationOp). This is confusing. Introduce PlanCost to represent the cost of a Plan and limit usage of Cost to representing the cost of a RelationOp.

Differential Revision: D85342911

### DIFF
--- a/axiom/optimizer/Optimization.cpp
+++ b/axiom/optimizer/Optimization.cpp
@@ -95,12 +95,12 @@ PlanAndStats Optimization::toVeloxPlan(
 void Optimization::trace(
     uint32_t event,
     int32_t id,
-    const Cost& cost,
+    const PlanCost& cost,
     RelationOp& plan) const {
   if (event & options_.traceFlags) {
     std::cout << (event == OptimizerOptions::kRetained ? "Retained: "
                                                        : "Abandoned: ")
-              << id << ": " << cost.toString(true, true) << ": " << " "
+              << id << ": " << cost.toString() << ": " << " "
               << plan.toString(true, false) << std::endl;
   }
 }
@@ -195,7 +195,8 @@ void reducingJoinsRecursive(
 // For an inner join, see if can bundle reducing joins on the build.
 std::optional<JoinCandidate> reducingJoins(
     const PlanState& state,
-    const JoinCandidate& candidate) {
+    const JoinCandidate& candidate,
+    bool enableReducingExistences) {
   std::vector<PlanObjectCP> tables;
   std::vector<PlanObjectSet> existences;
   float fanout = candidate.fanout;
@@ -229,7 +230,7 @@ std::optional<JoinCandidate> reducingJoins(
     }
   }
 
-  if (!state.dt->noImportOfExists) {
+  if (enableReducingExistences && !state.dt->noImportOfExists) {
     PlanObjectSet exists;
     float reduction = 1;
     VELOX_DCHECK(!candidate.tables.empty());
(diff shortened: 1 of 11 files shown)


axiom/optimizer/RelationOp.h
```diff
@@ -51,28 +51,39 @@ struct Cost {
...
-  /// Maximum memory occupancy. If the operation is blocking, e.g. group by, the
-  /// amount of spill is 'totalBytes' - 'peakResidentBytes'.
-  float peakResidentBytes{0};
+  float totalCost() const {
+    return unitCost * inputCardinality;
+  }
+
+  float resultCardinality() const {
+    return fanout * inputCardinality;
+  }
```
MBkkt: RelationOp has similar method but it uses raw members.
Can you make it use this method instead

axiom/optimizer/RelationOp.cpp
```diff
@@ -216,11 +215,19 @@ std::string Cost::toString(bool /*detail*/, bool isUnit) const {
...
+  return RelationOpPrinter::toOneline(*this);
+}
+
 void RelationOp::printCost(bool detail, std::stringstream& out) const {
   auto ctx = queryCtx();
   if (ctx && ctx->contextPlan()) {
     auto plan = ctx->contextPlan();
-    auto totalCost = plan->cost.unitCost + plan->cost.setupCost;
+    auto totalCost = plan->cost.cost;
     auto pct = 100 * cost_.inputCardinality * cost_.unitCost / totalCost;
```
MBkkt:
```suggestion
    auto pct = 100 * cost_.totalCost() / totalCost;
```

axiom/optimizer/Plan.cpp
```diff
@@ -120,12 +119,8 @@ std::string Plan::toString(bool detail) const {
...
 
 void PlanState::addCost(RelationOp& op) {
-  VELOX_DCHECK_EQ(cost.inputCardinality, 1);
-  cost.unitCost += cost.fanout * op.cost().unitCost;
-  cost.setupCost += op.cost().setupCost;
-  cost.fanout *= op.cost().fanout;
-  cost.totalBytes += op.cost().totalBytes;
-  cost.transferBytes += op.cost().transferBytes;
+  cost.cost += cost.cardinality * op.cost().unitCost;
+  cost.cardinality *= op.cost().fanout;
```
MBkkt:
```suggestion
  VELOX_DCHECK_EQ(cost.cardinality, op.cost().inputCardinality);
  cost.cost += op.cost().totalCost();
  cost.cardinality = op.cost().resultCardinality();
```

xiaoxmeng (approved):
@mbasmanova thanks!

axiom/optimizer/Optimization.cpp
```diff
@@ -195,7 +195,8 @@ void reducingJoinsRecursive(
 // For an inner join, see if can bundle reducing joins on the build.
 std::optional<JoinCandidate> reducingJoins(
     const PlanState& state,
-    const JoinCandidate& candidate) {
+    const JoinCandidate& candidate,
+    bool enableReducingExistences) {
```
MBkkt: I don't think we need to pass this.

Instead I think better to change derived table ctor to make `noImportOfExists = !enableReducingExistences`

mbasmanova: I thought about this, but DerivedTable.noImportOfExists has a different meaning. It means that existences have been imported already.

PR #553, opened 2025-10-24
fix: Check allowed-in-dt before adding orderBy or limit

mbasmanova:
Differential Revision: D85428042

When translating join's inputs with limit or order by, make sure to wrap these in their own DTs.

### DIFF
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -1827,6 +1827,11 @@ void ToGraph::makeQueryGraph(
       return;
     }
     case lp::NodeKind::kSort: {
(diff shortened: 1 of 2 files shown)


MBkkt:
I don't understand approach :/
Yes we can fix some bugs this way, as I already mentioned in the issue, but the whole pipeline still will be incorrect.
Can you consider my PR please?

MBkkt:
I didn't write my PR just for fun or for reason that I don't like current code and want to rewrite all this. I wrote it because I tried to achieve correct approach. 

To make current approach correct and better we need try to close dt only in join (in the caller), not earlier (in wrapDt now)

xiaoxmeng (approved):
@mbasmanova thanks!

PR #555, opened 2025-10-24
refactor: Rename asUnchecked to as in LogicalPlanNode and Expr

mbasmanova:
Differential Revision: D85430649

### DIFF
--- a/axiom/logical_plan/Expr.cpp
+++ b/axiom/logical_plan/Expr.cpp
@@ -124,7 +124,7 @@ void validateDereferenceInputs(
       inputs[1]->isConstant(),
       "Second input to DEREFERENCE must be a constant");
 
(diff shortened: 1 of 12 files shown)


MBkkt (commented):
instead dynamic_cast in this place should use static_cast, we already discussed this in other PR (some old refactoring), please keep it asUnchecked

Good code avoids dynamic_cast when possible, because it's ineffective

mbasmanova:
@MBkkt I'm renaming it for consistency. Everywhere else we use 'as'.

mbasmanova:
> I think we want as for dynamic cast, and asUnchecked for reinterpret/static casts?

It seems we don't need both. So far, all the use cases are like this: check kind, then cast. So, I think we can standardize on a single "as" method.

MBkkt:
@mbasmanova 

> It seems we don't need both. 

Ok, makes sense to me

PR #556, opened 2025-10-24
feat: Add kindName and toString to LogicalPlanNode and Expr

mbasmanova:
Differential Revision: D85433544


jagill (approved):
Also includes changes from #555

PR #558, opened 2025-10-24
fix: Do not link AxiomSql with Velox QueryBenchmark

MBkkt:
locally I have linking error, because num_drivers/etc flags are duplicated


mbasmanova (approved):
Thanks.

PR #559, opened 2025-10-24
fix: Use static_cast instead of dynamic_cast/reinterpret_cast

MBkkt:
Note about reinterpret_cast:
If reinterpret_cast correct in such place static_cast also will be correct and do same.
But if reinterpret_cast isn't correct (offset base class != offset derived class) it will be correct with static cast/dynamic cast.

I think better to use safe code.


MBkkt:
@mbasmanova added

Issue #530, opened 2025-10-20
Adding reducing semi joins may produce invalid plans


MBkkt:
It somehow related to this condition (don't reproduced if I wrote false here)

https://github.com/facebookincubator/axiom/blob/d650c041321fb801359f359d7d35631f3c7a605a/axiom/optimizer/DerivedTable.cpp#L315

Also if I change `noImportOfExists` to always be true issue also disappear

### How this can be reproduced?

Make `NextJoin::isWorse` always returns false and run tpch.q9:

https://github.com/facebookincubator/axiom/pull/527#issuecomment-3422585718

mbasmanova:
The simplest repro I have so far is the following:

```
select
	l.l_extendedprice, ps.ps_supplycost
from
	part as p,
	supplier as s,
	lineitem as l,
	partsupp as ps,
	orders as o,
	nation as n
where
	s.s_suppkey = l.l_suppkey
	and ps.ps_suppkey = l.l_suppkey
	and ps.ps_partkey = l.l_partkey
	and p.p_partkey = l.l_partkey
	and o.o_orderkey = l.l_orderkey
	and s.s_nationkey = n.n_nationkey
```

It doesn't require any code changes (e.g. the error reproduces w/o changing NextJoin::isWorse).

> Reason: Join's output column not found in either left or right sides: s_suppkey

PR #560, opened 2025-10-24
fix: Import of reducing semi joins in q9

mbasmanova:
Summary:
Partial fix for https://github.com/facebookincubator/axiom/issues/530

Adapted from Orri's https://github.com/facebookincubator/axiom/pull/551

When we import an already placed table or tables into a build side for cardinality reduction as a semijoin:

The colums on the right side of the semijoin must not be added to downstreamColumns. A semijoin (exists) or antijoin (not exists) does not produce columns, except maybe a mark. Also, if the join has filters, the columns in the filter that come from the right side are not downstream columns.

Co-authored-by: oerling

Differential Revision: D85464350

### DIFF
--- a/axiom/optimizer/DerivedTable.cpp
+++ b/axiom/optimizer/DerivedTable.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 #include "axiom/optimizer/DerivedTable.h"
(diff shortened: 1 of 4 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

PR #561, opened 2025-10-25
fix: Make sameOrEqual work properly

pashandor789:
This pull request fixes a logical error in the `Expr::sameOrEqual` method in `QueryGraph.cpp`. The change corrects the condition for comparing arguments in function call expressions.

* Fixed the argument comparison logic in `Expr::sameOrEqual` so that it returns `false` if any argument is not the same or equal, rather than if any argument is the same or equal.


mbasmanova (commented):
@pashandor789 Thank you for the fix. Would you add a test?

MBkkt:
@mbasmanova CI is green now

PR #568, opened 2025-10-28
feat: Add support for non-correlated subqueries used in filters

mbasmanova:
Summary: Add minimal support for non-correlated subqueries used in a filter. Enables TPC-H q15, 16 and 18.

Add logic to process non-correlated subqueries found in a filter.

IN <subquery> expression is handled by creating a new DT for the subquery and adding a semi-join edge to the current DT. The IN predicate is then replaced with a 'mark' column produced by the join.

Other <subquery> expressions are handled by creating a new DT for the subquery and replacing the expression with the only column produced by the DT.

Differential Revision: D85625835

### DIFF
--- a/axiom/optimizer/Subfields.cpp
+++ b/axiom/optimizer/Subfields.cpp
@@ -445,6 +445,12 @@ void ToGraph::markSubfields(
     return;
   }
 
+  if (expr->isSubquery()) {
+    // TODO We may not necessarily need all outputs of the subquery.
+    markAllSubfields(*expr->as<lp::SubqueryExpr>()->subquery());
+    return;
+  }
+
   VELOX_UNREACHABLE("Unhandled expr: {}", lp::ExprPrinter::toText(*expr));
 }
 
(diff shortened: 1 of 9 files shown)


axiom/optimizer/Subfields.cpp
```diff
@@ -445,6 +445,12 @@ void ToGraph::markSubfields(
     return;
   }
 
+  if (expr->isSubquery()) {
+    // TODO We may not necessarily need all outputs of the subquery.
+    markAllSubfields(*expr->as<lp::SubqueryExpr>()->subquery());
+    return;
+  }
```
MBkkt: This should be moved in control subfields (see TableWrite approach).
Here now you do same work for every output column.

axiom/optimizer/ToGraph.cpp
```diff
@@ -1466,9 +1477,35 @@ void ToGraph::addProjection(const lp::ProjectNode& project) {
   }
 }
 
+namespace {
+void extractSubqueries(
+    const lp::ExprPtr& expr,
+    std::vector<lp::SubqueryExprPtr>& subqueries) {
+  if (expr->isSubquery()) {
+    subqueries.push_back(
+        std::dynamic_pointer_cast<const lp::SubqueryExpr>(expr));
```
MBkkt: `std::static_pointer_cast`

PR #575, opened 2025-10-30
feat: Enable comparison with a correlated subquery

mbasmanova:
Summary:
Extend correlated subquery support to handle comparisons with a correlated aggregation subquery.

Enables TPC-H q2, 17  and 20.

Differential Revision: D85881217


xiaoxmeng (approved):
@mbasmanova thanks!

PR #584, opened 2025-11-01
refactor: Don't throw unnecessary exception

MBkkt:
This will simplify debugging and can improve performance.

>     if (first.specialFormKind() != velox::exec::SpecialFormKind::kConstant) {

'first' is velox::exec::Expr. Expr::specialFormKind() API is defined as 

```
  SpecialFormKind specialFormKind() const {
    return specialFormKind_.value();
  }
```

This API throws if Expr is not a special form. To avoid unnecessary exception, we change the code to use Expr::isConstant() API:

```
  bool isConstant() const {
    return specialFormKind_ == SpecialFormKind::kConstant;
  }
```


mbasmanova (approved):
@MBkkt Thank you for clarifying.

PR #590, opened 2025-11-03
perf: Flatten nested unions

MBkkt:
Flatten nested unions, e.g.

```
- Union
  - Union
     - A
     - B
   - C
```
becomes

```
- Union
   - A
   - B
   - C
```

Also, Union(UnionAll) becomes Union:

```
- Union
  - UnionAll
     - A
     - B
   - C
```

becomes

```
- Union
   - A
   - B
   - C
```

All UNION and UNION ALL children of a UNION are flattened: u(u(a, b), ua(c, d)) -> u(a, b, c, d).
All UNION ALL children of UNION ALL are flattened: ua(a, ua(b, c)) -> ua(a, b, c).

This produces simpler plans with fewer exchanges. Instead of N 2-way gathers we create 1 N-way gather.

### DIFF
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -1905,12 +1905,11 @@ void makeUnionDistributionAndStats(DerivedTableP setDt) {
 
 void translateSetOperationInput(
     const lp::LogicalPlanNode& input,
-    const std::function<const lp::LogicalPlanNode*(const lp::LogicalPlanNode&)>&
-        maybeFlatten,
+    const std::function<bool(const lp::LogicalPlanNode&)>& shouldFlatten,
     const std::function<void(const lp::LogicalPlanNode&)>& translateInput) {
-  if (const auto* setNode = maybeFlatten(input)) {
-    for (const auto& child : setNode->inputs()) {
-      translateSetOperationInput(*child, maybeFlatten, translateInput);
+  if (shouldFlatten(input)) {
+    for (const auto& child : input.inputs()) {
+      translateSetOperationInput(*child, shouldFlatten, translateInput);
     }
   } else {
     translateInput(input);
@@ -1923,23 +1922,22 @@ void ToGraph::translateUnion(const lp::SetNode& set) {
   auto* setDt = currentDt_;
   setDt->setOp = set.operation();
 
-  auto maybeFlatten =
-      [&](const lp::LogicalPlanNode& input) -> const lp::LogicalPlanNode* {
+  auto shouldFlatten = [&](const lp::LogicalPlanNode& input) {
     if (input.kind() != lp::NodeKind::kSet) {
-      return nullptr;
+      return false;
     }
-    const auto* inputSet = input.as<lp::SetNode>();
-    const auto inputSetOp = inputSet->operation();
-    if (inputSetOp == setDt->setOp) {
+    const auto inputSetOp = input.as<lp::SetNode>()->operation();
+    const auto parentSetOp = setDt->setOp;
+    if (inputSetOp == parentSetOp) {
       // Same set operation can be flattened.
-      return inputSet;
+      return true;
     }
     if (inputSetOp == lp::SetOperation::kUnionAll &&
-        setDt->setOp == lp::SetOperation::kUnion) {
+        parentSetOp == lp::SetOperation::kUnion) {
       // UNION ALL can be flattened into UNION.
-      return inputSet;
+      return true;
     }
-    return nullptr;
+    return false;
   };
 
   auto renames = std::move(renames_);
@@ -1984,7 +1982,7 @@ void ToGraph::translateUnion(const lp::SetNode& set) {
     setDt->children.push_back(newDt);
   };
 
-  translateSetOperationInput(set, maybeFlatten, translateUnionInput);
+  translateSetOperationInput(set, shouldFlatten, translateUnionInput);
   makeUnionDistributionAndStats(setDt);
 
   renames_ = std::move(renames);
--- a/axiom/optimizer/ToGraph.h
+++ b/axiom/optimizer/ToGraph.h
@@ -278,16 +278,15 @@ class ToGraph {
   // Adds a JoinEdge corresponding to 'join' to the enclosing DerivedTable.
   void translateJoin(const logical_plan::JoinNode& join);
 
-  // Creates derived tables for set operation inputs, put them to current
-  // DerivedTable tables and adds join edges between them to implement INTERSECT
-  // or EXCEPT.
+  // Given an INTERSECT or an EXCEPT set operation, create derived tables for
+  // inputs, add them to 'currentDt_' and connect them with join edges.
   void translateSetJoin(const logical_plan::SetNode& set);
 
-  // Creates derived tables for set operation inputs, put them in current
-  // DerivedTable as a children and set setOp to implement UNION or UNION ALL.
-  // This method recursively flattens
-  // - UNION ALL inside UNION and UNION ALL
-  // - UNION inside UNION.
+  // Given a UNION or UNION ALL set operation, recursively flattens the inputs
+  // and converts these into derived tables.
+  // Flattens UNION ALL inside UNION and UNION ALL.
+  // Flattens UNION inside UNION.
+  // Example: u(a, u(b, c)) -> u(a, b, c)
   void translateUnion(const logical_plan::SetNode& set);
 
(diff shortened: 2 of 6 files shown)


MBkkt:
image
You can use this to simplify review

axiom/optimizer/ToGraph.cpp
```diff
@@ -2186,19 +2182,17 @@ void ToGraph::makeQueryGraph(
...
+        translateUnion(set);
       } else {
-        translateSetJoin(*set, setDt);
+        translateSetJoin(set);
       }
-      currentDt_->addTable(setDt);
-      return;
-    }
+      outerDt->addTable(currentDt_);
+      currentDt_ = outerDt;
```
mbasmanova: Would it be safer / clearer to put this into SCOPE_EXIT right after setting outerDt?

axiom/optimizer/ToGraph.cpp
```diff
@@ -1910,90 +1908,88 @@ void ToGraph::makeUnionDistributionAndStats(
...
+    }
+  } else {
+    currentDt_ = newDt();
+    makeQueryGraph(input, kAllAllowedInDt);
+    auto* newDt = currentDt_;
+    currentDt_ = setDt;
+
+    const auto& type = input.outputType();
+
+    if (isLeftLeaf) {
```
mbasmanova: LeftLeaf is very hard to read. Is this the same as FirstInput? Would it be clearer to rename to isFirstInput?

axiom/optimizer/ToGraph.h
```diff
@@ -278,7 +278,7 @@ class ToGraph {
   // Adds a JoinEdge corresponding to 'join' to the enclosing DerivedTable.
   void translateJoin(const logical_plan::JoinNode& join);
 
-  void translateSetJoin(const logical_plan::SetNode& set, DerivedTableP setDt);
+  void translateSetJoin(const logical_plan::SetNode& set);
```
mbasmanova: Would you document this and other modified methods?

axiom/optimizer/ToGraph.cpp
```diff
@@ -1910,90 +1908,88 @@ void ToGraph::makeUnionDistributionAndStats(
...
-      if (node.kind() == lp::NodeKind::kSet) {
-        const auto* set = node.as<lp::SetNode>();
-        if (set->operation() == lp::SetOperation::kUnion ||
-            set->operation() == lp::SetOperation::kUnionAll) {
-          return set;
-        }
-      }
+  auto* setDt = currentDt_;
 
+  auto maybeFlatten =
```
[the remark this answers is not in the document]
mbasmanova: > additional allocations/move/copy data

I'm thinking that we could add a helper method that takes lp::SetNode and returns `std::vector<lp::LogicalPlanNodePtr>` of inputs. That method would flatten inputs when possible and drop direct lp::SortNodes.

For example, assuming neither a, b, or c are unions,

- flatten(u(a, u(b, c))) returns [a, b, c] 
- flatten(u(sort(a), ua(b, c)) returns [a, b, c]
- flatten(u(u(a, b), u(c, d))) returns [a, b, c, d]

The only allocation and copy is for std::vector of shared pointers. Doesn't feel worth optimizing if the overall logic becomes easier to understand.

mbasmanova:
> This is true only for intersects and except.

I see. And this is not part of this change? Are you planning to make a separate change for that?

> For union joins not changed, instead we just compare 1 N-way gather vs N 2-way gathers

Got it. Please, explain this in the PR description.

axiom/optimizer/ToGraph.cpp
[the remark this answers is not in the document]
MBkkt: Code that you suggest IMO reads and works worse

MBkkt:
@mbasmanova I updated code, I separated
1. tests from PlanTest
2. recursive traversal from actual union input derived table creation

axiom/optimizer/ToGraph.h
```diff
@@ -278,20 +278,17 @@ class ToGraph {
...
-      const logical_plan::SetNode& set,
-      DerivedTableP setDt,
-      bool isTopLevel,
-      bool& isLeftLeaf);
+  // Creates derived tables for set operation inputs, put them to current
+  // DerivedTable tables and adds join edges between them to implement INTERSECT
+  // or EXCEPT.
+  void translateSetJoin(const logical_plan::SetNode& set);
+
+  // Creates derived tables for set operation inputs, put them in current
```
mbasmanova: typos; perhaps,

```
// Given a UNION or UNION ALL set operation, recursively flattens the inputs and converts these into derived tables.
// Flattens UNION ALL inside UNION and UNION ALL. 
// Flattend UNION inside UNION.
// Example: u(a, u(b, c)) -> u(a, b, c)
```

axiom/optimizer/tests/SetTest.cpp
```diff
@@ -0,0 +1,478 @@
...
+    test::QueryTestBase::TearDownTestCase();
+  }
+
+ private:
+  inline static std::shared_ptr<velox::exec::test::TempDirectoryPath>
+      gTempDirectory;
+};
+
+TEST_F(SetTest, unionAll) {
+  auto nationType =
```
mbasmanova: Can we use HiveQueriesTestBase::getSchema method here?

axiom/optimizer/tests/SetTest.cpp
```diff
@@ -0,0 +1,478 @@
...
+            .build();
+
+    auto plan = toSingleNodePlan(logicalPlan);
+
+    if (rootType == lp::SetOperation::kUnion) {
+      auto matcher =
+          core::PlanMatcherBuilder()
+              .tableScan()
+              .localPartition({
+                  core::PlanMatcherBuilder().tableScan().project().build(),
```
mbasmanova: > core::PlanMatcherBuilder().tableScan().project().build(),

perhaps, introduce a local variable to avoid lengthy copy-paste?

axiom/optimizer/tests/SetTest.cpp
```diff
@@ -0,0 +1,478 @@
...
+                      .build())
+              .build();
+
+      AXIOM_ASSERT_PLAN(plan, matcher);
+    }
+
+    auto referencePlan =
+        exec::test::PlanBuilder(pool_.get())
+            .tableScan("nation", nationType)
+            .filter("n_nationkey < 11 or n_nationkey > 13")
```
mbasmanova: Does this test expose the limitation of the optimizer? It should have produced a plan like this without any unions, but it doesn't.

mbasmanova (commented):
@MBkkt Thank you for iterating. Overall looks good. A few comments.

axiom/optimizer/ToGraph.cpp
```diff
@@ -1878,122 +1877,120 @@ void ToGraph::translateSetJoin(const lp::SetNode& set, DerivedTableP setDt) {
...
-            set->operation() == lp::SetOperation::kUnionAll) {
-          return set;
-        }
-      }
+void ToGraph::translateUnion(const lp::SetNode& set) {
+  auto* setDt = currentDt_;
+  setDt->setOp = set.operation();
 
+  auto maybeFlatten =
+      [&](const lp::LogicalPlanNode& input) -> const lp::LogicalPlanNode* {
```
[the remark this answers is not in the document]
MBkkt: I don't want to pass `setDt->setOp` recursively, so I think lambda is better.

About boolean and name I agree and will change it

axiom/optimizer/tests/SetTest.cpp
```diff
@@ -0,0 +1,478 @@
...
+using namespace velox;
+namespace lp = facebook::axiom::logical_plan;
+
+class SetTest : public test::QueryTestBase {
+ public:
+  static void SetUpTestCase() {
+    test::QueryTestBase::SetUpTestCase();
+
+    std::string path;
+    if (FLAGS_data_path.empty()) {
```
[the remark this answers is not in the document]
MBkkt: Yes, fixed

axiom/optimizer/tests/SetTest.cpp
MBkkt: Ok, changed

axiom/optimizer/ToGraph.cpp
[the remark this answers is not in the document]
MBkkt: I changed name (to shouldFlatten) + make it return bool

PR #592, opened 2025-11-03
refactor: Split CLI logic into Console and SqlQueryRunner

mbasmanova:
Summary: Console class is responsible for interacting with the user and printing query results. SqlQueryRunner is reponsible for running SQL queries.

Differential Revision: D86141222


xiaoxmeng (approved):
@mbasmanova thanks!

Issue #557, opened 2025-10-24
Filter isn't pushed down for the join with limits on both sides


pashandor789:
```
TEST_F(PlanTest, joinWithFilterOverLimit) {
  testConnector_->addTable("t", ROW({"a", "b", "c"}, BIGINT()));
  testConnector_->addTable("u", ROW({"x", "y", "z"}, BIGINT()));

  lp::PlanBuilder::Context ctx(kTestConnectorId);
  auto logicalPlan =
      lp::PlanBuilder(ctx)
          .tableScan("t")
          .limit(100)
          .filter("b > 50")
          .join(
              lp::PlanBuilder(ctx).tableScan("u").limit(50).filter("y < 100"),
              "a = x",
              lp::JoinType::kInner)
          .build();

  {
    auto plan = toSingleNodePlan(logicalPlan);
    auto matcher = core::PlanMatcherBuilder()
                       .tableScan("t")
                       .limit()
                       .filter("b > 50")
                       .hashJoin(core::PlanMatcherBuilder()
                                     .tableScan("u")
                                     .limit()
                                     .filter("y < 100")
                                     .build())
                       .build();

    AXIOM_ASSERT_PLAN(plan, matcher);
  }
}
```

doesn't produce an effective plan 

```
-- Filter[0][expression: lt("y",100)] -> a:BIGINT, b:BIGINT, c:BIGINT, x:BIGINT, y:BIGINT, z:BIGINT
  -- HashJoin[6][INNER a=x] -> a:BIGINT, b:BIGINT, c:BIGINT, x:BIGINT, y:BIGINT, z:BIGINT
    -- Filter[1][expression: gt("b",50)] -> a:BIGINT, b:BIGINT, c:BIGINT
      -- Limit[3][100] -> a:BIGINT, b:BIGINT, c:BIGINT
        -- TableScan[2][t] -> a:BIGINT, b:BIGINT, c:BIGINT
    -- Limit[5][50] -> x:BIGINT, y:BIGINT, z:BIGINT
      -- TableScan[4][u] -> x:BIGINT, y:BIGINT, z:BIGINT
```

```
dt1: a, b, c, x, y, z
  output:
    a := dt2.a
    b := dt2.b
    c := dt2.c
    x := dt4.x
    y := dt4.y
    z := dt4.z
  tables: dt2, dt4
  joins:
    dt2 INNER dt4 ON dt2.a = dt4.x
  syntactic join order: 1, 14
[...]

Issue #536, opened 2025-10-21
finalizeDt on both sides of the join makes redundant enumeration


pashandor789:
# Example

```
auto logicalPlan =
      lp::PlanBuilder(context)
          .tableScan("nation")
          .limit(0, 10)
          .orderBy({"n_nationkey"})
          .as("n1")
          .join(
              lp::PlanBuilder(context)
                  .tableScan("region")
                  .limit(0, 5)
                  .orderBy({"r_regionkey"})
                  .as("r1"),
              "n1.n_regionkey = r1.r_regionkey",
              lp::JoinType::kInner)
          .build();
```

This queries cause a fail below.

```
E20251021 17:04:08.954496 902249 Exceptions.h:53] Line: /home/ivanovp/projects/verax/axiom/optimizer/Optimization.cpp:1297, Function:crossJoin, Expression:  No cross joins, Source: RUNTIME, ErrorCode: NOT_IMPLEMENTED
unknown file: Failure
C++ exception with description "Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: NOT_IMPLEMENTED
Reason: No cross joins
```

# Why is it happening?

translateJoin calls makeQueryGraph(*joinLeft) / makeQueryGraph(*joinRight)

consider we've dt1 when we are entering the translateJoin.
on the left makeQueryGraph we have limit + sort as in example, there's a code

```
case lp::NodeKind::kSort:
      // Multiple orderBys are allowed before a limit. Last one wins. Previous
      // are dropped. If arrives after limit, then starts a new DT.

      makeQueryGraph(*node.onlyInput(), allowedInDt);

      if (currentDt_->hasLimit()) {
        finalizeDt(*node.onlyInput());
      }

      return addOrderBy(*node.asUnchecked<lp::SortNode>());

```

so we makeQueryGraph for the input and now currentDt (let it be dt1) has a state:
```
dt1.tables = [table1]
```

then we call finalizeDt (code is below), which creates a new one and then put the previous dt into tables of the current
so now the state will be:
```
dt1.tables = [table1]
dt2.tables = [dt1]
```

```
void ToGraph::finalizeDt(
    const lp::LogicalPlanNode& node,
    DerivedTableP outerDt) {
  DerivedTableP dt = currentDt_;
  setDtUsedOutput(dt, node);

  currentDt_ = outerDt != nullptr ? outerDt : newDt();
  currentDt_->addTable(dt);

  dt->makeInitialPlan();
}
```

[...]

PR #599, opened 2025-11-06
fix: Translation of Joins into query graph

MBkkt:
## First commit

Fix translation of Joins to DTs. Join(Filter(A), Filter(B)) used to be translated into a single DT, but now it becomes 3 DTs:
 
1. DT for Filter(A)
2. DT for Filter(B)
3. DT with the join. 

Extra wrapping fixes cases where A or B are DTs themselves (see example in #557). The cases where A and B are BaseTables end up with a single DT like before thanks to pre-existing flattenDt logic. In these cases, we first produce 3 DTs, then flatten these into one.

When translating a join, left side is wrapped in a new DT if it contains any of Unnest, Aggregate, Limit, Filter, Sort. Same applies to the right side. For non-inner joins, right side is additionally wrapped in a new DT if it contains another Join.

There was important part of how joins are handled (in case when left part of join tree contained non-deterministic filter).
This code was simplified just to `separate DT for filter input + filter`

Part of https://github.com/facebookincubator/axiom/issues/536 (dropping unnecessary order by is still TBD)
Fixes https://github.com/facebookincubator/axiom/issues/557

## Second commit

Fixes https://github.com/facebookincubator/axiom/issues/606 

This cannot be separated, because it depends on how joins translated into graph (first commit).

The general idea is simple.
Lets make following assumption (this requires to change code in ToGraph).
1. if we have left join (right rows are optional), then right table is not null in JoinEdge 
2. if we have right join (left rows are optional), then left table is not null in JoinEdge 
3. if we have outer join (both, right and left rows are optional), then both, right and left tables are not null in JoinEdge

If these assumptions are true we can modify `distributeConjuncts` and `placeConjuncts` to don't push-down filter that depends on optional rows (because we know tables in which pushdown isn't possible).

I also noticed because asymmetric structure of current join enumeration, we cannot allow rightTable to be null in case of right join, so if join order isn't syntactic we need to change right join to left, to allow us to produce more optimal plans

## Co-authored-by

Co-authored-by: oerling
Co-authored-by: mbasmanova

### DIFF
--- a/axiom/optimizer/Optimization.cpp
+++ b/axiom/optimizer/Optimization.cpp
@@ -1490,7 +1490,7 @@ void Optimization::placeDerivedTable(DerivedTableCP from, PlanState& state) {
 bool Optimization::placeConjuncts(
     RelationOpPtr plan,
     PlanState& state,
-    bool allowNondeterministic) {
+    bool joinsPlaced) {
   PlanStateSaver save(state);
 
   PlanObjectSet columnsAndSingles = state.columns;
@@ -1498,7 +1498,7 @@ bool Optimization::placeConjuncts(
       [&](auto dt) { columnsAndSingles.unionObjects(dt->columns); });
 
   PlanObjectSet noPushdownTables;
-  if (!allowNondeterministic) {
+  if (!joinsPlaced) {
     for (const auto* join : state.dt->joins) {
       if (join->leftOptional()) {
         // No pushdown to the left side of a RIGHT or FULL join.
@@ -1512,16 +1512,15 @@ bool Optimization::placeConjuncts(
   }
   ExprVector filters;
   for (auto& conjunct : state.dt->conjuncts) {
-    if (!allowNondeterministic && conjunct->containsNonDeterministic()) {
+    if (!joinsPlaced && conjunct->containsNonDeterministic()) {
       continue;
     }
     if (state.placed.contains(conjunct)) {
       continue;
     }
-    if (!allowNondeterministic) {
-      const auto allTables = conjunct->allTables();
-      if (allTables.size() == 1 &&
-          noPushdownTables.contains(allTables.toObjects()[0])) {
+    if (!joinsPlaced) {
+      const auto* singleTable = conjunct->singleTable();
+      if (singleTable && noPushdownTables.contains(singleTable)) {
         continue;
       }
     }
--- a/axiom/optimizer/Optimization.h
+++ b/axiom/optimizer/Optimization.h
@@ -245,7 +245,7 @@ class Optimization {
   bool placeConjuncts(
       RelationOpPtr plan,
       PlanState& state,
-      bool allowNondeterministic);
+      bool joinsPlaced);
 
   // Helper function that calls makeJoins recursively for each of
   // 'nextJoins'. The point of making 'nextJoins' first and only then
--- a/axiom/optimizer/QueryGraph.h
+++ b/axiom/optimizer/QueryGraph.h
@@ -531,11 +531,10 @@ class JoinEdge {
     VELOX_DCHECK_NOT_NULL(rightTable_);
     // filter_ is only for non-inner joins.
     VELOX_DCHECK(filter_.empty() || !isInner());
+    // Cannot be both semi and anti join.
     VELOX_DCHECK(!rightExists_ || !rightNotExists_);
-
-    if (markColumn_) {
-      VELOX_DCHECK(rightExists_);
-    }
+    // Mark column only for semi joins.
+    VELOX_DCHECK(!markColumn_ || rightExists_);
   }
 
   static JoinEdge* makeInner(PlanObjectCP leftTable, PlanObjectCP rightTable) {


axiom/optimizer/tests/HiveQueriesTest.cpp
```diff
@@ -179,5 +179,116 @@ TEST_F(HiveQueriesTest, orderOfOperations) {
...
+              lp::PlanBuilder(context).tableScan("region").limit(0, 5).orderBy(
+                  {"r_regionkey"}),
+              "n_regionkey = r_regionkey",
+              lp::JoinType::kInner)
+          .build();
+
+  {
+    auto plan = toSingleNodePlan(logicalPlan);
+    // TODO: We don't need orderBy before the join.
+    // This will be fixed in the next PR.
```
MBkkt: This will be fixed in the next PR

axiom/optimizer/ToGraph.cpp
```diff
@@ -2139,47 +2079,38 @@ void ToGraph::makeQueryGraph(
...
     case lp::NodeKind::kJoin: {
-      if (!contains(allowedInDt, lp::NodeKind::kJoin)) {
-        wrapInDt(node);
-        return;
+      const auto& join = *node.as<lp::JoinNode>();
+      const auto& left = *join.left();
+      const auto& right = *join.right();
+      // TODO Allow mixing Unnest with Join in a single DT.
+      // https://github.com/facebookincubator/axiom/issues/286
+      allowedInDt = makeDtIf(allowedInDt, lp::NodeKind::kUnnest);
```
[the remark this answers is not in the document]
mbasmanova: Let's change to 'deny' with variable number of arguments. Thanks.

axiom/optimizer/ToGraph.h
```diff
@@ -289,9 +289,7 @@ class ToGraph {
   // Example: u(a, u(b, c)) -> u(a, b, c)
   void translateUnion(const logical_plan::SetNode& set);
 
-  void translateUnnest(
-      const logical_plan::UnnestNode& logicalUnnest,
-      bool isNewDt);
+  void addUnnest(const logical_plan::UnnestNode& unnest);
```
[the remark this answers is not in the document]
mbasmanova: Is the idea that 'addXxx' simply extends currentDt_, but never creates new DTs? If so, then this naming convention makes sense to me. Thank you for clarifying.

axiom/optimizer/tests/tpch.plans/q15.plans
```diff
@@ -14,7 +14,7 @@ dt1: s_suppkey, s_name, s_address, s_phone, total_revenue
   joins:
     t2 INNER dt3 ON t2.s_suppkey = dt3.l_suppkey
     dt3 INNER dt7 ON dt3.total_revenue = dt7.max
-  syntactic join order: 8, 13, 77
+  syntactic join order: 8, 13, 76
```
mbasmanova: Is this change related to this PR?

axiom/optimizer/tests/ToGraphTest.cpp
```diff
@@ -0,0 +1,215 @@
...
+#include "axiom/optimizer/tests/QueryTestBase.h"
+#include "velox/exec/tests/utils/PlanBuilder.h"
+
+namespace facebook::axiom::optimizer::test {
+namespace {
+
+using namespace velox;
+namespace lp = facebook::axiom::logical_plan;
+
+class ToGraphTest : public QueryTestBase {
```
mbasmanova: > ToGraphTest

This name is too generic. We cannot possibly have a single file for all tests for to-graph logic. Since this change is about joins, let's add tests to Join-specific test file. Perhaps, HiveJoinQueriesTest to match HiveAggregation/LimitQueriesTest. Or, JoinTest to match SetTest.

axiom/optimizer/ToGraph.h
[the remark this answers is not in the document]
MBkkt: > Is the idea that 'addXxx' simply extends currentDt_, but never creates new DTs?

Not exactly, I think they can finalize current dt (like addFilter if limit or (aggregation and subquery))

I think main point of such naming they "add something in current graph", translateAggregation is opposite, it doesn't add aggregation to current graph, instead it's allocate something and return it as return value.

axiom/optimizer/tests/TpchPlanTest.cpp
```diff
@@ -566,9 +567,9 @@ TEST_F(TpchPlanTest, q22) {
 }
 
 // Use to re-generate the plans stored in tpch.plans directory.
-TEST_F(TpchPlanTest, DISABLED_makePlans) {
+TEST_F(TpchPlanTest, makePlans) {
```
mbasmanova: Please, revert these changes.

mbasmanova (commented):
@MBkkt Looks great % changes to TpchPlanTest.cpp. Please, revert these. If some of these are needed, please, submit a separate PR.

mbasmanova:
CI is red.

mbasmanova (approved):
Thanks.

mbasmanova:
@MBkkt I think you are right. We need to set WORKING_DIRECTORY for axiom_optimizer_tpch_plan_test target. Would you like to make a PR?

CC: @assignUser @kgpai

MBkkt:
@mbasmanova 

>  We need to set WORKING_DIRECTORY for axiom_optimizer_tpch_plan_test target.

target doesn't have WORKING_DIRECTORY

To fix this we need to change code, but I'm not sure about good solution, mainly because this part

```cpp
    auto path =
        velox::test::getDataFilePath("", "axiom/optimizer/tests/" + filePath);
```

something like should be

```
auto path = parent(std::source_location::file_name) + filePath
```

> Would you like to make a PR?

If it won't be fixed I will make PR on weekends, because always having this diff locally is quite inconvenient

MBkkt:
@mbasmanova I made PR for this issue with WORKING DIRECTORY, see https://github.com/facebookincubator/axiom/pull/600

peterenescu:
@MBkkt Internal tests are failing for IN subqueries. Currently looking into these test failures (just an fyi for why your import is on hold)
cc: @mbasmanova

MBkkt:
@mbasmanova Why does left join presented in the logical plan?

axiom/optimizer/Optimization.cpp
```diff
@@ -1498,6 +1497,19 @@ bool Optimization::placeConjuncts(
   state.dt->singleRowDts.forEach<DerivedTable>(
       [&](auto dt) { columnsAndSingles.unionObjects(dt->columns); });
 
+  PlanObjectSet noPushdownTables;
+  if (!allowNondeterministic) {
```
mbasmanova: I see that the logic for disabling pushdown is repeated twice. Here and in DerivedTable::distributeConjuncts.  Here, this logic is conditioned on '!allowNondeterministic', which is strange. Why would deterministic nature of the filter have anything to do with outer joins.

If we do not fix the dependencies, we'll have to continue adding one-off logic like this in other places. I suggest we agree to disagree and proceed to land PRs #621 and original version of #599.

MBkkt: There are the two places:
1. rewrite: pushdown filter expression
2. rule: place conjunct

> Why would deterministic nature of the filter have anything to do with outer joins.

This bool named bad, actually it's before (false) and after join (true)


> If we do not fix the dependencies, we'll have to continue adding one-off logic like this in other places

Would you explain where? I don't see, I think this issue is isolated to filter expressions "pushdown"

MBkkt: We have filter expression it can be pushdown-ed:
1) in query graph, without cost (that's why I called this rewrite), this named distributeConjuncts. In this place I checked can or can't we pushdown expression.
2) in plane enumeration, with cost (that's why I called this rule), this named placeConjuncts. There's two places where placeConjuncts called, one before join, one after. After join is safe, it's same as add postprocess, in other words, no pushdown at all. Before join we need check can we or can't we do this, I also checked this.

There's no other places that pushdown filter expressions, and I doubt it make sense to add some other pushdown mechanism instead of extend this.

MBkkt: I also renamed `allowNondeterministic` to `joinsPlaced` and optimize code a little

PR #600, opened 2025-11-06
fix: Allow to run tpch plan test from any directory

MBkkt:
Use source directory to find path to tpch query directories instead of rely on directory where tests are run

It's better because now binary can be executed from any directory.


MBkkt:
@mbasmanova These changes are trivial. 
Would it be possible to you to run these in fb internal workflow and merge if it passes?

mbasmanova:
I'm not sure I understand why we need these changes. The original logic uses the pattern we adopted in Velox and it seems to work well. What is not working and how are these changes fixing that?

kgpai:
I dont see any major concerns with this, looks good save the question on enabling the disabled test.

MBkkt:
@kgpai @mbasmanova Could you merge this? Or do you see some issue with this?

MBkkt:
@mbasmanova I rarely run Velox tests locally, maybe with them similar issue exists.

If it's true, I will fix this for Velox too, but later

kgpai:
@MBkkt So unfortunately this doesnt work internally for us. The root of the project is fbcode for us , so the path returned by __FILE__ parent_path is like so `"fbcode/axiom/optimizer/tests"` but the directory from which automated tests run is '.../fbcode/' , thus it will look for 'fbcode' in 'fbcode'. You have to account for fbcode some way or other as seen here: https://github.com/facebookincubator/velox/blob/c746114662b3ecf55eda5b7dd2e95155509f98fe/velox/dwio/common/tests/utils/DataFiles.cpp#L27 
You could probably make changes there where it uses some other logic for filepath but still preserves fbcode notion ?

MBkkt:
@kgpai Interesting, ok, I will adjust this for fbcode

Issue #606, opened 2025-11-08
Non-inner joins + single column filters don't work


MBkkt:
There 2 reasons:
1. When we do `distributeConjuncts` we don't account that we can pushdown such filters through such joins
2. Same when we actually enumerate plans in `makeJoins`

```cpp
TEST_F(PlanTest, pushdownFilterThroughJoin) {
  testConnector_->addTable("t1", ROW({"id1", "data1"}, {BIGINT(), BIGINT()}));
  testConnector_->addTable("t2", ROW({"id2", "data2"}, {BIGINT(), BIGINT()}));

  auto makePlan = [&](lp::JoinType joinType) {
    lp::PlanBuilder::Context ctx{kTestConnectorId};
    return lp::PlanBuilder{ctx}
        .from({"t1"})
        .join(lp::PlanBuilder{ctx}.from({"t2"}), "id1 = id2", joinType)
        .filter("data1 IS NULL")
        .filter("data2 IS NULL")
        .build();
  };

  {
    SCOPED_TRACE("Inner Join");
    auto logicalPlan = makePlan(lp::JoinType::kInner);
    auto matcher = core::PlanMatcherBuilder{}
                       .tableScan("t1")
                       .filter("data1 IS NULL")
                       .hashJoin(
                           core::PlanMatcherBuilder{}
                               .tableScan("t2")
                               .filter("data2 IS NULL")
                               .build(),
                           core::JoinType::kInner)
                       .build();
    auto plan = toSingleNodePlan(logicalPlan);
    AXIOM_ASSERT_PLAN(plan, matcher);
  }
  {
    SCOPED_TRACE("Left Join");
    auto logicalPlan = makePlan(lp::JoinType::kLeft);
    auto matcher = core::PlanMatcherBuilder{}
                       .tableScan("t1")
                       .filter("data1 IS NULL")
                       .hashJoin(
                           core::PlanMatcherBuilder{}.tableScan("t2").build(),
                           core::JoinType::kLeft)
                       .filter("data2 IS NULL")
                       .build();
    auto plan = toSingleNodePlan(logicalPlan);
    AXIOM_ASSERT_PLAN(plan, matcher);
  }
  {
    SCOPED_TRACE("Right Join");
    auto logicalPlan = makePlan(lp::JoinType::kRight);
    auto matcher = core::PlanMatcherBuilder{}
                       .tableScan("t1")
                       .hashJoin(
                           core::PlanMatcherBuilder{}
[...]

PR #621, opened 2025-11-12
fix: Filter pushdown through the outer join

mbasmanova:
Summary:
ToGraph converts logical plan tree into a Query Graph. The conversion involves creating graph objects such as tables, columns, expressions. New columns are created from nodes such as Values, TableScan, Project, Aggregate. Filter, Sort, Limit do not make new columns. Available columns are stored in renames_ map keyed on the symbol name found in the logical plan.

Each column indicates where it is coming from: a BaseTable, a ValuesTable or a DerivedTable. Columns produced by TableScan nodes are associated with the corresponding BaseTable. Columns produced by the Aggregation are associated with the containing DerivedTable. A filter expression that depends solely on columns from a single BaseTable can be pushed down into that table.

Before this change, the Join node didn’t make new columns. This is fine for INNER join, but is incorrect for outer joins (left, right and full). Outer joins need to make new columns for the ‘optional’ side of the join. LEFT join makes new columns for the right side. RIGHT join makes new columns for the left side. FULL join makes new columns for both sides.

When a join doesn’t make a new column, the filter on top of that join may get incorrectly pushed down below the join. Consider an example,

```
SELECT * FROM (
  SELECT * FROM t LEFT JOIN u ON t.id = u.id
) 
WHERE u.data is null
```

This query has a join followed by a filter that depends only on columns from the right side of the join. However, this filter cannot be pushed down under the join because the join changes the column, by adding NULL rows.

This change fixes plans for queries with filters over outer joins.

This issue was uncovered by changes in #599.

Fixes https://github.com/facebookincubator/axiom/issues/606

Co-authored-by: MBkkt

Differential Revision: D86776319

### DIFF
--- a/axiom/optimizer/Optimization.cpp
+++ b/axiom/optimizer/Optimization.cpp
@@ -1017,6 +1017,53 @@ void Optimization::joinByIndex(
   }
 }
 
+struct ProjectionBuilder {
+  ColumnVector columns;
+  ExprVector exprs;
+
+  // Project 'expr' as 'column'.
+  void add(ColumnCP column, ExprCP expr) {
+    columns.emplace_back(column);
+    exprs.emplace_back(expr);
+  }
+
+  RelationOp* build(RelationOp* input) {
+    return make<Project>(
+        input, exprs, columns, isRedundantProject(input, exprs, columns));
+  }
+
+  ColumnVector inputColumns() const {
+    ColumnVector columns;
+    columns.reserve(exprs.size());
+    for (const auto* expr : exprs) {
+      VELOX_DCHECK(expr->isColumn());
+      columns.emplace_back(expr->as<Column>());
+    }
+
+    return columns;
+  }
+
+  PlanObjectSet outputColumns() const {
+    PlanObjectSet columnSet;
+    columnSet.unionObjects(columns);
+    return columnSet;
+  }
+};
+
+folly::F14FastMap<PlanObjectCP, ExprCP> makeJoinColumnMapping(
+    JoinEdgeP joinEdge) {
+  folly::F14FastMap<PlanObjectCP, ExprCP> mapping;
+  for (auto i = 0; i < joinEdge->leftColumns().size(); ++i) {
+    mapping.emplace(joinEdge->leftColumns()[i], joinEdge->leftExprs()[i]);
+  }
+
+  for (auto i = 0; i < joinEdge->rightColumns().size(); ++i) {
+    mapping.emplace(joinEdge->rightColumns()[i], joinEdge->rightExprs()[i]);
+  }
+
+  return mapping;
+}
+
 void Optimization::joinByHash(
     const RelationOpPtr& plan,
     const JoinCandidate& candidate,
@@ -1125,14 +1172,29 @@ void Optimization::joinByHash(
   PlanObjectSet probeColumns;
   probeColumns.unionObjects(plan->columns());
 
-  ColumnVector columns;
-  PlanObjectSet columnSet;
   ColumnCP mark = nullptr;
+
+  auto* joinEdge = candidate.join;
+
+  PlanObjectSet joinColumns;
+  joinColumns.unionObjects(joinEdge->leftColumns());
+  joinColumns.unionObjects(joinEdge->rightColumns());
+
+  // Mapping from join output column to probe or build side input.
+  auto joinColumnMapping = makeJoinColumnMapping(joinEdge);
+
+  ProjectionBuilder projectionBuilder;
+  bool needsProjection = false;
 
   state.downstreamColumns().forEach<Column>([&](auto column) {
     if (column == build.markColumn) {
       mark = column;
-      columnSet.add(column);
+      return;
+    }
+
+    if (joinColumns.contains(column)) {
+      projectionBuilder.add(column, joinColumnMapping.at(column));
+      needsProjection = true;
       return;
     }
 
@@ -1141,24 +1203,24 @@ void Optimization::joinByHash(
       return;
     }
 
-    columnSet.add(column);
-    columns.push_back(column);
+    projectionBuilder.add(column, column);
   });
 
   // If there is an existence flag, it is the rightmost result column.
   if (mark) {
     const_cast<Value*>(&mark->value())->trueFraction =
         std::min<float>(1, candidate.fanout);
-    columns.push_back(mark);
+    projectionBuilder.add(mark, mark);
   }
-  state.columns = columnSet;
+
+  state.columns = projectionBuilder.outputColumns();
   const auto fanout = fanoutJoinTypeLimit(joinType, candidate.fanout);
 
   PrecomputeProjection precomputeProbe(probeInput, state.dt);
   auto probeKeys = precomputeProbe.toColumns(probe.keys);
   probeInput = std::move(precomputeProbe).maybeProject();
 
-  auto* join = make<Join>(
+  RelationOp* join = make<Join>(
       JoinMethod::kHash,
       joinType,
       probeInput,
@@ -1167,10 +1229,14 @@ void Optimization::joinByHash(
       std::move(buildKeys),
       candidate.join->filter(),
       fanout,
-      std::move(columns));
+      projectionBuilder.inputColumns());
 
   state.addCost(*join);
   state.cost.cost += buildState.cost.cost;
+
+  if (needsProjection) {
+    join = projectionBuilder.build(join);
+  }
 
   state.addNextJoin(&candidate, join, {buildOp}, toTry);
 }
@@ -1189,6 +1255,8 @@ void Optimization::joinByHashRight(
   probeFilterColumns.unionColumns(candidate.join->filter());
   probeFilterColumns.intersect(availableColumns(candidate.tables[0]));
 
+  const auto& dc = state.downstreamColumns();
+
   PlanObjectSet probeTables;
   PlanObjectSet probeColumns;
   for (auto probeTable : candidate.tables) {
@@ -1197,7 +1265,7 @@ void Optimization::joinByHashRight(
     probeTables.add(probeTable);
   }
 
-  probeColumns.intersect(state.downstreamColumns());
+  probeColumns.intersect(dc);
   probeColumns.unionColumns(probe.keys);
   probeColumns.unionSet(probeFilterColumns);
   state.columns.unionSet(probeColumns);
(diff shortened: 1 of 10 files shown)


axiom/optimizer/tests/JoinTest.cpp
```diff
@@ -0,0 +1,137 @@
...
+  {
+    SCOPED_TRACE("Right Join");
+
+    // Force RIGHT join. By default, the optimizer prefers LEFT join.
+    const auto originalSyntacticJoinOrder =
+        optimizerOptions_.syntacticJoinOrder;
+    optimizerOptions_.syntacticJoinOrder = true;
+    SCOPE_EXIT {
+      optimizerOptions_.syntacticJoinOrder = originalSyntacticJoinOrder;
+    };
```
MBkkt: This is true for my code, but probably not for your.
I can be wrong about this, but better to check.
You can implement this or remove this if I'm right about this.

axiom/optimizer/ToGraph.cpp
```diff
@@ -1247,6 +1247,29 @@ void extractNonInnerJoinEqualities(
...
 } // namespace
 
+void ToGraph::addJoinColumns(
+    const logical_plan::LogicalPlanNode& joinSide,
+    ColumnVector& columns,
+    ExprVector& exprs) {
+  const auto channels = usedChannels(joinSide);
+  const auto& names = joinSide.outputType()->names();
+  for (auto i = 0; i < channels.size(); ++i) {
+    const auto& name = names[channels[i]];
```
MBkkt: Please use range based for when possible

```suggestion
  const auto& names = joinSide.outputType()->names();
  for (auto channel : usedChannels(joinSide)) {
    const auto& name = names[channel];
```

axiom/optimizer/ToGraph.cpp
```diff
@@ -1302,14 +1325,33 @@ void ToGraph::translateJoin(const lp::JoinNode& join) {
...
+
+    ColumnVector rightColumns;
+    ExprVector rightExprs;
+    if (rightOptional) {
+      addJoinColumns(*join.right(), rightColumns, rightExprs);
+    }
 
     auto* edge = make<JoinEdge>(
-        leftTableVector.size() == 1 ? leftTableVector[0] : nullptr,
+        leftTable,
```
MBkkt: With your current code leftTable is never null, I doubt this is correct

axiom/optimizer/ToGraph.cpp
```diff
@@ -1302,14 +1325,33 @@ void ToGraph::translateJoin(const lp::JoinNode& join) {
...
         rightTable,
         JoinEdge::Spec{
             .filter = std::move(conjuncts),
             .leftOptional = leftOptional,
-            .rightOptional = rightOptional});
+            .leftColumns = leftColumns,
+            .leftExprs = leftExprs,
+            .rightOptional = rightOptional,
+            .rightColumns = rightColumns,
+            .rightExprs = rightExprs,
```
MBkkt: Please use `std::move` for leftColumns, leftExprs, rightColumns, rightExprs

axiom/optimizer/QueryGraph.h
```diff
@@ -467,15 +469,47 @@ class JoinEdge {
...
+
+    /// True for EXISTS subquery. Mutually exclusive with 'rightNotExists'.
     bool rightExists{false};
+
+    /// True for NOT EXISTS subquery. Mutually exclusive with 'rightExists'.
     bool rightNotExists{false};
+
+    /// Marker column produced by 'exists' or 'not exists' join. Requires
+    /// 'rightExists' or 'rightNotExists' to be true.
     ColumnCP markColumn{nullptr};
```
MBkkt: Comment sounds good to me, but it's opposite to what is written in the code :(

https://github.com/facebookincubator/axiom/blob/fa2b483ac91c271d1c0bddf93c22098ed16ac51c/axiom/optimizer/ToGraph.cpp#L1748

If you want such comments, please write debug asserts for them in ctor of JoinEdge, it's not really difficult but will help a lot.

axiom/optimizer/QueryGraph.h
```diff
@@ -443,13 +443,15 @@ struct JoinSide {
...
+      return velox::core::JoinType::kFull;
+    }
     if (isOptional) {
       return velox::core::JoinType::kLeft;
     }
     if (isNonOptionalOfOuter) {
       return velox::core::JoinType::kRight;
     }
-
     if (markColumn) {
```
MBkkt: As you can see markColumn can happens only when both isExists and isNotExists are false

MBkkt (changes requested):
I like to see that you're working on this, but I'm not sure it's a good direction.

There're multiple different issues, but what's more important overall fix idea doesn't look good to me.

Firstly because this solution introduce additional columns that can produce additional unnecessary projections in some cases.

Another issue with this solution that it requires to change code in every join method that we want to implement:
1. hash join
2. hash join right
3. merge join?
4. index join?
5. etc

This just doesn't scale.


The issue that we pushdown some filter, lets just not pushdown if it's not valid.
Why do we actually want this logic with join producing some new columns?

Also this seems to be a lot of more ineffective compared to table solution.
Because it's required to create columns and vectors of columns and expressions for each outer join.
And what is more important it creates map during join enumeration.

axiom/optimizer/ToGraph.cpp
mbasmanova: It is never null for an outer join (left, right, full). It still can be null for an inner join. I believe hyper edge (many tables on one side) is possible only for an inner join.

axiom/optimizer/Plan.cpp
```diff
@@ -236,6 +236,16 @@ const PlanObjectSet& PlanState::downstreamColumns() const {
...
     }
+
+    if (addFilter) {
+      for (auto i = 0; i < join->leftColumns().size(); ++i) {
+        result.unionColumns(join->leftExprs()[i]);
+      }
+
+      for (auto i = 0; i < join->rightColumns().size(); ++i) {
+        result.unionColumns(join->rightExprs()[i]);
+      }
```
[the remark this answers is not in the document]
mbasmanova: Yup. Changed to 

```
      addExprs(join->leftExprs());
      addExprs(join->rightExprs());
```

axiom/optimizer/QueryGraph.h
```diff
@@ -467,15 +469,47 @@ class JoinEdge {
...
+    ExprVector leftExprs;
+
+    /// Columns produced by the 'right' side of a LEFT or FULL OUTER join.
+    /// Requires 'rightOptional' to be true.
+    ColumnVector rightColumns;
+
+    /// Input expressions corresponding 1:1 to 'rightColumns'.
+    ExprVector rightExprs;
+
     bool directed{false};
```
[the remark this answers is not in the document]
MBkkt: Right now 
1. directed is true when it's join edge for unnest
2. false otherwise

axiom/optimizer/ToVelox.cpp
```diff
@@ -303,17 +303,55 @@ velox::core::TypedExprPtr createArrayForInList(
...
+        if (arg->type()->isRow()) {
+          auto& type = arg->type()->childAt(
+              arg->type()->as<velox::TypeKind::ROW>().getChildIdx(step.field));
+          return std::make_shared<velox::core::FieldAccessTypedExpr>(
+              type, arg, step.field);
+        }
+
+        if (arg->type()->isMap()) {
+          return stepToMapSubscript(step, arg, subscript);
+        }
```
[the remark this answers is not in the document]
MBkkt: Ok

axiom/optimizer/Optimization.cpp
```diff
@@ -1017,6 +1017,53 @@ void Optimization::joinByIndex(
   }
 }
 
+struct ProjectionBuilder {
```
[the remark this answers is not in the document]
MBkkt: > I agree that a point solution you proposed is simpler, but I feel that this is a workaround for just one case and it doesn't correct the design flaw we have discovered.

Well, at least I didn't see cases where this is bad, compared to this solution.
Maybe if you point about such cases, I will agree, but as for now, it looks like just better solution.

I understand the idea, but current implementation have flaws:
1. We need to think about this in every join method implementation
2. Outer Join shouldn't lead to additional projections, it's just not well, but current code allows optimizer to produce project before/in/after join for same expressions, I think it's insane design
3. We can do projections earlier than they're really needed.

I think I don't agree with "outer join should produce columns", because yes, we don't want pushdown filters, but we don't want create columns right after join.
I think some alternative solution can have some special expressions for expressions from optional side of join.
But it seems really difficult to implement this way, instead a lot of simpler just modify how do we pushdown and place filter expressions

WDYT?

axiom/optimizer/PlanObject.h
```diff
@@ -188,6 +188,16 @@ class PlanObjectSet : public BitSet {
...
+  /// Returns the only object stored in this set. The caller must ensure the set
+  /// contains exactly one object.
+  template <typename T = PlanObject>
+  const T* onlyObject() const {
+    VELOX_DCHECK_EQ(size(), 1);
+
+    auto id = velox::bits::findFirstBit(bits_.data(), 0, bits_.size() * 64);
+    return queryCtx()->objectAt(id)->template as<T>();
+  }
+
```
MBkkt: :+1: I thought about this multiple times, thanks

axiom/optimizer/Optimization.cpp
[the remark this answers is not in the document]
MBkkt: So IMO we should fix rules explicitly instead of creating more complicated plan shape to workaround incorrect rules.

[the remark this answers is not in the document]
MBkkt: @mbasmanova Sorry for multiple messages, I initially just don't like the code because it looks like it works worse, should be written multiple times and can produce projections earlier.

But now I finally have good explanation why, I hope it makes sense to you.
> So IMO we should fix rules explicitly instead of creating more complicated plan shape to workaround incorrect rules.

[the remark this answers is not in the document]
MBkkt: > I don't think this is the case. Do you mind sharing the output of DerivedTablePrinter? I believe it will show that filter depends on something like dtN.foo and dtN.foo is defined as an alias for tM.foo. Hence, the filter is effectively represented as depending solely on tM. It doesn't really depend on dtN.

Yes, but it's not true that some information is loss.

> It doesn't really depend on dtN.

It is really depend, because derived table has outer join

> it has join that has optional side, this defines how we pushdown and place filter expressions.

PR #633, opened 2025-11-17
feat: Drop redundant CAST

mbasmanova:
Summary:
Drop redundant casts when processing expressions in ToGraph.

CAST(x as t) ==> x if typeof(x) == t

Differential Revision: D87238235


xiaoxmeng (approved):
@mbasmanova thanks!

PR #636, opened 2025-11-18
refactor: Move ConnectorMetadata::rowIdHandles

MBkkt:
Move `rowIdHandles` from the `ConnectorMetadata` class to the `Table` class. This improves the API design by making row ID handling a concern of the table itself rather than the connector metadata.


MBkkt:
@mbasmanova Could you check this?

PR #639, opened 2025-11-18
fix: Aggregation cardinality estimate

mbasmanova:
Fix cardinality estimate and planning logic for aggregations. This fixes the plan for TPC-H q16.

Aggregation is planned as follows:

If 1 worker and 1 thread: all aggregations are planned as single aggregations.

A global aggregation always has partial + final since combining values on a single thread is never better than combining these on multiple threads with guaranteed reduction to one row per thread. 

Otherwise we compare a plan with a partial aggregation to a plan with a single aggregation and pick the one with less cost. See Optimization::addAggregation().

For use cases where statistics are not available, we provide an option to disable cost-based decisions and always split aggregation into partial + final: alwaysPlanPartialAggregation.

The cost of group-by is in Aggregation::setCostWithGroups(). This has an auxiliary function maxGroups() which calculates the maximum number of possible combinations of grouping keys. 

We distinguish between the number of input rows and the number of distinct input rows the actual input is drawn from. 

We calculate the size of the input domain by grouping the keys by table. If a key depends on many tables, we assign the key to its largest table.  If there is a single key per table in the grouping keys, we use its cardinality. This is always <= the cardinality of the table. If there are many keys from the table, we use a saturating product function to multiply the key cardinalities with a maximum of table cardinality for the count of distinct combinations. If the product is far from the max, this behaves like a multiplication. If the product is close to max, the function approaches max when the product goes to infinity.  In this way, more keys produce more groups but will not overflow the table size.

If we have many tables, we combine the table per-table cardinalities with a saturating product. Here we use the greater of 3*largest table size and an arbitrary 1e10, which is a proxy for  a large group by (10 billion groups).

Having determined the size of the pool from which the input is drawn, we use the coupon collector formula to see what reduction we expect. The size of the group-by is  expectedNumDistincts(inputRowCount, maxGroups).

expectedNumDistincts. This uses the coupon collector formula to estimate how many distinct values will occur in n tuples of input if there are k possible distinct tuples in the input.

We apply this formula to get the total expected fanout: numDistinct / numInput. This is always <= 1.

## Cost of partial aggregation

The partial capacity is the number of rows at the predicted width that fit in the memory budget for partial aggregation.

We see how many rows of input it will take to get this many distinct groups in the input of partial aggregation. With few groups, this will be a large number and with every row being unique this will be the capacity of the partial aggregation. The partial capacity / number of inputs to hit capacity is the reduction from partial. This reduction can however not be greater (lesser fanout) than the total reduction.

Given the reduction and given the expected reduction in the first abandon partial min rows rows, we see if partial will be abandoned at run time.

We set the cost and cardinality accordingly. Abandoned partial has a fanout of 1 and a low cost per byte of input row.

## Cost of final / single aggregation

If we are planning a final aggregation, we scale the input by the reduction expected from partial. This reduces the number of rows that will be shuffled into the final aggregation but does not reduce the size of the hash table. The final table will in the end contain all distinct groups from the input.

We check if we have a local exchange. This is the case if we have more than one thread per worker.. If so, we add a fraction of the shuffle cost as the cost of local exchange.

The cost has unitCost (per row cost) and total memory and fanout. The unit cost formula is the same for partial, final and single aggregations. The cost includes:

- Compute the hash of the grouping keys := Costs::kHashColumnCost * numKeys
- Lookup the hash in the hash table := Costs::hashTableCost(nOut)
- Compare the keys := kKeyCompareCost * numKeys
- Access the row of the hash table (twice) := 2 * Costs::hashRowCost(nOut, rowBytes)
- Update accumulators := aggregates.size() * Costs::kSimpleAggregateCost

The memory estimate is the estimated row size plus a typical overhead of 12. The table size is the number of distinct values at 12 bytes per value. 

The formula represents the added memory cost for larger tables. A more precise formula can be learned from examples.

The intended function of these mechanisms is to decide between partial + final aggregation and a single level of aggregation. Additionally, the cost should be comparable to the costs predicted for other operations like joins and shuffles.

Extracted from https://github.com/facebookincubator/axiom/pull/573
Co-authored-by: oerling

Differential Revision: D87333918

### DIFF
--- a/axiom/optimizer/RelationOp.cpp
+++ b/axiom/optimizer/RelationOp.cpp
@@ -26,6 +26,12 @@
 #include "velox/expression/ScopedVarSetter.h"
 
 namespace facebook::axiom::optimizer {
+
+void PlanCost::add(RelationOp& op) {
+  cost += op.cost().totalCost();
+  cardinality = op.cost().resultCardinality();
+}
+
 namespace {
 
 const auto& relTypeNames() {
@@ -215,6 +221,16 @@ std::string Cost::toString(bool /*detail*/, bool isUnit) const {
   return out.str();
 }
 
+void RelationOp::checkInputCardinality() const {
+  if (input_ != nullptr) {
+    const auto inputCardinality = input_->resultCardinality();
+    VELOX_CHECK(std::isfinite(inputCardinality));
+
+    // TODO Assert that inputCardinality > 0.
+    VELOX_CHECK_GE(inputCardinality, 0);
+  }
+}
+
 std::string RelationOp::toString() const {
   return RelationOpPrinter::toText(*this);
 }
@@ -392,12 +408,17 @@ Join::Join(
   cost_.fanout = fanout;
 
   const float buildSize = right->resultCardinality();
-  const auto numRightColumns =
-      static_cast<float>(right->input()->columns().size());
-  auto rowCost = numRightColumns * Costs::kHashExtractColumnCost;
-  const auto numLeftKeys = static_cast<float>(leftKeys.size());
-  cost_.unitCost = Costs::hashProbeCost(buildSize) + cost_.fanout * rowCost +
-      numLeftKeys * Costs::kHashColumnCost;
+  const auto numKeys = leftKeys.size();
+  const auto probeCost = Costs::hashTableCost(buildSize) +
+      // Multiply by min(fanout, 1) because most misses will not compare and if
+      // fanout > 1, there is still only one compare.
+      (Costs::kKeyCompareCost * numKeys * std::min<float>(1, fanout)) +
+      numKeys * Costs::kHashColumnCost;
+
+  const auto rowBytes = byteSize(right->input()->columns());
+  const auto rowCost = Costs::hashRowCost(buildSize, rowBytes);
+
+  cost_.unitCost = probeCost + cost_.fanout * rowCost;
 }
 
 namespace {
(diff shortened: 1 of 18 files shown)


mbasmanova:
Agree. Here is what I plan to do:

- Split this change into 2. One is to fix cardinality estimates. Another to adjust the magic numbers in the cost model.
- Will try to figure out some tests for cardinality estimates.
- For magic numbers, will try to provide a small program that runs various operation, measures times and produces the recommendations for the magic numbers.

mbasmanova:
@MBkkt Yes, it is understood that accurate cardinality estimates are the key to optimizer performance. These are way more important than perfect cost model. In fact, no cost model can compensate for wrong cardinality estimates.

axiom/optimizer/RelationOp.cpp
```diff
@@ -215,6 +221,17 @@ std::string Cost::toString(bool /*detail*/, bool isUnit) const {
   return out.str();
 }
 
+void RelationOp::checkInputCardinality() const {
+  if (input_ != nullptr) {
+    const auto inputCardinality = input_->resultCardinality();
+    VELOX_CHECK(!std::isnan(inputCardinality));
+    VELOX_CHECK(std::isfinite(inputCardinality));
```
MBkkt:
```suggestion
    VELOX_CHECK(std::isfinite(inputCardinality));
```
isfinite check that number isn't nan or inf
https://en.cppreference.com/w/cpp/numeric/math/isfinite

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -189,9 +189,8 @@ TEST_F(PlanTest, agg) {
     auto matcher = core::PlanMatcherBuilder()
                        .tableScan()
                        .project({"a", "a + b"})
-                       .partialAggregation()
                        .localPartition()
-                       .finalAggregation()
```
MBkkt: Can you explain please why does old plan worse than new?
It looks like in old plan
1. we computed in parallel `sum(a+b)` for each table partition
2. then we gather results
3. and computed in single `sum(a+b)` for partial results

for new plan it's just a single threaded compute, which looks worse?

mbasmanova: In this test we use TestConnector that doesn't provide stats. Hence, we see table cardinality = 1 and grouping key cardinalities also = 1. Processing 1 row using partial+final is strictly worse than processing it using final alone.

That said, running without stats is probably quite common. It would make sense to add an optimizer option to disable cost-based planning for aggregations similarly how we do that for joins (syntacticJoinOrder). I'll add that.

mbasmanova:
> is it expected that there's no build job for this PR in github?

I don't think so. Might be some infra issue.

axiom/optimizer/tests/PlanTest.cpp
mbasmanova: Yes, I was thinking about that.

PR #649, opened 2025-11-24
fix: WIP Fix cardinality estimates for aggregation

mbasmanova:
Created from @oerling 's branch

### DIFF
--- a/axiom/optimizer/RelationOp.cpp
+++ b/axiom/optimizer/RelationOp.cpp
@@ -22,6 +22,7 @@
 #include "axiom/optimizer/QueryGraph.h"
 #include "axiom/optimizer/RelationOpPrinter.h"
 #include "axiom/optimizer/RelationOpVisitor.h"
+#include "axiom/optimizer/StatsUtils.h"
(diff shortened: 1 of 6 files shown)


axiom/optimizer/RelationOp.cpp
```diff
@@ -559,39 +563,348 @@ Unnest::Unnest(
...
+  } else if (table->is(PlanType::kValuesTableNode)) {
+    auto valuesTable = table->as<ValuesTable>();
+    return valuesTable->cardinality();
+  } else if (table->is(PlanType::kUnnestTableNode)) {
+    auto unnestTable = table->as<UnnestTable>();
+    return unnestTable->cardinality();
+  } else {
+    return 1.0;
+  }
+}
```
MBkkt: This kind of function is written 3rd time!
Can you please make it single function please, this looks good, but it will be a lot of better if it will be consistent across different files

axiom/optimizer/Cost.cpp
MBkkt: unrelated?

Issue #654, opened 2025-11-24
downstreamColumns doesn't work for LambdaExpr


MBkkt:
We have such fix, but I'm not sure it's completely correct...

https://github.com/serenedb/axiom/commit/0bfd55d22e80cda43e301395bb10f8ff38938e23

```sql
select n.nspname schema_name, c.relname table_name 
from pg_catalog.pg_class c 
left join 
pg_catalog.pg_namespace n 
on n.oid  = c.relnamespace 
where c.relkind = ANY(ARRAY['r', 'p', 'f']) order by 1, 2;
```

in graph everything is good, but placeConjunct forget about lambda body has column `c.relkind`

We at least for now, for simplicity compiles `op ANY/ALL` into lambda + presto any/all_match

PR #661, opened 2025-11-26
fix: Populate 'columns' for Lambda expression

mbasmanova:
Summary:
Lambda expression doesn't have inputs, but lamda's body may use columns other than lambda arguments (captures). Record these in Lambda::columns.

Fixes: https://github.com/facebookincubator/axiom/issues/654

Differential Revision: D87900986


MBkkt (approved):
Thanks.

I also had a question about subexprssions.
If they're not necessary can you explain why and add such comment in LambdaExpr ctor?

xiaoxmeng (approved):

mbasmanova:
> If they're not necessary can you explain why and add such comment in LambdaExpr ctor?

I don't know yet. I see these are used in parallel expressions (cse border logic). We may need to update LambdaExpr to populate these properly, but I don't quite understand how to trigger / test this logic.

PR #673, opened 2025-12-02
feat: Add support for constant folding metadata-only subqueries

mbasmanova:
Summary:
Constant fold scalar non-correlated subqueries that can be computed using metadata alone.

For example, Hive connector can compute subqueries that use only partition keys:

```
SELECT ... WHERE ds = (SELECT max(ds) from t)
```

Extend ConnectorMetadata to add APIs to return a set of columns that have discrete values that can be listed from metadata alone. Extend subquery handling in ToGraph to constant fold scalar non-correlated subquery that represents a global aggregation over base table, uses only discrete-value columns and uses aggregation that either ignores duplicates or runs over distinct values.

Differential Revision: D87949063

### DIFF
--- a/axiom/connectors/ConnectorMetadata.h
+++ b/axiom/connectors/ConnectorMetadata.h
@@ -277,6 +277,28 @@ struct LookupKeys {
   bool isAscending{true};
 };
 
+class DiscretePredicates {
+ public:
+  explicit DiscretePredicates(std::vector<const Column*> columns)
+      : columns_(std::move(columns)) {
+    VELOX_CHECK(!columns_.empty());
+  }
+
+  virtual ~DiscretePredicates() = default;
+
+  const std::vector<const Column*>& columns() const {
+    return columns_;
+  }
+
+  /// Returns the next batch of values. Returns empty vector if there are no
+  /// more values. Each velox::Variant is of type 'row' and contains one value
+  /// per column specified in the ctor in order.
+  virtual std::vector<velox::Variant> next() = 0;
+
+ private:
+  const std::vector<const Column*> columns_;
+};
+
 /// Represents a physical manifestation of a table. There is at least
 /// one layout but for tables that have multiple sort orders, partitionings,
 /// indices, column groups, etc. there is a separate layout for each. The layout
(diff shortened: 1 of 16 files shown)


axiom/optimizer/ToGraph.cpp
```diff
@@ -1138,9 +1345,28 @@ AggregationPlanCP ToGraph::translateAggregation(const lp::AggregateNode& agg) {
...
 
+    auto aggName = toName(aggregate->name());
+
     const auto& metadata =
         velox::exec::getAggregateFunctionMetadata(aggregate->name());
 
+    if (metadata.ignoreDuplicates) {
+      funcs = funcs | FunctionSet::kIgnoreDuplicatesAggregate;
+    } else {
+      if ((aggName == toName("min") || aggName == toName("max")) &&
```
MBkkt: @mbasmanova 
```
      if ((aggName == toName("min") || aggName == toName("max")) &&
```
this doesn't work when function names prefixed, e.g. presto_min, presto_max

axiom/connectors/ConnectorMetadata.h
```diff
@@ -277,6 +277,28 @@ struct LookupKeys {
...
+  virtual ~DiscretePredicates() = default;
+
+  const std::vector<const Column*>& columns() const {
+    return columns_;
+  }
+
+  /// Returns the next batch of values. Returns empty vector if there are no
+  /// more values. Each velox::Variant is of type 'row' and contains one value
+  /// per column specified in the ctor in order.
+  virtual std::vector<velox::Variant> next() = 0;
```
MBkkt: This API although simple, in practice is ineffective and inconvenient in some cases.

For an example for SereneDB pg_catalog is system tables that implemented as ValuesNode's.
Because of this we have api to get `std::vector<velox::RowVectorPtr>` for some system table in query.

And for this API something like `virtual velox::RowVectorPtr next() = 0;` and ctor
```
  virtual std::unique_ptr<DiscretePredicates> discretePredicates(
      MemoryPool& pool,
      [[maybe_unused]] const std::vector<const Column*>& columns) const {
    return nullptr;
  }
```

makes more sense

axiom/connectors/ConnectorMetadata.h
```diff
@@ -349,6 +371,27 @@ class TableLayout {
     return sortOrder_;
   }
 
+  /// Set of columns that have discrete values that can be enumerated using
+  /// 'discretePredicates' API.
+  /// For example, Hive connector returns a list of partition keys.
+  virtual const std::vector<const Column*>& discretePredicateColumns() const {
+    static const std::vector<const Column*> kEmpty;
+    return kEmpty;
```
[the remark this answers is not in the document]
mbasmanova: Changing in #689

PR #687, opened 2025-12-03
fix: Restore fanout check for Aggregation

mbasmanova:
Summary:
in Aggregation::setCostWithGroups, use doubles instead of float for intermediates. Less probability of going to 0 or infinity.  Also, computing the number of distincts in a final hash table is done with the pre-partial input size. The input size of final is the size reduced by partial. So, to calculate fanout, make sure it is <= 1 because the number of groups from the reduced input can be less than the number of groups from the unreduced input. 

Co-authored-by: oerling

Differential Revision: D88300623

### DIFF
--- a/axiom/optimizer/RelationOp.cpp
+++ b/axiom/optimizer/RelationOp.cpp
@@ -200,7 +200,7 @@ Distribution TableScan::outputDistribution(
       std::move(orderKeys),
       std::move(orderTypes),
       distribution.numKeysUnique <= numPrefix ? distribution.numKeysUnique : 0,
-      1.0F / baseTable->filterSelectivity,
+      1.0f / baseTable->filterSelectivity,
   };
 }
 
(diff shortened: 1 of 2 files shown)


axiom/optimizer/QueryGraph.h
```diff
@@ -914,6 +925,23 @@ struct ValuesTable : public PlanObject {
...
+    float cardinality = 0;
+    const auto* vectors = *std::get_if<Vectors>(&data);
+    for (auto& vector : *vectors) {
+      cardinality += vector->size();
+    }
+
+    return cardinality;
+  }
+
+  float cardinality_;
```
MBkkt:
```suggestion
  const float cardinality_;
```

axiom/optimizer/QueryGraph.h
```diff
@@ -889,13 +889,24 @@ struct BaseTable : public PlanObject {
...
+        data{std::move(_data)},
+        cardinality_{cardinality(data)} {
+    VELOX_CHECK_NOT_NULL(dataType);
+  }
 
   /// Correlation name, distinguishes between uses of the same values node.
   Name cname{nullptr};
 
-  const logical_plan::ValuesNode& values;
+  const velox::Type* dataType;
```
MBkkt:
```suggestion
  const velox::Type& dataType;
```
it cannot be null, right?
Why

xiaoxmeng (approved):
@mbasmanova thanks!

MBkkt:
@mbasmanova Also tests are red

PR #711, opened 2025-12-17
feat: Multi Statement CLI Execution

greenmtnboy:
Summary:
Leverage the new multi-statement parsing to execute multiple statements when a delimeted query is submitted.

Aligned with presto-cli in that:

Extra padding for results when run non-interactive is minimal. [We'll iterate on this in a followup; it's currently more verbose].

In interactive mode, clearly separate result sets.

Differential Revision: D89218112


mbasmanova (approved):
Thanks.

PR #713, opened 2025-12-18
feat: Add SqlStatement::kindName() convenience method

hdikeman:
Summary:
For convenience, I am adding a kindName() API for ease of logging statement types.

Differential Revision: D89478655

### DIFF
--- a/axiom/sql/presto/CMakeLists.txt
+++ b/axiom/sql/presto/CMakeLists.txt
@@ -16,7 +16,7 @@ add_subdirectory(grammar)
 add_subdirectory(ast)
 add_subdirectory(example)
 
-add_library(axiom_sql_presto_parser PrestoParser.cpp)
+add_library(axiom_sql_presto_parser PrestoParser.cpp SqlStatement.cpp)
 
 target_link_libraries(
   axiom_sql_presto_parser
--- a/axiom/sql/presto/SqlStatement.h
+++ b/axiom/sql/presto/SqlStatement.h
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include "axiom/common/Enums.h"
 #include "axiom/logical_plan/LogicalPlanNode.h"
 
 namespace axiom::sql::presto {
@@ -28,6 +29,8 @@ enum class SqlStatementKind {
   kExplain,
 };
 
+AXIOM_DECLARE_ENUM_NAME(SqlStatementKind);
+
 class SqlStatement {
  public:
   explicit SqlStatement(
@@ -37,6 +40,8 @@ class SqlStatement {
       : kind_{kind}, views_{std::move(views)} {}
 
   virtual ~SqlStatement() = default;
+
+  std::string_view kindName() const;
 
   SqlStatementKind kind() const {
     return kind_;
--- a/axiom/sql/presto/SqlStatement.cpp
+++ b/axiom/sql/presto/SqlStatement.cpp
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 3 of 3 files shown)


axiom/sql/presto/CMakeLists.txt
```diff
@@ -16,10 +16,15 @@ add_subdirectory(grammar)
 add_subdirectory(ast)
 add_subdirectory(example)
 
+add_library(axiom_sql_presto_statement SqlStatement.cpp)
```
mbasmanova: Avoid adding new cmake targets. See https://github.com/facebookincubator/axiom/blob/main/CONTRIBUTING.md#cmake-guidelines

Perhaps, add this file to axiom_sql_presto_parser target.

axiom/sql/presto/SqlStatement.h
```diff
@@ -67,8 +75,16 @@ class SqlStatement {
     return dynamic_cast<const T*>(this);
   }
 
-  /// A set of views used in the query. Each view is identified by a connector
-  /// ID and a view name. Map value contains the text of the view.
+  /// Returns the logical plan of the statement. Throws UNSUPPORTED if the
```
mbasmanova: @hdikeman This API doesn't make sense to me. Do we need this? Let's discuss offline.

> For an EXPLAIN statement, returns the underlying query being explained. 

Why would we do this?

axiom/sql/presto/SqlStatement.h
```diff
@@ -134,24 +152,29 @@ class CreateTableAsSelectStatement : public SqlStatement {
...
   const std::string& tableName() const {
     return tableName_;
   }
 
+  /// Returns the schema of the table being created.
   const facebook::velox::RowTypePtr& tableSchema() const {
     return tableSchema_;
   }
 
+  /// Returns any additional properties of the created table.
```
mbasmanova: redundant comments

axiom/sql/presto/SqlStatement.h
hdikeman: I suppose it's to avoid ugliness like this:
```
    if (statement->isExplain()) {
      auto explain = statement->as<::axiom::sql::presto::ExplainStatement>();
      return extractPlan(explain->statement());
    }

    switch (statement->kind()) {
      case ::axiom::sql::presto::SqlStatementKind::kSelect:
        return statement->as<::axiom::sql::presto::SelectStatement>()->plan();
      case ::axiom::sql::presto::SqlStatementKind::kCreateTableAsSelect:
        return statement
            ->as<::axiom::sql::presto::CreateTableAsSelectStatement>()
            ->plan();
      case ::axiom::sql::presto::SqlStatementKind::kInsert:
        return statement->as<::axiom::sql::presto::InsertStatement>()->plan();
      default:
        VELOX_FAIL(
            "Cannot extract plan from statement type '{}'", statement->kind());
    }
```
Idea being if you're optimizing, it doesn't make sense to force the client to use logic like the above

Then if additional statement types are added which have a plan (e.g. UpdateStatement), you would have to find all these bits of logic and add handling for them.

That being said if you have an alternative proposal I'm not dead set on this particular `plan()` API

axiom/sql/presto/SqlStatement.h
```diff
@@ -67,8 +75,16 @@ class SqlStatement {
...
-  /// ID and a view name. Map value contains the text of the view.
+  /// Returns the logical plan of the statement. Throws UNSUPPORTED if the
+  /// statement does not contain a plan. For an EXPLAIN statement, returns the
+  /// underlying query being explained. Statement implementations which contain
+  /// a query plan must override this method to return the plan associated with
+  /// the statement.
+  virtual const facebook::axiom::logical_plan::LogicalPlanNodePtr& plan() const;
+
+  /// A set of views used in the query. Each view is identified by a combination
+  /// of connector ID and the view name. Map value contains the text of the
```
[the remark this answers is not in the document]
hdikeman: I didn't add this variable, I assume this was added by Amit or somebody to support DPAS, which requires this mapping. I just added the comment

mbasmanova (approved):
Thanks.

PR #726, opened 2025-12-23
fix(parser): Handle floating point values with scientific notation

hdikeman:
Summary:
While we handle DECIMAL_VALUE float constants correctly, we don't handle scientific notation DOUBLE_VALUE constants. The handling is actually identical, we can use the `std::stod` function in either case

In this change I am adding that handling

Differential Revision: D89737221


mbasmanova (approved):
Thanks.

PR #744, opened 2026-01-09
fix(parser): Fix SELECT DISTINCT "$path" ... queries

mbasmanova:
Differential Revision: D90394616


xiaoxmeng (approved):
@mbasmanova thanks!

PR #749, opened 2026-01-13
refactor(parser): Prevent accidental usage of DuckDB SQL Parser

mbasmanova:
Summary:
PrestoParser relies on logical_plan::PlanBuilder for type resolution. By default, PlanBuilder offers APIs that accept SQL strings and utilize the DuckDB Parser for parsing. However, in PrestoParser, we aim to avoid these DuckDB-based APIs and transition other usages of PlanBuilder to leverage PrestoParser instead.

Migration Plan:

1. Enable SQL Parser Specification: Update the construction of PlanBuilder to allow specifying which SQL parser to use.
2. PrestoParser Integration: Implement an always-throwing parser and use it to construct PlanBuilder in PrestoParser to ensure DuckDB parsing is not inadvertently used.
3. Migrate other usages of PlanBuilder to adopt PrestoSQL Parser, removing dependency on DuckDB SQL.

This diff implements steps 1 and 2. A follow-up is needed to implement step 3.

Differential Revision: D90560345

### DIFF
--- a/axiom/logical_plan/PlanBuilder.h
+++ b/axiom/logical_plan/PlanBuilder.h
@@ -28,11 +28,29 @@ namespace facebook::axiom::logical_plan {
 
 class NameMappings;
 
+class ThrowingSqlExpressionsParser : public velox::parse::SqlExpressionsParser {
+ public:
+  velox::core::ExprPtr parseExpr(const std::string& expr) override {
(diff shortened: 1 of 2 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

PR #756, opened 2026-01-14
fix: Fix 'Expr is not a column' error in joinByHash

mbasmanova:
Summary:
TPC-H-based repro:

```
WITH n AS (SELECT coalesce(n_regionkey, 0) as n_regionkey FROM nation)
SELECT * FROM region left join n on n_regionkey = r_regionkey;

Query failed: Exception: VeloxRuntimeError
Error Source: RUNTIME
Error Code: INVALID_STATE
Retriable: False
Expression: expr->isColumn()
Function: inputColumns
File: fbcode/axiom/optimizer/Optimization.cpp
Line: 1159
```

In this example, coalesce(n_regionkey, 0) expression is used as a join key and is also projected out of the join. The error happened because we tried to project 'coalesce(n_regionkey, 0)' from the results of the join, which is not possible. The fix is to compute 'coalesce(n_regionkey, 0)' expression using a Project node before the join.

Differential Revision: D90701308

### DIFF
--- a/axiom/optimizer/Optimization.cpp
+++ b/axiom/optimizer/Optimization.cpp
@@ -1137,12 +1137,16 @@ void Optimization::joinByIndex(
   }
 }
 
+namespace {
+
 struct ProjectionBuilder {
   ColumnVector columns;
   ExprVector exprs;
 
   // Project 'expr' as 'column'.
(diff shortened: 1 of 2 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

PR #759, opened 2026-01-14
fix: Correlated EXISTS subqueries that do not project columns used in correlated conjuncts

mbasmanova:
Summary:
Fixes queries like

> SELECT * FROM region WHERE EXISTS (SELECT 1 FROM nation WHERE r_regionkey = n_regionkey)

These queries used to fail with 

```
Reason: (1 vs. 2) Correlated conjuncts referencing multiple outer tables are not supported: eq(t2.r_regionkey, t4.n_regionkey)
Retriable: False
Expression: 1 == tables.size()
Function: processSubqueries
File: fbcode/axiom/optimizer/ToGraph.cpp
Line: 2044
```

Differential Revision: D90724079

### DIFF
--- a/axiom/optimizer/DerivedTable.cpp
+++ b/axiom/optimizer/DerivedTable.cpp
@@ -477,6 +477,17 @@ bool DerivedTable::isWrapOnly() const {
 }
 
 ExprCP DerivedTable::exportExpr(ExprCP expr) {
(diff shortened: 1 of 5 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

PR #763, opened 2026-01-16
feat: Add ISerializable support for Expr

kunigami:
Implements serialization/deserialization for all Expr types except
SubqueryExpr by integrating with the Velox ISerializable interface.

Expression types covered:
- ConstantExpr: Uses Variant::serialize() for value serialization
- InputReferenceExpr: Serializes column name
- CallExpr: Serializes function name and input expressions
- SpecialFormExpr: Serializes form enum and input expressions
- LambdaExpr: Serializes signature (ROW type) and body expression
- AggregateExpr: Serializes name, inputs, filter, sorting fields, and isDistinct
- WindowExpr: Serializes name, inputs, partition keys, sorting fields, frame, and ignoreNulls

Also includes helper serialization for:
- SortOrder enum
- SortingField struct
- WindowExpr::Frame struct (window type, bound types, bound values)

SubqueryExpr is skipped because it requires plan node serialization
which is not yet implemented.

Differential Revision: D90826663

### DIFF
--- a/axiom/logical_plan/tests/ExprSerdeTest.cpp
+++ b/axiom/logical_plan/tests/ExprSerdeTest.cpp
@@ -1,0 +1,182 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <gtest/gtest.h>
+
+#include "axiom/logical_plan/Expr.h"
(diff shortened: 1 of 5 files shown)


axiom/logical_plan/tests/ExprSerdeTest.cpp
```diff
@@ -0,0 +1,126 @@
...
+ * limitations under the License.
+ */
+
+#include <gtest/gtest.h>
+
+#include "axiom/logical_plan/Expr.h"
+#include "velox/common/serialization/Serializable.h"
+#include "velox/type/Type.h"
+
+namespace facebook::axiom::logical_plan::test {
```
mbasmanova: The convention is to put the test inside anonymous namespace within namespace that contains code under test.

```
namespace facebook::axiom::logical_plan {
namespace {
```

axiom/logical_plan/tests/ExprSerdeTest.cpp
```diff
@@ -0,0 +1,185 @@
...
+class ExprSerdeTest : public testing::Test {
+ protected:
+  static void SetUpTestSuite() {
+    velox::Type::registerSerDe();
+    Expr::registerSerDe();
+    velox::functions::prestosql::registerAllScalarFunctions();
+    velox::aggregate::prestosql::registerAllAggregateFunctions();
+  }
+
+  /// Verifies round-trip serialization by comparing toString() outputs.
```
mbasmanova: /// is for public APIs. Use //

kunigami:
build is failing. will take a look

PR #797, opened 2026-01-25
feat(cli): Create SqlQueryRunner helpers on the fly

jagill:
Summary:
Remove PrestoParser, SchemaResolver, and VeloxHistory as instance variables and
instead create as needed. This allows us to have each query have its own
default schema (and even connector). Also allow caller to specify query maxWait
in RunOptions

Also change some VELOX_CHECKs to VELOX_USER_CHECKs when the issue is in
user-supplied SQL statements.

Differential Revision: D91166982

### DIFF
--- a/axiom/cli/SqlQueryRunner.h
+++ b/axiom/cli/SqlQueryRunner.h
@@ -16,22 +16,19 @@
 #pragma once
 
 #include <folly/executors/IOThreadPoolExecutor.h>
-#include "axiom/connectors/SchemaResolver.h"
 #include "axiom/optimizer/DerivedTable.h"
-#include "axiom/optimizer/VeloxHistory.h"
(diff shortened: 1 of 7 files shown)


axiom/cli/SqlQueryRunner.h
```diff
@@ -143,10 +153,11 @@ class SqlQueryRunner {
       const RunOptions& options);
 
   static void waitForCompletion(
-      const std::shared_ptr<facebook::axiom::runner::LocalRunner>& runner) {
+      const std::shared_ptr<facebook::axiom::runner::LocalRunner>& runner,
+      int32_t timeout) {
```
mbasmanova: Should we check that timeout >= 0? Should we document what timeout 0 means? No timeout?

mbasmanova (approved):
Thanks.

PR #811, opened 2026-01-28
fix: Fix qualified cross join handling in optimizer

oerling:
Summary:
This diff fixes the handling of qualified cross joins in the
Axiom optimizer.  A qualified cross join is a cross join with a
condition that always evaluates to true (e.g., `1=1` or an expression
that effectively reduces to a tautology). Non-equality joins are
treated as cross joins until we add conectors that support index base
range lookup.

The key changes ensure that join edges properly track their endpoint
tables even when there are no column references in the join
condition. This enables queries with outer cross joins.

Note: Explicit right outer cross joins will be supported when normalizing right joins to left joins.

Differential Revision: D91637123


mbasmanova (approved):
Thanks.

PR #812, opened 2026-01-28
feat: Add support for EXISTS subqueries with non-equi correlated filter

mbasmanova:
Summary:
Enable queries like

> select * from nation where exists (select 1 from region where r_name like concat('%', n_name, '%'))

Differential Revision: D91686683


xiaoxmeng (approved):
@mbasmanova thanks!

PR #814, opened 2026-01-28
feat: Handle LIMIT 0 in ToGraph (#16146)

natashasehgal:
Summary: 
This change handles LIMIT 0 by skipping the input subtree and replacing it with an empty `ValuesTable`. This optimization avoids building unnecessary plan nodes when the result is known to be zero rows.

Changes:
- Early LIMIT 0 detection in ToGraph: When LimitNode::count() == 0, we skip building the input subtree entirely and create an empty ValuesTable with the proper output schema.

- Schema preservation in `ToVelox`: When converting an empty `ValuesTable` to Velox, we create a RowVector with 0 rows but proper type schema. This ensures column names and types are preserved (matching SQL behavior where SELECT * FROM t LIMIT 0 returns column metadata with 0 rows).

- Also added an update to AXIOM_ASSERT_PLAN to fix issue of double evaluation.

Differential Revision: D91646381

### DIFF
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -1861,16 +1861,21 @@ void ToGraph::makeValuesTable(const lp::ValuesNode& values) {
         &registerVariant(velox::Variant::array(std::move(variants)))->array();
   }
 
-  auto* valuesTable =
-      make<ValuesTable>(toType(values.outputType()), std::move(data));
-  valuesTable->cname = newCName("vt");
+  auto* valuesTable = makeValuesTable(values, std::move(data));
   planLeaves_[&values] = valuesTable;
+  currentDt_->addTable(valuesTable);
+}
 
-  auto channels = usedChannels(values);
-  const auto& type = values.outputType();
-  const auto& names = values.outputType()->names();
+ValuesTable* ToGraph::makeValuesTable(
+    const lp::LogicalPlanNode& node,
+    ValuesTable::Data data) {
+  auto* valuesTable = make<ValuesTable>(toType(node.outputType()), data);
+  valuesTable->cname = newCName("vt");
+
+  const auto& type = node.outputType();
+  const auto& names = type->names();
   const auto cardinality = valuesTable->cardinality();
-  for (auto i : channels) {
+  for (auto i : usedChannels(node)) {
     VELOX_DCHECK_LT(i, type->size());
 
     const auto& name = names[i];
@@ -1882,7 +1887,7 @@ void ToGraph::makeValuesTable(const lp::ValuesNode& values) {
     renames_[name] = column;
   }
 
-  currentDt_->addTable(valuesTable);
+  return valuesTable;
 }
 
 void ToGraph::addProjection(const lp::ProjectNode& project) {
@@ -2267,6 +2272,12 @@ void ToGraph::addLimit(const lp::LimitNode& limit) {
     currentDt_->limit = limit.count();
     currentDt_->offset = limit.offset();
   }
+}
+
+void ToGraph::makeEmptyValuesTable(const lp::LogicalPlanNode& node) {
+  auto* emptyData = &registerVariant(velox::Variant::array({}))->array();
+  auto* valuesTable = makeValuesTable(node, emptyData);
+  currentDt_->addTable(valuesTable);
 }
 
 void ToGraph::addWrite(const lp::TableWriteNode& tableWrite) {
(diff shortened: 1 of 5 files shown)


axiom/optimizer/ToVelox.cpp
```diff
@@ -646,6 +646,10 @@ velox::core::PlanNodePtr ToVelox::makeOrderBy(
     const OrderBy& op,
     runner::ExecutableFragment& fragment,
     std::vector<runner::ExecutableFragment>& stages) {
+  if (op.limit == 0) {
```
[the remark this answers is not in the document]
natashasehgal: Makes sense to prune earlier in optimization stages. If the result is zero rows perhaps distribution properties can be ignored/become irrelevant checking this

axiom/optimizer/ToVelox.cpp
```diff
@@ -1390,6 +1389,19 @@ velox::core::PlanNodePtr ToVelox::makeValues(
     }
   }
 
+  if (newValues.empty()) {
+    auto* pool = queryCtx()->optimization()->evaluator()->pool();
+    std::vector<velox::VectorPtr> emptyChildren;
+    emptyChildren.reserve(newType->size());
+    for (size_t i = 0; i < newType->size(); ++i) {
```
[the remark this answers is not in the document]
natashasehgal: Yes updated

axiom/optimizer/ToGraph.cpp
```diff
@@ -1865,11 +1865,18 @@ void ToGraph::makeValuesTable(const lp::ValuesNode& values) {
...
-  for (auto i : channels) {
+  addValuesTableColumns(valuesTable, values, valuesTable->cardinality());
+
+  currentDt_->addTable(valuesTable);
+}
+
+void ToGraph::addValuesTableColumns(
+    ValuesTable* valuesTable,
+    const lp::LogicalPlanNode& node,
+    float cardinality) {
```
mbasmanova: Do we need to pass 'cardinality'? Can we just use valuesTable->cardinality()?

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -953,6 +953,75 @@ TEST_F(PlanTest, limitAfterOrderBy) {
   }
 }
 
+TEST_F(PlanTest, zeroLimit) {
+  testConnector_->addTable("t", ROW({"a"}, {BIGINT()}));
+  testConnector_->addTable("u", ROW({"x"}, {BIGINT()}));
+  // SELECT * FROM t LIMIT 0
```
mbasmanova: Let's use parseSelect and remove this comment. 

```
auto logicalPlan = parseSelect("SELECT * FROM t LIMIT 0", kTestConnectorId);
```

mbasmanova:
> Based on https://github.com/serenedb/axiom/commit/0d8a8e207ed54b089cfe32517cd585cf4817f662

This is no longer the case, is it? Perhaps, update PR description.

MBkkt (changes requested):
As I said solution with ToGraph is a bad idea, right now PR contains bugs and more complex code and no benefits.

Bug related to such case `SELECT * (SELECT * FROM t LIMIT 1) OFFSET 1`

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -953,6 +953,73 @@ TEST_F(PlanTest, limitAfterOrderBy) {
   }
 }
 
+TEST_F(PlanTest, zeroLimit) {
+  testConnector_->addTable("t", ROW({"a"}, {BIGINT()}));
+  testConnector_->addTable("u", ROW({"x"}, {BIGINT()}));
+  {
+    auto logicalPlan = parseSelect("SELECT * FROM t LIMIT 0", kTestConnectorId);
```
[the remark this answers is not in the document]
natashasehgal: Updated this. 

Should we be using SCOPED_TRACE on all optimizer tests as it's end to end? I noticed Presto Parser has sql style queries but no SCOPED_TRACE

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -953,6 +953,72 @@ TEST_F(PlanTest, limitAfterOrderBy) {
...
+    return toSingleNodePlan(parseSelect(sql, kTestConnectorId));
+  };
+
+  {
+    const auto matcher = core::PlanMatcherBuilder{}.values().build();
+
+    {
+      const auto query = "SELECT * FROM t LIMIT 0";
+      SCOPED_TRACE(query);
+      AXIOM_ASSERT_PLAN(planSql(query), matcher);
```
mbasmanova: You may need to fix AXIOM_ASSERT_PLAN. Currently, it will call planSql(query) twice:

```
#define AXIOM_ASSERT_PLAN(plan, matcher) \
  ASSERT_TRUE(matcher->match(plan)) << plan->toString(true, true);
```

axiom/optimizer/tests/PlanTest.cpp
```diff
@@ -953,6 +953,72 @@ TEST_F(PlanTest, limitAfterOrderBy) {
   }
 }
 
+TEST_F(PlanTest, zeroLimit) {
+  testConnector_->addTable("t", ROW({"a"}, {BIGINT()}));
```
mbasmanova: nit: ROW({"a"}, BIGINT()) for readability

MBkkt (commented):
It probably works, but doesn't look good to me

Issue #813, opened 2026-01-28
Support CREATE TABLE without select in Axiom


hdikeman:
Pure table create is not currently supported, just CREATE TABLE AS SELECT

All the pieces are in place from an execution point of view, the only missing piece is the PrestoParser support (pure create would not go through optimization), which should be straightforward given that CTAS is supported. CTAS should contain a superset of anything required for a pure create

I will start working on this

PR #815, opened 2026-01-28
feat(parser): Add support for CREATE TABLE

hdikeman:
Summary:
Adding support for pure CREATE TABLE statements

Originally I asserted that CREATE TABLE components are a superset of CTAS: this is not true, there are two pieces which are not present in CTAS:

1. `LIKE` clauses, which pull from an existing table schema to define a table schema
2. constraints, which place extra conditions or add characteristics to columns

A new type was defined for constraint, but the logic for `LIKE` table lookup is identical to a table scan, so an existing utility can be used.

Common functionality was factored out between CREATE TABLE and CTAS where appropriate, but besides that, the bulk of this change was handling LIKE and constraint clauses

Differential Revision: D91715648

### DIFF
--- a/axiom/sql/presto/PrestoParser.cpp
+++ b/axiom/sql/presto/PrestoParser.cpp
@@ -31,6 +31,7 @@
 #include "velox/exec/WindowFunction.h"
 #include "velox/functions/FunctionRegistry.h"
 #include "velox/functions/prestosql/types/TimestampWithTimeZoneType.h"
+#include "velox/type/parser/ParserUtil.h"
 
 namespace axiom::sql::presto {
 namespace {
@@ -138,8 +139,7 @@ void findAggregates(
     case core::IExpr::Kind::kFieldAccess:
       return;
     case core::IExpr::Kind::kCall: {
-      if (facebook::velox::exec::getAggregateFunctionEntry(
-              expr->as<core::CallExpr>()->name())) {
+      if (exec::getAggregateFunctionEntry(expr->as<core::CallExpr>()->name())) {
         if (aggregateSet.emplace(expr).second) {
           aggregates.emplace_back(lp::ExprApi(expr));
         }
@@ -197,7 +197,7 @@ class ExprAnalyzer : public DefaultTraversalVisitor {
 
   void visitFunctionCall(FunctionCall* node) override {
     const auto& name = node->name()->suffix();
-    if (facebook::velox::exec::getAggregateFunctionEntry(name)) {
+    if (exec::getAggregateFunctionEntry(name)) {
       VELOX_USER_CHECK(
           !aggregateName_.has_value(),
           "Cannot nest aggregations inside aggregation: {}({})",
@@ -282,9 +282,8 @@ class RelationPlanner : public AstVisitor {
 
   lp::ExprApi toExpr(
       const ExpressionPtr& node,
-      std::unordered_map<
-          const facebook::velox::core::IExpr*,
-          lp::PlanBuilder::AggregateOptions>* aggregateOptions = nullptr) {
+      std::unordered_map<const core::IExpr*, lp::PlanBuilder::AggregateOptions>*
+          aggregateOptions = nullptr) {
     switch (node->type()) {
       case NodeType::kIdentifier:
         return lp::Col(canonicalizeIdentifier(*node->as<Identifier>()));
@@ -573,10 +572,10 @@ class RelationPlanner : public AstVisitor {
       case NodeType::kTimestampLiteral: {
         auto literal = node->as<TimestampLiteral>();
 
-        auto timestamp = facebook::velox::util::fromTimestampWithTimezoneString(
+        auto timestamp = util::fromTimestampWithTimezoneString(
             literal->value().c_str(),
             literal->value().size(),
-            facebook::velox::util::TimestampParseMode::kPrestoCast);
+            util::TimestampParseMode::kPrestoCast);
 
         VELOX_USER_CHECK(
             !timestamp.hasError(),
@@ -586,11 +585,9 @@ class RelationPlanner : public AstVisitor {
 
         if (timestamp.value().timeZone != nullptr) {
           return lp::Cast(
-              facebook::velox::TIMESTAMP_WITH_TIME_ZONE(),
-              lp::Lit(literal->value()));
+              TIMESTAMP_WITH_TIME_ZONE(), lp::Lit(literal->value()));
         } else {
-          return lp::Cast(
-              facebook::velox::TIMESTAMP(), lp::Lit(literal->value()));
+          return lp::Cast(TIMESTAMP(), lp::Lit(literal->value()));
         }
       }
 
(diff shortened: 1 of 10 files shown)


hdikeman:
This is in service of https://github.com/facebookincubator/axiom/issues/813

axiom/sql/presto/SqlStatement.h
```diff
@@ -121,6 +126,73 @@ class InsertStatement : public SqlStatement {
...
+  const std::string& tableName() const {
+    return tableName_;
+  }
+
+  const facebook::velox::RowTypePtr& tableSchema() const {
+    return tableSchema_;
+  }
+
+  const std::unordered_map<std::string, facebook::axiom::logical_plan::ExprPtr>&
+  properties() const {
```
mbasmanova: Would you document what are the properties and why are they expressed using ExprPtr? Which kinds of expressions are allowed? Let's add sanity checks.

axiom/sql/presto/tests/PrestoParserTest.cpp
```diff
@@ -1532,6 +1565,120 @@ TEST_F(PrestoParserTest, createTableAsSelect) {
       });
 }
 
+TEST_F(PrestoParserTest, createTable) {
```
mbasmanova: Would it make sense to create a separate test file for DDL statements?

axiom/sql/presto/SqlStatement.h
```diff
@@ -121,6 +126,73 @@ class InsertStatement : public SqlStatement {
   const facebook::axiom::logical_plan::LogicalPlanNodePtr plan_;
 };
 
+class CreateTableStatement : public SqlStatement {
+ public:
+  struct Constraint {
+    enum class Type {
+      /// e.g. CONSTRAINT constraint_name UNIQUE(col)
```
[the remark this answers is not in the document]
hdikeman: Thanks, let me add more details to the description here

axiom/sql/presto/tests/PrestoParserTest.cpp
hdikeman: Hmm, in my mind having all the PrestoParser-only tests together was preferable

But I will defer to your judgement

axiom/sql/presto/PrestoParser.cpp
```diff
@@ -2109,6 +2116,84 @@ SqlStatementPtr parseCreateTableAsSelect(
...
+  std::vector<TypePtr> types;
+  std::vector<CreateTableStatement::Constraint> constraints;
+
+  for (const auto& element : createTable.elements()) {
+    switch (element->type()) {
+      case NodeType::kColumnDefinition: {
+        auto* columnDef = element->as<ColumnDefinition>();
+        names.push_back(columnDef->name()->value());
+
+        auto type = typeFromString(columnDef->columnType());
```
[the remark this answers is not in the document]
hdikeman: Sure let me open an issue and follow up there

opened https://github.com/facebookincubator/velox/issues/16170

axiom/sql/presto/SqlStatement.h
```diff
@@ -121,6 +126,73 @@ class InsertStatement : public SqlStatement {
...
+    std::string name;
+    std::vector<std::string> columns;
+    Type type;
+  };
+
+  CreateTableStatement(
+      std::string connectorId,
+      std::string tableName,
+      facebook::velox::RowTypePtr tableSchema,
+      std::unordered_map<std::string, facebook::axiom::logical_plan::ExprPtr>
```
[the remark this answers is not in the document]
hdikeman: I left the obvious ones alone, e.g. tableSchema() I thought was self-explanatory. But I can add something if desired

axiom/sql/presto/tests/PrestoParserTest.cpp
hdikeman: Will handle this in a follow-up, as per offline discussion. But PrestoParserTest.cpp is getting pretty large

hdikeman:
@mbasmanova thanks for comments, just updated, these are the main points of edits:

1. added additional validation to CreateTableStatement and CreateTableAsSelectStatement. Since the constructors got quite large, I moved them to SqlStatement.cpp. Also documented the validations in the class definitions
2. updated documentation as per earlier review comments. Hopefully the constraint one is a bit more clear now
3. as per discussion, avoided adding type constraints to table properties given Presto behavior
4. added more failure test coverage after offline review comments from @HeidiHan0000 (thanks Heidi!)

Apologies, did not have a chance to open the Velox docs PR today, I will do this tomorrow and drop a link here once I do

Thanks

axiom/sql/presto/PrestoParser.cpp
```diff
@@ -2109,6 +2116,84 @@ SqlStatementPtr parseCreateTableAsSelect(
...
+        VELOX_UNREACHABLE(
+            "Unexpected table element type: {}",
+            static_cast<int>(element->type()));
+    }
+  }
+
+  return std::make_shared<CreateTableStatement>(
+      connectorTable.first,
+      connectorTable.second,
+      facebook::velox::ROW(std::move(names), std::move(types)),
```
mbasmanova: Can we drop facebook::velox::? We don't use this with typeFromString.

axiom/sql/presto/SqlStatement.cpp
```diff
@@ -34,6 +35,25 @@ const auto& statementKindNames() {
   return kNames;
 }
 
+std::string toLower(const std::string& s) {
+  std::string result = s;
+  std::transform(result.begin(), result.end(), result.begin(), ::tolower);
```
mbasmanova: std::tolower ?

axiom/sql/presto/SqlStatement.cpp
```diff
@@ -42,4 +62,56 @@ std::string_view SqlStatement::kindName() const {
...
+    facebook::velox::RowTypePtr tableSchema,
+    std::unordered_map<std::string, facebook::axiom::logical_plan::ExprPtr>
+        properties,
+    bool ifNotExists,
+    std::vector<Constraint> constraints)
+    : SqlStatement(SqlStatementKind::kCreateTable),
+      connectorId_{std::move(connectorId)},
+      tableName_{std::move(tableName)},
+      tableSchema_{std::move(tableSchema)},
+      properties_{std::move(properties)},
```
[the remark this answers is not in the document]
hdikeman: Sounds good

There is this utility in ToGraph:
```
bool looksConstant(const lp::ExprPtr& expr) {
  if (expr->isConstant()) {
    return true;
  }
  if (expr->isInputReference()) {
    return false;
  }
  return std::ranges::all_of(expr->inputs(), looksConstant);
}
```
I'm not sure that there's any point to recursing for anything but Call or SpecialForm expressions, so for now I can do this:
```
bool isConstantFoldable(const lp::ExprPtr& expr) {
  if (expr->isConstant()) {
    return true;
  }
  if (expr->isCall() || expr->isSpecialForm()) {
    for (const auto& input : expr->inputs()) {
      if (!isConstantFoldable(input)) {
        return false;
      }   
    }   
    return true;
  }
  return false;
}
```

But I could squash these two utilities together and put in logical_plan/Expr.h somewhere if you would prefer as well

PR #816, opened 2026-01-28
refactor: Introduce DesiredDistribution

mbasmanova:
Summary:
De-flake Optimization::makePlan by introducing a dedicated struct to represent the desired distribution.

The DesiredDistribution struct specifies:
 - The desired partitioning keys
 - The partitioning function

DesiredDistribution does not support gather or broadcast distributions. If a broadcast distribution is required, it must be applied explicitly.

Differential Revision: D91708782


xiaoxmeng (approved):
@mbasmanova LGTM. Thanks!

PR #817, opened 2026-01-28
feat(parser): Add GROUPING SETS, ROLLUP, and CUBE support

natashasehgal:
Summary: Adds support for GROUPING SETS, ROLLUP, and CUBE.

**Design:**
- Unified code path for GROUP BY and GROUPING SETS via expandGroupingSets()
- Two caching mechanisms: AST pointer cache (exprCache) for subquery ordinal deduplication. And Expression cache (exprToIndex) for grouping key deduplication
- PlanBuilder: string, expression, and index-based grouping sets APIs with rollup()/cube() convenience methods
- CUBE limited to 30 columns similar to [Presto](https://github.com/prestodb/presto/blob/master/presto-main-base/src/main/java/com/facebook/presto/sql/analyzer/StatementAnalyzer.java#L3920)

**Cache Detail:**
Two caches used during grouping set expansion:

1. exprCache (AST pointer → ExprApi)

Caches resolved expressions during expandGroupingSets(). For ordinals like GROUP BY 1, the cache key is the SELECT item's AST pointer. 

This ensures the same ordinal always returns the same ExprApi instance - necessary for subqueries where ExprMap (IExpr* based) cannot deduplicate different instances of the same subquery.

2. exprToIndex (IExpr → int32_t)

Builds unique grouping keys when converting to index-based representation. For GROUPING SETS ((a, b), (a, c)), this produces groupingKeys = [a, b, c] and indices = [[0, 1], [0, 2]].

Differential Revision: D91610080

### DIFF
--- a/axiom/logical_plan/PlanBuilder.cpp
+++ b/axiom/logical_plan/PlanBuilder.cpp
@@ -15,6 +15,7 @@
  */
 #include "axiom/logical_plan/PlanBuilder.h"
 #include <velox/common/base/Exceptions.h>
+#include <numeric>
 #include <vector>
 #include "axiom/connectors/ConnectorMetadata.h"
 #include "axiom/logical_plan/NameMappings.h"
@@ -487,10 +488,10 @@ PlanBuilder& PlanBuilder::with(const std::vector<ExprApi>& projections) {
   return *this;
 }
 
-PlanBuilder& PlanBuilder::aggregate(
-    const std::vector<std::string>& groupingKeys,
+namespace {
+std::vector<PlanBuilder::AggregateOptions> parseAggregateOptions(
     const std::vector<std::string>& aggregates) {
-  std::vector<AggregateOptions> options;
+  std::vector<PlanBuilder::AggregateOptions> options;
   options.reserve(aggregates.size());
   for (const auto& sql : aggregates) {
     auto aggregateExpr = velox::duckdb::parseAggregateExpr(sql, {});
@@ -508,8 +509,17 @@ PlanBuilder& PlanBuilder::aggregate(
         std::move(sortingKeys),
         aggregateExpr.distinct);
   }
+  return options;
+}
+} // namespace
 
-  return aggregate(parse(groupingKeys), parse(aggregates), options);
+PlanBuilder& PlanBuilder::aggregate(
+    const std::vector<std::string>& groupingKeys,
+    const std::vector<std::string>& aggregates) {
+  return aggregate(
+      parse(groupingKeys),
+      parse(aggregates),
+      parseAggregateOptions(aggregates));
 }
 
 PlanBuilder& PlanBuilder::aggregate(
@@ -531,38 +541,7 @@ PlanBuilder& PlanBuilder::aggregate(
   std::vector<AggregateExprPtr> exprs;
   exprs.reserve(aggregates.size());
 
-  VELOX_USER_CHECK(options.size() == aggregates.size());
-  for (size_t i = 0; i < aggregates.size(); ++i) {
-    const auto& aggregate = aggregates[i];
-
-    ExprPtr filter;
-    if (options[i].filter != nullptr) {
-      filter = resolveScalarTypes(options[i].filter);
-    }
-
-    std::vector<SortingField> sortingFields;
-    sortingFields.reserve(options[i].orderBy.size());
-    for (const auto& key : options[i].orderBy) {
-      auto expr = resolveScalarTypes(key.expr.expr());
-
-      sortingFields.push_back(
-          SortingField{expr, SortOrder(key.ascending, key.nullsFirst)});
-    }
-
-    AggregateExprPtr expr;
-    expr = resolveAggregateTypes(
-        aggregate.expr(), filter, sortingFields, options[i].distinct);
-
-    if (aggregate.name().has_value()) {
-      const auto& alias = aggregate.name().value();
-      outputNames.push_back(newName(alias));
-      newOutputMapping->add(alias, outputNames.back());
-    } else {
-      outputNames.push_back(newName(expr->name()));
-    }
-
-    exprs.emplace_back(std::move(expr));
-  }
+  resolveAggregates(aggregates, options, outputNames, exprs, *newOutputMapping);
 
   node_ = std::make_shared<AggregateNode>(
       nextId(),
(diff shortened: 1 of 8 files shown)


axiom/logical_plan/PlanBuilder.cpp
```diff
@@ -580,6 +580,81 @@ PlanBuilder& PlanBuilder::aggregate(
...
+      newOutputMapping->add(alias, outputNames.back());
+    } else {
+      outputNames.push_back(newName(expr->name()));
+    }
+
+    exprs.emplace_back(std::move(expr));
+  }
+
+  if (!groupingSets.empty()) {
+    static const std::string kGroupingSetIndexName = "$grouping_set_id";
```
mbasmanova: What is happening here? Shouldn't we extend logical plan to allow for grouping sets representation? We don't want to convert grouping sets into regular group bys here. That's the job of the optimizer.

axiom/sql/presto/ast/AstBuilder.cpp
```diff
@@ -925,24 +925,40 @@ std::any AstBuilder::visitSingleGroupingSet(
...
+      std::make_shared<GroupingSets>(getLocation(ctx), sets));
 }
 
 std::any AstBuilder::visitGroupingSet(
     PrestoSqlParser::GroupingSetContext* ctx) {
   trace("visitGroupingSet");
-  return visitChildren("visitGroupingSet", ctx);
+
+  // This is called via visitSingleGroupingSet, which handles it directly.
+  // We return the expressions for consistency if called directly.
```
mbasmanova: If we don't expect this method to be called, then let's assert.

axiom/logical_plan/PlanBuilder.h
```diff
@@ -259,6 +259,20 @@ class PlanBuilder {
...
 
+  /// Aggregate with grouping sets support for ROLLUP, CUBE, and GROUPING SETS.
+  /// @param groupingKeys All grouping key expressions used across all sets.
+  /// @param groupingSets Vector of grouping sets, where each set is a vector
+  /// of indices into groupingKeys. For example, with groupingKeys [a, b, c]:
+  ///   - ROLLUP(a,b,c) would have sets: [[0,1,2], [0,1], [0], []]
+  ///   - CUBE(a,b) would have sets: [[0,1], [0], [1], []]
+  /// @param aggregates The aggregate expressions to compute.
+  /// @param options Per-aggregate options (filter, orderBy, distinct).
+  PlanBuilder& aggregateWithGroupingSets(
```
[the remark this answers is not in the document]
natashasehgal: Done please check

axiom/logical_plan/PlanBuilder.cpp
```diff
@@ -533,51 +543,233 @@ PlanBuilder& PlanBuilder::aggregate(
...
 
-  return *this;
+PlanBuilder& PlanBuilder::cube(
+    const std::vector<ExprApi>& groupingKeys,
+    const std::vector<ExprApi>& aggregates,
+    const std::vector<AggregateOptions>& options,
+    const std::string& groupingSetIndexName) {
+  // CUBE(a, b) -> [[0,1], [0], [1], []]
+  const size_t n = groupingKeys.size();
+  VELOX_USER_CHECK_LE(n, 30, "CUBE supports at most 30 columns");
```
mbasmanova: Where does this limit come from? If this is a Presto-specific limit, please, move it to PrestoParser and add a reference to documentation. If Presto documentation is lacking, please, open a PR in PrestoDB report to add.

Is there no limit for rollup?

axiom/logical_plan/PlanBuilder.h
```diff
@@ -259,6 +259,71 @@ class PlanBuilder {
...
+  /// @param options Per-aggregate options (filter, orderBy, distinct).
+  /// @param groupingSetIndexName Name of the output column that identifies
+  /// which grouping set each row belongs to.
+  PlanBuilder& aggregateWithGroupingSets(
+      const std::vector<std::vector<ExprApi>>& groupingSets,
+      const std::vector<ExprApi>& aggregates,
+      const std::vector<AggregateOptions>& options,
+      const std::string& groupingSetIndexName);
+
+  /// String-based overload for testing convenience.
```
mbasmanova: perhaps,

/// Convenience overload that parses SQL strings for grouping keys and aggregates.

axiom/logical_plan/PlanBuilder.cpp
```diff
@@ -533,51 +543,233 @@ PlanBuilder& PlanBuilder::aggregate(
...
+}
+
+PlanBuilder& PlanBuilder::aggregateWithGroupingSets(
+    const std::vector<std::vector<ExprApi>>& groupingSets,
+    const std::vector<ExprApi>& aggregates,
+    const std::vector<AggregateOptions>& options,
+    const std::string& groupingSetIndexName) {
+  // Extract unique grouping keys from all sets and build index mapping.
+  std::vector<ExprApi> groupingKeys;
+  std::unordered_map<
```
mbasmanova: Use F14:

```
using ExprSet =
    folly::F14FastSet<core::ExprPtr, core::IExprHash, core::IExprEqual>;

template <typename V>
using ExprMap =
    folly::F14FastMap<core::ExprPtr, V, core::IExprHash, core::IExprEqual>;
```

(perhaps, add these to fbcode/velox/parse/IExpr.h)

axiom/sql/presto/PrestoParser.cpp
```diff
@@ -1271,35 +1271,217 @@ class RelationPlanner : public AstVisitor {
...
-          VELOX_CHECK(item->is(NodeType::kSingleColumn));
-
-          const auto* singleColumn = item->as<SingleColumn>();
-          groupingKeys.emplace_back(toExpr(singleColumn->expression()));
-        } else {
-          groupingKeys.emplace_back(toExpr(expr));
+    if (useGroupingSets) {
+      groupingSets = expandGroupingSets(groupingElements, selectItems);
+      // Extract unique grouping keys for later use in output mapping.
+      std::unordered_set<
```
mbasmanova: Use ExprSet/ExprMap. Are we deduplicating twice? One here and once again in PlanBuilder? Is this intentional?

axiom/logical_plan/PlanBuilder.h
```diff
@@ -259,6 +259,71 @@ class PlanBuilder {
       const std::vector<ExprApi>& aggregates,
       const std::vector<AggregateOptions>& options);
 
+  /// Aggregate with grouping sets support for ROLLUP, CUBE, and GROUPING SETS.
+  /// @param groupingSets Vector of grouping sets, where each set contains
+  /// the grouping key expressions for that set. The function extracts unique
```
mbasmanova: Let's also clarify:

```
/// Grouping keys appear in output in first-occurrence order.
/// For explicit ordering, use the overload with separate groupingKeys and index-based sets.
```

Let's change the order of methods to start with most flexible (though less convenient): ExprApi + index-based

axiom/logical_plan/PlanBuilder.h
```diff
@@ -259,6 +259,71 @@ class PlanBuilder {
...
+  PlanBuilder& aggregateWithGroupingSets(
+      const std::vector<ExprApi>& groupingKeys,
+      const std::vector<std::vector<int32_t>>& groupingSets,
+      const std::vector<ExprApi>& aggregates,
+      const std::vector<AggregateOptions>& options,
+      const std::string& groupingSetIndexName);
+
+  /// ROLLUP convenience API. Expands ROLLUP(a, b, c) to grouping sets:
+  /// [[0,1,2], [0,1], [0], []]
+  PlanBuilder& rollup(
```
mbasmanova: Is there test coverage for rollup and cube APIs? If not, let's add.

axiom/logical_plan/PlanBuilder.cpp
natashasehgal: Great idea, have a PR for this https://github.com/facebookincubator/velox/pull/16272

axiom/sql/presto/PrestoParser.cpp
```diff
@@ -1271,35 +1271,217 @@ class RelationPlanner : public AstVisitor {
...
+            for (auto& groupingSet : allSets) {
+              for (const auto& e : exprs) {
+                groupingSet.push_back(e);
+              }
+            }
+          }
+          break;
+        }
+
+        case NodeType::kRollup: {
```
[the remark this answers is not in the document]
natashasehgal: Thanks for the suggestion. I refactored further using a helper function

axiom/logical_plan/PlanBuilder.cpp
natashasehgal: Added a test and moved to PrestoParser

This limit comes from Presto code and is not documented currently. I created an issue for tracking which I linked in this usage: https://github.com/prestodb/presto/issues/27096 , I have a fix linked to the issue

There is no limit for rollup because it is linear. Cube is exponential

axiom/logical_plan/tests/PlanBuilderTest.cpp
```diff
@@ -148,5 +150,140 @@ TEST_F(PlanBuilderTest, setOperationTypeCoercion) {
...
+  std::vector<Variant> data{
+      Variant::row({1, 10, 100, 1000}),
+  };
+
+  // Create grouping sets using ordinals: (0, 1) and (0, 1, 2)
+  auto plan = PlanBuilder()
+                  .values(rowType, data)
+                  .aggregate(
+                      {"a", "b", "c"}, // grouping keys
+                      {{0, 1}, {0, 1, 2}}, // grouping sets using indices
```
mbasmanova: inconsistent terminology: ordinals vs. indices

I believe ordinals typically 1-based, while indices 0-based

axiom/logical_plan/tests/PlanBuilderTest.cpp
```diff
@@ -148,5 +150,140 @@ TEST_F(PlanBuilderTest, setOperationTypeCoercion) {
...
+  std::vector<Variant> data{
+      Variant::row({1, 10}),
+  };
+
+  VELOX_ASSERT_THROW(
+      PlanBuilder()
+          .values(rowType, data)
+          .aggregate({{}}, {}, "$grouping_set_id")
+          .build(),
+      "Aggregation node must specify at least one aggregate or grouping key");
```
mbasmanova: Would you also add a test that using out-of-bound indices ?

axiom/sql/presto/tests/PrestoParserTest.cpp
```diff
@@ -979,6 +979,217 @@ TEST_F(PrestoParserTest, simpleGroupBy) {
...
+      agg->groupingSets(),
+      testing::ElementsAre(
+          testing::ElementsAre(0), testing::ElementsAre(0, 1)));
+}
+
+TEST_F(PrestoParserTest, cubeColumnLimit) {
+  // CUBE is limited to 30 columns (2^30 grouping sets).
+  // Generate a query with 31 columns to verify the limit is enforced.
+  std::string columns;
+  std::string cubeColumns;
```
mbasmanova: columns and cubeColumns appear to be the same? Do we need both?

axiom/sql/presto/PrestoParser.cpp
```diff
@@ -1271,43 +1271,295 @@ class RelationPlanner : public AstVisitor {
...
-
-          ordinalExpressions.emplace(
-              singleColumn->expression().get(), groupingKeys.back());
-        } else {
-          groupingKeys.emplace_back(toExpr(expr));
+    if (useGroupingSets) {
+      groupingSets = expandGroupingSets(groupingElements, selectItems);
+      // Extract unique grouping keys and build index-based grouping sets.
+      // This avoids duplicate grouping keys, as deduplication is done in
+      // PlanBuilder.
```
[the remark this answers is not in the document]
natashasehgal: FYI the expr-based overload for PlanBuilder::aggregate() de-duplicates but the index-based overload does not. It trusts caller to provide de-duplicated keys

I added a defensive-check and test for it in latest version

velox::core::ExprMap<bool> seen;
for (const auto& key : groupingKeys) {
  VELOX_USER_CHECK(
      seen.try_emplace(key.expr(), true).second, "Duplicate grouping key");
}

mbasmanova (approved):
Thanks.

bigfootjon (approved):

PR #826, opened 2026-01-30
feat: Make getSplits call async

NikhilCollooru:
Summary: Make the getSplits call async via coroutines so that the thread is not blocked while splits are being created/generated.

Differential Revision: D91800596

### DIFF
--- a/axiom/runner/LocalRunner.h
+++ b/axiom/runner/LocalRunner.h
@@ -16,6 +16,8 @@
 
 #pragma once
 
+#include <folly/coro/AsyncScope.h>
+
 #include "axiom/connectors/ConnectorSplitManager.h"
 #include "axiom/runner/MultiFragmentPlan.h"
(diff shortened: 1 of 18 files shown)


axiom/runner/LocalRunner.h
```diff
@@ -86,6 +88,8 @@ class LocalRunner : public Runner,
           std::make_shared<ConnectorSplitSourceFactory>(),
       std::shared_ptr<velox::memory::MemoryPool> outputPool = nullptr);
 
+  ~LocalRunner();
```
MBkkt: This class have multiple data races and also wrong logic (e.g. waitForCompletion doesn't work).

I don't think it makes sense to introduce more parallelism in already broken code.

Probably better approach is fix this class first.

Also in practice it's important to be `next` call async, because it's commonly called from connection thread, and we don't want to block connection thread while query is executed

PR #833, opened 2026-01-31
fix: Finalize DT when scalar subquery used with aggregation

mbasmanova:
Summary:
When a scalar subquery is used in a projection and the query also has a global aggregation, we need to finalize the DT before adding the subquery. Otherwise, the scalar subquery would be placed in the same DT as the aggregation, but it needs to be cross-joined AFTER the aggregation.

Fixes queries like:

> SELECT (SELECT count(*) FROM region) c, array_agg(r_name) FROM region

Differential Revision: D91978687

### DIFF
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -2095,11 +2095,13 @@ void ToGraph::processSubqueries(
   Subqueries subqueries;
   extractSubqueries(expr, subqueries);
 
-  if (filter) {
-    if (currentDt_->hasLimit() ||
-        (currentDt_->hasAggregation() && !subqueries.empty())) {
-      finalizeDt(input);
-    }
+  if (filter && currentDt_->hasLimit()) {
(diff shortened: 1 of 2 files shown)


axiom/optimizer/ToGraph.cpp
```diff
@@ -2056,18 +2056,52 @@ ColumnCP ToGraph::addMarkColumn() {
...
-  if (filter) {
-    if (currentDt_->hasLimit() ||
-        (currentDt_->hasAggregation() && !subqueries.empty())) {
-      finalizeDt(input);
-    }
+  if (filter && currentDt_->hasLimit()) {
+    // Filter cannot be added after LIMIT.
+    finalizeDt(input);
+  } else if (currentDt_->hasAggregation() && !subqueries.empty()) {
+    // Scalar subqueries are placed as joins. Joins cannot be added after
```
srsuryadev: This can be made as a single if block

srsuryadev (approved):

PR #843, opened 2026-02-02
feat: Add support for localtime and localtimestamp functions

natashasehgal:
This change implements parsing of SQL special date/time functions `localtime` and `localtimestamp`. 

Also has test coverage for `current_time`.

E.g 
```sql
SELECT * FROM t WHERE time_col > localtime
```

Differential Revision: D92071638

### DIFF
--- a/axiom/sql/presto/ExpressionPlanner.cpp
+++ b/axiom/sql/presto/ExpressionPlanner.cpp
@@ -732,9 +732,9 @@ lp::ExprApi ExpressionPlanner::toExpr(
         case CurrentTime::Function::kTimestamp:
           return lp::Call("current_timestamp");
         case CurrentTime::Function::kLocaltime:
-          VELOX_NYI("LOCALTIME is not supported yet.");
+          return lp::Call("localtime");
         case CurrentTime::Function::kLocaltimestamp:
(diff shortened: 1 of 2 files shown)


axiom/sql/presto/PrestoParser.cpp
```diff
@@ -542,6 +542,22 @@ class RelationPlanner : public AstVisitor {
...
+      case NodeType::kCurrentTime: {
+        auto* currentTime = node->as<CurrentTime>();
+        switch (currentTime->function()) {
+          case CurrentTime::Function::kDate:
+            return lp::Call("current_date");
+          case CurrentTime::Function::kTimestamp:
+            return lp::Call("current_timestamp");
+          case CurrentTime::Function::kLocaltime:
+            return lp::Call("localtime");
+          default:
```
natashasehgal: Thoughts on adding `current_time` and `localtimestamp` in Velox? This way we can support current time functions `kTime` and `kLocaltimestamp` either in the same PR or a followup. 

For now it is falling through to `VELOX_UNSUPPORTED`. I can update default: to explicit case statements that fall through

cc @mbasmanova

natashasehgal: @mbasmanova updated now that it is available on Velox

PR #855, opened 2026-02-04
refactor: Centralize DerivedTable initialization in initializePlans

mbasmanova:
Summary:
Introduce a new `initializePlans()` method that handles all DerivedTable initialization in a single traversal. Previously, initialization was scattered across `Optimization` constructor and `ToGraph` methods, with `makeInitialPlan()` called separately for each DerivedTable as it was created.

The new approach:
- Performs top-down `distributeConjuncts()` to push filters down the tree
- Performs bottom-up `finalizeJoins()` (which combines addImpliedJoins, linkTablesToJoins, setStartTables, and guessFanout) and memoization
- Handles UNION/UNION ALL branches by aggregating cardinality and column statistics from children

This centralizes the initialization logic, making it easier to reason about the order of operations and ensuring all DerivedTables are properly initialized before optimization begins.

Differential Revision: D92290059

### DIFF
--- a/axiom/optimizer/DerivedTable.cpp
+++ b/axiom/optimizer/DerivedTable.cpp
@@ -75,7 +75,87 @@ void fillJoins(
     }
   }
 }
+
+MemoKey memoKey(const DerivedTable& dt) {
+  return MemoKey::create(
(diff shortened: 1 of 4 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

PR #861, opened 2026-02-06
feat: Add support for non-equi correlated scalar subqueries

mbasmanova:
Summary:
Add support for decorrelating scalar subqueries with non-equi correlation conditions (e.g., `n_regionkey < r_regionkey`).

The decorrelation strategy for non-equi conditions uses:
1. **AssignUniqueId** - Assigns a unique identifier to each outer row
2. **LEFT JOIN** - Joins outer with inner using the non-equi predicate as a filter
3. **Streaming Aggregation** - Groups by the unique ID to compute the aggregate per outer row

Differential Revision: D92435311

### DIFF
--- a/axiom/optimizer/FunctionRegistry.cpp
+++ b/axiom/optimizer/FunctionRegistry.cpp
@@ -73,6 +73,24 @@ bool FunctionRegistry::registerCardinality(std::string_view name) {
   return true;
 }
 
+bool FunctionRegistry::registerArbitrary(std::string_view name) {
+  VELOX_USER_CHECK(!name.empty());
+  if (arbitrary_.has_value() && arbitrary_.value() != name) {
+    return false;
+  }
+  arbitrary_ = name;
+  return true;
+}
+
+bool FunctionRegistry::registerCount(std::string_view name) {
+  VELOX_USER_CHECK(!name.empty());
+  if (count_.has_value() && count_.value() != name) {
+    return false;
+  }
+  count_ = name;
+  return true;
+}
+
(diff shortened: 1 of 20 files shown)


axiom/optimizer/tests/PlanMatcher.h
```diff
@@ -342,6 +354,16 @@ class PlanMatcherBuilder {
...
 
+  /// Matches an AssignUniqueId node, which assigns unique identifiers to
+  /// each input row. Used for decorrelating subqueries with non-equi
+  /// correlation conditions.
+  PlanMatcherBuilder& assignUniqueId();
+
+  /// Matches an AssignUniqueId node and captures the unique ID column name
+  /// as a symbol alias for use in parent matchers via symbol rewriting.
+  /// @param alias The alias to use for the unique ID column.
+  PlanMatcherBuilder& assignUniqueId(const std::string& alias);
```
MBkkt: string_view

axiom/optimizer/FunctionRegistry.cpp
```diff
@@ -211,6 +229,8 @@ void FunctionRegistry::registerPrestoFunctions(std::string_view prefix) {
   registry->registerElementAt(fullName("element_at"));
   registry->registerSubscript(fullName("subscript"));
   registry->registerCardinality(fullName("cardinality"));
+  registry->registerArbitrary(fullName("arbitrary"));
+  registry->registerCount(fullName("count"));
```
MBkkt: Also min, max needed, because hack for min/max with 2 arguments

PR #864, opened 2026-02-06
fix(cli): Add unregistering metadata to Connectors destructor

jagill:
Summary:
Not destroying them caused a use-after-free error, as the
HiveConnectorMetadata held on to a raw pointer to HiveConnector and
HiveConnector was destroyed before HiveConnectorMetadata.

Differential Revision: D92519163


pedroerp (approved):
Thank you

PR #865, opened 2026-02-06
feat(optimizer): Support ORDER BY in aggregates for distributed mode

kagamiori:
Summary:
Previously, aggregation functions with ORDER BY keys (e.g., `array_agg(x ORDER BY y)`) would throw an error 
when running in distributed mode (multiple workers or drivers).

This diff removes the hard check in `ToGraph::translateAggregation()` that blocked ORDER BY aggregates in 
distributed mode, and instead enforces single-step aggregation in `Optimization::addAggregation()` when any 
aggregate has ORDER BY keys. This follows the same strategy as Presto-Java: https://github.com/prestodb/presto/blob/c3b3eeb6260a44753dd917ec129ff52a248effe6/presto-main-base/src/main/java/com/facebook/presto/operator/aggregation/AggregationUtils.java#L59

This diff also enforce the optimizer to generate single-step aggregation when it has ORDER BY, even when OptimizerOption::alwaysPlanPartialAggregation == true.

**Why take this approach?**
Below are what other engines do with aggregation over sorted input.
* Presto-Java: Disallow aggregation over sorted input to be decomposed into partial + final steps.
* Spark: Spark allows aggregation over sorted input to be decomposed into partial + final steps, at the price of 
having each individual aggregation functions handle the ordering of inputs themselves. [In Spark-4.1.0, only 5 aggregations allow sorted input](https://spark.apache.org/docs/latest/sql-ref-functions-builtin.html#aggregate-functions).  
These functions are aware of and handle aggregation over sorted input by keeping the order-by value in 
companion with input values in the accumulator of every group. Functions that support both sorted and unsorted 
inputs require different code paths for the two cases. In contrast, Velox supports aggregation over sorted input at 
operator level and doesn't require individual functions to handle the ordering of inputs specially. This setup 
consequently requires the aggregation over sorted input after inputs have been shuffled by the grouping keys.
* DuckDB-1.3.1: DuckDB is for single-node with shared-memory among threads. DuckDB allows aggregation over 
sorted input to be decomposed, but the approach is specific to DuckDB's implementation of aggregation: Each 
thread first keeps all its input rows in a buffer during the first stage. After that finishes, each thread accesses all 
buffered rows of one grouping key, sorting and aggregating them. This approach would be similar to the single-
step aggregation approach in distributed execution mode.

Differential Revision: D92352633

### DIFF
--- a/axiom/optimizer/Optimization.cpp
+++ b/axiom/optimizer/Optimization.cpp
@@ -1002,6 +1002,71 @@ ExprVector computePreGroupedKeys(
                                                         : fromClusterKeys;
 }
 
+// Creates a two-phase (partial + final) aggregation plan for distributed
+// execution. First applies a partial aggregation on local data, then
+// repartitions by grouping keys and performs the final aggregation. Returns a
+// pair of the root to the aggregation plan and cost of this aggregation.
+std::pair<Aggregation*, PlanCost> makeSplitAggregationPlan(
+    const AggregationPlan* aggPlan,
+    RelationOpPtr plan,
+    const ExprVector& groupingKeys,
+    const AggregateVector& aggregates) {
+  PlanCost splitAggCost;
+  Aggregation* splitAggPlan;
+
+  auto* partialAgg = make<Aggregation>(
+      plan,
+      groupingKeys,
+      /*preGroupedKeys*/ ExprVector{},
+      aggregates,
+      velox::core::AggregationNode::Step::kPartial,
+      aggPlan->intermediateColumns());
+
+  splitAggCost.add(*partialAgg);
+  plan = repartitionForAgg(aggPlan, partialAgg, splitAggCost);
+
+  const auto numKeys = aggPlan->groupingKeys().size();
+
+  ExprVector finalGroupingKeys;
+  finalGroupingKeys.reserve(numKeys);
+  for (auto i = 0; i < numKeys; ++i) {
+    finalGroupingKeys.push_back(aggPlan->intermediateColumns()[i]);
+  }
+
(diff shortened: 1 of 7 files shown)


axiom/optimizer/Optimization.cpp
```diff
@@ -1054,6 +1111,34 @@ void Optimization::addAggregation(
...
+    const auto& [splitAggPlan, splitAggCost] =
+        makeSplitAggregationPlan(aggPlan, plan, groupingKeys, aggregates);
+    plan = splitAggPlan;
+    state.cost.add(splitAggCost);
+    return;
+  }
+
+  // Check if any aggregate has ORDER BY keys. If so, we must use single-step
+  // aggregation because partial aggregation cannot preserve global ordering.
+  auto hasOrderByAggregate = [&aggregates]() {
```
mbasmanova: nit: std::any_of

axiom/optimizer/tests/TpchPlanTest.cpp
```diff
@@ -705,7 +705,8 @@ TEST_F(TpchPlanTest, DISABLED_makePlans) {
     };
 
     auto logicalPlan = parseTpchSql(q);
-    planVelox(logicalPlan, options, fmt::format("{}/q{}", path, q));
+    planVelox(
+        logicalPlan, options, std::nullopt, fmt::format("{}/q{}", path, q));
```
mbasmanova: add parameter name for std::nullopt for clarity

/*optimizerOptions=*/std::nullopt

mbasmanova (approved):
Thanks.

PR #868, opened 2026-02-07
feat: Add support for correlated scalar subqueries without aggregation

mbasmanova:
Summary:
Add support for correlated scalar subqueries that don't contain aggregation functions. Previously, correlated scalar subqueries required aggregation (e.g., `SELECT sum(x) FROM u WHERE u.id = t.id`), which naturally guarantees a single row per correlation group.

For subqueries without aggregation (e.g., `SELECT c FROM u WHERE d = b`), we now use the EnforceDistinct operator to validate single-row semantics at runtime. The approach is:
1. Add AssignUniqueId to the outer query to generate a row number column
2. Create a LEFT JOIN with the subquery
3. Add EnforceDistinct on the row number column to detect and fail if multiple rows match

The EnforceDistinct operator throws a user-friendly error "Scalar sub-query has returned multiple rows" if a correlated subquery returns more than one row for any correlation group.

Differential Revision: D92610178

### DIFF
--- a/axiom/logical_plan/PlanBuilder.h
+++ b/axiom/logical_plan/PlanBuilder.h
@@ -309,7 +309,7 @@ class PlanBuilder {
       JoinType joinType);
 
   PlanBuilder& crossJoin(const PlanBuilder& right) {
-    return join(right, /* condition */ "", JoinType::kInner);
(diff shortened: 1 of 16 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

PR #870, opened 2026-02-07
feat: Add suport for parsing of POSITION

natashasehgal:
Summary:
Add support for parsing on position

POSITION(x IN Y) ->  strpos(Y, X)

Reference: https://prestodb.io/docs/current/functions/string.html#strpos-string-substring-instance-bigint

Differential Revision: D92632138


mbasmanova (approved):
Thanks.

PR #876, opened 2026-02-09
fix: EXPLAIN DROP TABLE

HeidiHan0000:
Summary:

```
explain drop table some_table;
```
fails with 
`Unsupported statement type: DropTable - Native Coordinator Query Failed
`

This is because it hit the explain branch before it reached the kDropTable branch in doParse, and doPlan does not handle kDropTable.

fix: move the handling of kDropTable from doParse -> doPlan

Differential Revision: D92474492

### DIFF
--- a/axiom/sql/presto/tests/PrestoParserTest.cpp
+++ b/axiom/sql/presto/tests/PrestoParserTest.cpp
@@ -775,6 +775,24 @@ TEST_F(PrestoParserTest, explainInsert) {
   }
 }
 
+TEST_F(PrestoParserTest, explainCreateTable) {
+  testExplainDdl(
+      "EXPLAIN CREATE TABLE t (id INTEGER)", SqlStatementKind::kCreateTable);
+  testExplainDdl(
+      "EXPLAIN CREATE TABLE IF NOT EXISTS t (id INTEGER)",
+      SqlStatementKind::kCreateTable);
+  testExplainDdl(
+      "EXPLAIN CREATE TABLE t (id INTEGER, ds VARCHAR) "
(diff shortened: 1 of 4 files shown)


axiom/sql/presto/SqlStatement.h
```diff
@@ -175,6 +176,12 @@ class CreateTableStatement : public SqlStatement {
     return tableName_;
   }
 
+  /// Returns the table name exactly as specified in the SQL statement.
+  /// This preserves the original format (1-part, 2-part, or 3-part name).
+  const std::string& originalTableName() const {
```
mbasmanova: Do we need the same API for CTAS?

Would you update PR description to explain why we need this change?

HeidiHan0000: Removed, as per offline discussion

axiom/sql/presto/tests/PrestoParserTest.cpp
```diff
@@ -120,6 +120,19 @@ class PrestoParserTest : public testing::Test {
     }
   }
 
+  /// Verifies that EXPLAIN can parse DDL statements (CREATE TABLE, DROP TABLE).
+  /// DDL statements do not have logical plans, so we only verify parsing.
+  void testExplainDdl(std::string_view sql, SqlStatementKind expectedKind) {
+    SCOPED_TRACE(sql);
+    auto parser = makeParser();
+
+    auto statement = parser.parse(sql);
```
mbasmanova: use parseSql

mbasmanova (approved):
Thanks.

PR #881, opened 2026-02-10
fix(optimizer): Fix repartitionForAgg() when child is partitioned by a supset of grouping keys

kagamiori:
Summary:
repartitionForAgg() used to incorrectly skip shuffles when the child plan is already partitioned by a superset of the required grouping keys. This could cause incorrect result, however, because rows with the same values of required grouping keys can be on different workers if they have different values at the rest partitioning columns of the child plan. E.g., suppose the child plan is already parrtitioned by column `c0`, `c1`, while the parent aggregation requires grouping by column `c0`, then two rows `(c0=0, c1=1)` and `(c0=0, c1=2)` may be on different workers and hence need to be shuffled to the same worker by `c0` before the parent aggregation.

This diff fixes this bug by reversing the condition when the shuffle can be omitted, i.e., if the parent aggregation requires grouping by a superset of keys that the child plan is already partitioned on, it is guaranteed that rows with the same values of grouping keys are already on the same worker, so the shuffle can be omitted.

Differential Revision: D92882000

### DIFF
--- a/axiom/optimizer/Optimization.cpp
+++ b/axiom/optimizer/Optimization.cpp
@@ -463,7 +463,6 @@ RelationOpPtr repartitionForAgg(
     AggregationPlanCP const agg,
     const RelationOpPtr& plan,
     PlanCost& cost) {
-  // No shuffle if all grouping keys are in partitioning.
   if (isSingleWorker() || plan->distribution().isGather()) {
     return plan;
   }
@@ -484,13 +483,19 @@ RelationOpPtr repartitionForAgg(
     keyValues.push_back(agg->intermediateColumns()[i]);
   }
 
-  bool shuffle = false;
-  for (auto& key : keyValues) {
-    auto nthKey = position(plan->distribution().partitionKeys(), *key);
-    if (nthKey == kNotFound) {
-      shuffle = true;
-      break;
-    }
+  // Check if grouping keys is a superset of all partition keys. If partition
+  // keys are empty or contain columns not in grouping keys, we need to shuffle.
+  // Empty partition keys means unknown/unpartitioned distribution, so shuffle
+  // is required to ensure rows with same grouping keys are co-located.
+  bool shuffle = plan->distribution().partitionKeys().empty();
+  if (!shuffle) {
+    const auto& partitionKeys = plan->distribution().partitionKeys();
+    shuffle = std::any_of(
+        partitionKeys.begin(),
+        partitionKeys.end(),
+        [&keyValues](const auto& key) {
+          return position(keyValues, *key) == kNotFound;
+        });
   }
   if (!shuffle) {
     return plan;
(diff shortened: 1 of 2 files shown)


axiom/optimizer/Optimization.cpp
```diff
@@ -484,12 +483,17 @@ RelationOpPtr repartitionForAgg(
...
-    if (nthKey == kNotFound) {
-      shuffle = true;
-      break;
+  // Check if grouping keys is a superset of all partition keys. If partition
+  // keys are empty or contain columns not in grouping keys, we need to shuffle.
+  // Empty partition keys means unknown/unpartitioned distribution, so shuffle
+  // is required to ensure rows with same grouping keys are co-located.
+  bool shuffle = plan->distribution().partitionKeys().empty();
+  if (!shuffle) {
+    for (auto& partKey : plan->distribution().partitionKeys()) {
```
mbasmanova: nit: std::any_of

axiom/optimizer/tests/AggregationPlanTest.cpp
```diff
@@ -271,5 +271,79 @@ TEST_F(AggregationPlanTest, orderBy) {
...
+    testConnector_->dropTableIfExists("t");
+  };
+
+  // Test case 1: current partitionKeys ⊆ required groupingKeys --> no shuffle
+  // needed.
+  {
+    auto logicalPlan = lp::PlanBuilder()
+                           .tableScan(kTestConnectorId, "t")
+                           .aggregate({"a", "b"}, {})
+                           .project({"a", "b", "a + b as d"})
```
mbasmanova: nit: with({"a + b as d"})

axiom/optimizer/tests/AggregationPlanTest.cpp
```diff
@@ -271,5 +271,79 @@ TEST_F(AggregationPlanTest, orderBy) {
...
+                       .project()
+                       // No shuffle here - partitionKeys ⊆ groupingKeys
+                       .localPartition()
+                       .singleAggregation({"a", "b", "d"}, {})
+                       .shuffle()
+                       .build();
+    AXIOM_ASSERT_DISTRIBUTED_PLAN(plan.plan, matcher);
+  }
+
+  // Test case 2: current partitionKeys ⊄ required groupingKeys --> shuffle is
```
mbasmanova: > // Test case 2

Drop numbers from comments. These are hard to maintain when new test cases are adding, existing removed or moved around.

axiom/optimizer/tests/AggregationPlanTest.cpp
```diff
@@ -271,5 +271,79 @@ TEST_F(AggregationPlanTest, orderBy) {
...
+  // needed.
+  {
+    auto logicalPlan = lp::PlanBuilder()
+                           .tableScan(kTestConnectorId, "t")
+                           .aggregate({"a", "b", "c"}, {})
+                           .aggregate({"a", "b"}, {})
```
mbasmanova: Should the optimizer keep only one aggregation (a, b)? Maybe create a separate GitHub issue to optimize that.

kagamiori: Created https://github.com/facebookincubator/axiom/issues/886.

PR #883, opened 2026-02-11
docs: Add coding style rules for LLM

mbasmanova:
Summary:
Add `.llms/rules/coding-style.md` to capture Axiom-specific coding conventions that complement the Velox coding style guide. These rules are automatically loaded by LLM to ensure consistent code suggestions.

Key additions:
- Naming conventions: Avoid abbreviations, prefer full descriptive names (e.g., `outputColumn` not `outputCol`)
- Exceptions for well-established abbreviations: `id`, `url`, `sql`, `expr`, `it`, `numXxx`
- Comment style: Use `///` only for public API docs (Doxygen), `//` for private/internal comments
- Active verbs in comments: "Builds query plans" not "This class builds query plans"
- Avoid redundant comments that repeat what the code already says

Differential Revision: D92964619

### DIFF
--- a/axiom/.llms/rules/coding-style.md
+++ b/axiom/.llms/rules/coding-style.md
@@ -1,0 +1,110 @@
+---
+oncalls: ['velox']
+apply_to_regex: 'fbcode/axiom/.*\.(cpp|h|md)'
+---
+
+# Axiom Coding Style
+
+Axiom follows the Velox coding style. See `CODING_STYLE.md` in the Velox repository for the complete guide:
+- Meta-internal: `fbcode/velox/public_tld/CODING_STYLE.md`
+- OSS: `velox/CODING_STYLE.md`
+
+The rules below are additions or clarifications specific to Axiom.
+
+## Naming Conventions
+
+- **Do not abbreviate** variable, function, or class names. Use full, descriptive names.
+- Clarity is more important than brevity.
+
+### Examples
+
+| ❌ Avoid | ✅ Prefer |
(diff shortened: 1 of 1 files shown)


axiom/.llms/rules/coding-style.md
```diff
@@ -0,0 +1,105 @@
...
+| `sel` | `selectivity` |
+| `rowCount` | `numRows` |
+
+### Exceptions
+
+- Well-established abbreviations in the domain are acceptable (e.g., `id`, `url`, `sql`, `expr`).
+- Loop indices like `i`, `j`, `k` are acceptable for simple loops.
+- Iterator variables named `it` are acceptable.
+- `numXxx` naming pattern is acceptable and preferred over `xxxCount` or `xxxCnt` (e.g., `numRows`, `numKeys`).
+- Type aliases defined elsewhere (e.g., `ExprCP`, `ColumnCP`) should be used as-is.
```
MBkkt: Will be nice to say something about this:
`const auto* expr = ...;` vs `ExprCP expr = ...` vs `auto expr = ...;` ...

Because these usings for raw pointers, is quite non-standard thing in modern C++ projects, so it's not clear how to deal with them better.

Also a lot of place in code right now, because of these usings are non-null raw pointers (they cannot be null because it will be crash/check/dcheck in such case).
Maybe it's better to don't have aliases for raw reference/pointers and try to avoid them?
It will make code safer and easier to write/read and check for both, devs and machine.

PR #885, opened 2026-02-11
feat(optimizer): Support DISTINCT aggregation with single set of distinct arguments

kagamiori:
Summary:
Previously, DISTINCT aggregation (e.g., count(DISTINCT b)) was only supported in single worker, single thread mode and no ORDER BY in the same expression.

This diff adds support for DISTINCT aggregation in distributed mode by transforming it into a two-level aggregation: the inner level performs GROUP BY (grouping_keys + distinct_args) to deduplicate rows, and the outer level applies the aggregation functions without the DISTINCT flag. For example, `SELECT a, count(DISTINCT b), sum(DISTINCT b) FROM t GROUP BY a` becomes an inner `GROUP BY (a, b)` followed by an outer `SELECT a, count(b), sum(b) GROUP BY a`. In this diff, ORDER BY remains disallowed with DISTINCT. A follow-up diff will add support for it.

The transformation is not supported if there exists
1. any aggregate is non-DISTINCT,
2. any aggregate has a filter condition,
3. aggregates have different sets of distinct arguments, e.g., `agg(a, b), agg(b, a, b)` is allowed, but `agg(a, b), agg(a, a)` and `agg(a, b), agg(c, d)` are not allowed.
4. any aggregate has ORDER BY.

Differential Revision: D92920128

### DIFF
--- a/axiom/optimizer/Optimization.cpp
+++ b/axiom/optimizer/Optimization.cpp
@@ -558,8 +558,9 @@ bool isSingleWorker() {
 }
 
 RelationOpPtr repartitionForAgg(
-    AggregationPlanCP const agg,
     const RelationOpPtr& plan,
+    const ExprVector& groupingKeys,
+    const ColumnVector& intermediateColumns,
     PlanCost& cost) {
   if (isSingleWorker() || plan->distribution().isGather()) {
     return plan;
@@ -567,7 +568,7 @@ RelationOpPtr repartitionForAgg(
 
   // If no grouping and not yet gathered on a single node,
   // add a gather before final agg.
-  if (agg->groupingKeys().empty()) {
+  if (groupingKeys.empty()) {
     auto* gather =
         make<Repartition>(plan, Distribution::gather(), plan->columns());
     cost.add(*gather);
@@ -577,8 +578,9 @@ RelationOpPtr repartitionForAgg(
   // 'intermediateColumns' contains grouping keys followed by partial agg
   // results.
   ExprVector keyValues;
-  for (auto i = 0; i < agg->groupingKeys().size(); ++i) {
-    keyValues.push_back(agg->intermediateColumns()[i]);
+  keyValues.reserve(groupingKeys.size());
+  for (auto i = 0; i < groupingKeys.size(); ++i) {
+    keyValues.push_back(intermediateColumns[i]);
   }
 
   // Check if grouping keys is a superset of all partition keys. If partition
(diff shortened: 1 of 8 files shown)


axiom/optimizer/Optimization.cpp
```diff
@@ -1093,6 +1126,78 @@ RelationOpPtr Optimization::planSingleAggregation(
...
+        }
+      }
+    }
+  }
+
+  return commonArgs;
+}
+
+// Returns a copy of aggregates with isDistinct set to false.
+AggregateVector removeDistinctFromAggregation(
```
mbasmanova: nit: perhaps, add a method on Aggregate: 

AggregateCP Aggregate::dropDistinct();

axiom/optimizer/tests/AggregationPlanTest.cpp
```diff
@@ -342,5 +343,145 @@ TEST_F(AggregationPlanTest, repartitionForAggPartitionSubset) {
...
+
+    auto plan = test::QueryTestBase::planVelox(logicalPlan);
+
+    auto expectedMatcher =
+        core::PlanMatcherBuilder()
+            .tableScan()
+            .project()
+            .shuffle()
+            .localPartition()
+            .singleAggregation({"expr", "\"dt1.__p11\""}, {})
```
mbasmanova: Do not use internal optimizer names (dt1.__p11). These are not stable. Check out "Symbol Rewriting" in fbcode/axiom/optimizer/tests/PlanMatcher.h

axiom/optimizer/Optimization.cpp
```diff
@@ -1142,6 +1176,61 @@ RelationOpPtr Optimization::planSingleAggregation(
...
+            "DISTINCT aggregates have multiple sets of arguments");
+      }
+    }
+  }
+
+  return commonArgs;
+}
+
+// Returns a copy of aggregates with isDistinct set to false.
+AggregateVector removeDistinctFromAggregation(
```
mbasmanova: removeDistinctFromAggregation → consider dropDistinctFromAggregates to match the dropDistinct() method name on Aggregate.

axiom/optimizer/tests/AggregationPlanTest.cpp
```diff
@@ -342,5 +343,171 @@ TEST_F(AggregationPlanTest, repartitionForAggPartitionSubset) {
...
+            .singleAggregation(innerGroupingKeys, {})
+            .shuffle()
+            .localPartition()
+            .singleAggregation(outerGroupingKeys, aggregates)
+            .shuffle()
+            .build();
+      };
+
+  {
+    // Test multiple DISTINCT aggregates on the same set of columns.
```
mbasmanova: The tests cover the positive cases well. A few additional scenarios to consider:

- No grouping keys with DISTINCT: SELECT count(DISTINCT b) FROM t (no GROUP BY). The inner GROUP BY would be just (b) and the outer GROUP BY would be empty. This is an important edge case for the repartitionForAgg gather path.
- Single aggregate: SELECT a, count(DISTINCT b) FROM t GROUP BY a — a simpler case that exercises the common path.

axiom/optimizer/Optimization.h
```diff
@@ -236,6 +236,17 @@ class Optimization {
   void addAggregation(DerivedTableCP dt, RelationOpPtr& plan, PlanState& state)
       const;
 
+  // Transforms aggregations where all aggregates are DISTINCT with the same
+  // args into a two-level aggregation: inner level GROUP BY (keys +
+  // distinct_args) -> outer level AGG without DISTINCT.
+  void applySingleDistinctTransformation(
```
mbasmanova: perhaps, transformDistinctToGroupBy

PR #923, opened 2026-02-19
fix(parser): Correlated subqueries in JOIN ON clauses can now reference right-side columns

mbasmanova:
Summary:
When a JOIN ON clause contained a correlated subquery referencing columns from the right side of the join, the parser threw Cannot resolve column. 

This was because toExpr processed the ON clause using only the left builder's scope — subqueries could only resolve correlated references against left-side columns.

Root cause in PrestoParser.cpp processFrom:
  1. Left side built → leftBuilder has left columns
  2. Right side built → rightBuilder has right columns
  3. builder_ = leftBuilder — ON clause processed with only left-side scope
  4. When toExpr encountered a subquery, it captured scope from builder_ (left only)
  5. Correlated refs to right-side columns (e.g., r.r_regionkey) failed

Fix: After building both sides, create a combined scope lambda that tries the left scope first, then the right scope. Create a temporary builder with this combined scope for processing the ON clause. This allows subqueries to resolve correlated references from either side.

Differential Revision: D93757619

### DIFF
--- a/axiom/sql/presto/PrestoParser.cpp
+++ b/axiom/sql/presto/PrestoParser.cpp
@@ -1153,14 +1153,28 @@ class RelationPlanner : public AstVisitor {
 
       auto leftBuilder = builder_;
 
-      lp::PlanBuilder::Scope scope;
-      leftBuilder->captureScope(scope);
+      lp::PlanBuilder::Scope leftScope;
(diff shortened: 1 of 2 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

PR #956, opened 2026-02-24
fix: De-flake LocalRunnerTest.error

mbasmanova:
Summary:
  Fix a race condition in LocalRunner::abort() that causes a heap-use-after-free detected by ASAN under stress testing.

  The bug: abort() iterates over stages_ to call Task::setError() on each task, but waitForCompletion() running on another thread can concurrently clear stages_, freeing
  the tasks while abort() still holds references to them.

  The fix: take a local copy of the task shared pointers under the mutex before iterating, so the tasks remain alive regardless of what waitForCompletion() does to
  stages_.

Differential Revision: D94276142

### DIFF
--- a/axiom/runner/LocalRunner.cpp
+++ b/axiom/runner/LocalRunner.cpp
@@ -261,12 +261,21 @@ void LocalRunner::abort() {
     }
   }
   VELOX_CHECK(state_ != State::kInitialized);
-  // Setting errors is thread safe. The stages do not change after
-  // initialization.
-  for (auto& stage : stages_) {
-    for (auto& task : stage) {
-      task->setError(error_);
+
+  // Take a local copy of tasks under the mutex to prevent use-after-free if
+  // waitForCompletion() clears stages_ concurrently.
+  std::vector<std::shared_ptr<velox::exec::Task>> tasks;
+  {
+    std::lock_guard<std::mutex> l(mutex_);
+    for (auto& stage : stages_) {
+      for (auto& task : stage) {
+        tasks.push_back(task);
+      }
     }
+  }
+
+  for (auto& task : tasks) {
+    task->setError(error_);
   }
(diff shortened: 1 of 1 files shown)


axiom/runner/LocalRunner.cpp
```diff
@@ -261,13 +261,22 @@ void LocalRunner::abort() {
...
-      task->setError(error_);
+
+  // Take a local copy of tasks under the mutex to prevent use-after-free if
+  // waitForCompletion() clears stages_ concurrently.
+  std::vector<std::shared_ptr<velox::exec::Task>> tasks;
+  {
+    std::lock_guard<std::mutex> l(mutex_);
+    for (auto& stage : stages_) {
+      for (auto& task : stage) {
+        tasks.push_back(task);
```
hdikeman: setError seems fairly lightweight, assuming the onError_ callback is not set. Is it worth making copies or should we just set the errors here, under the lock?

mbasmanova: It doesn't see lightweight to me. Also, it may change w/o notice. Am I missing anything?

```
void Task::setError(const std::exception_ptr& exception) {
  TestValue::adjust("facebook::velox::exec::Task::setError", this);
  {
    std::lock_guard<std::timed_mutex> l(mutex_);
    if (not isRunningLocked()) {
      return;
    }
    if (exception_ != nullptr) {
      return;
    }
    exception_ = exception;
  }
  terminate(TaskState::kFailed);
  if (onError_) {
    onError_(exception_);
  }
}

void Task::setError(const std::string& message) {
  // The only way to acquire an std::exception_ptr is via throw and
  // std::current_exception().
  try {
    throw std::runtime_error(message);
  } catch (const std::runtime_error&) {
    setError(std::current_exception());
  }
}
```

PR #966, opened 2026-02-26
feat(optimizer): Optimize row_number <= N filters

mbasmanova:
Summary:
Optimize queries that filter on ranking function output by absorbing the filter
predicate into the ranking operator as a limit, avoiding full window computation.

Supported query shapes:

```sql
-- row_number/rank/dense_rank with upper-bound filter → TopNRowNumber
SELECT * FROM (
  SELECT *, row_number() OVER (ORDER BY b) as rn FROM t
) WHERE rn <= 5

-- row_number without ORDER BY + filter → RowNumber with limit
SELECT * FROM (
  SELECT *, row_number() OVER () as rn FROM t
) WHERE rn <= 5

-- Less-than and equality predicates
WHERE rn < 5   -- limit = 4
WHERE rn = 1   -- limit = 1
```

When additional predicates exist alongside the ranking predicate (e.g.
`rn <= 5 AND a > 10`), the ranking predicate is absorbed into the limit and the
remaining predicates are preserved.

Detection: since a filter on a window function output triggers a DT boundary, the ranking predicate starts in the outer DT. During filter pushdown, addFilter validates the predicate using `isRankingUpperBoundPredicate` and pushes it into the inner DT's conjuncts. In addWindow, `tryExtractRankingLimit` extracts the limit from the conjunct and passes it to `makeWindowOp`, which creates TopNRowNumber or RowNumber with the limit.

Differential Revision: D94566040

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -124,6 +124,23 @@ for the complete guide. Key rules are summarized below.
 - Keep method implementations in `.cpp` except for trivial one-liners.
 - Avoid default arguments when all callers can pass values explicitly.
 
+## Commit Messages
+
+Use conventional commit prefixes with `[Project]` tags:
+
+```
+[Axiom] feat(optimizer): Add support for window functions
+[Axiom] fix: Validate HAVING column references in SQL parser
+[Axiom] refactor(parser): Extract GroupByPlanner class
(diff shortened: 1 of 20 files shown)


xiaoxmeng (approved):
@mbasmanova thanks!

Issue #925, opened 2026-02-20
SELECT expressions lose mappings


peterenescu:
Queries that use **expressions** (not simple column references) (i.e. `col + 1`) in both the `SELECT` and `ORDER BY` crash due to unfound mappings.

## Reproduction

**Failing query:**
```sql
SELECT a + b FROM ( VALUES (1, 2) ) t(a, b) ORDER BY a + b;
```

**Succeeding query:**
```sql
SELECT a FROM ( VALUES (1, 2) ) t(a, b) ORDER BY a;
```

## Expected Behavior

Both queries should execute successfully.

## Additional Cases to Verify

```sql
-- Expression with constants
SELECT a + 1 FROM ( VALUES (1) ) t(a) ORDER BY a + 1;

-- Multiple expressions
SELECT a + b, a - b FROM ( VALUES (1, 2) ) t(a, b) ORDER BY a + b;

-- Nested subquery with expressions
SELECT x FROM (SELECT a + b AS x FROM ( VALUES (1, 2) ) t(a, b)) ORDER BY x;

-- Expression in ORDER BY not in SELECT
SELECT a FROM ( VALUES (1, 2) ) t(a, b) ORDER BY a + b;
```

mbasmanova:
This is fixed now:

```
SQL> SELECT a + b FROM ( VALUES (1, 2) ) t(a, b) ORDER BY a + b;
----
expr
----
   3
(1 rows in 1 batches)
```

PR #986, opened 2026-03-02
feat(parser): Resolve ORDER BY expressions against projection outputs

peterenescu:
Summary:
When ORDER BY references an expression that also appears in SELECT (e.g. `SELECT a + b FROM t ORDER BY a + b`), the expression needs to resolve against the projected output. However, after `project()`, the result is stored under an auto-generated internal name (e.g. `expr_0`) not registered in `outputMapping_`, so `resolveScalarTypes()` couldn't match it.

This adds a `projectionExpressionMap_` to `PlanBuilder` that maps unresolved projection expressions to their output column names:

- In `project()`, non-trivial expressions are saved into the map.
- In `resolveScalarTypes()`, before leaf-level column resolution, the entire expression is checked against the map. If matched, an `InputReferenceExpr` pointing to the projection output is returned directly.

Resolves https://github.com/facebookincubator/axiom/issues/925

Differential Revision: D94931869

### DIFF
--- a/axiom/logical_plan/PlanBuilder.cpp
+++ b/axiom/logical_plan/PlanBuilder.cpp
@@ -727,6 +727,13 @@ PlanBuilder& PlanBuilder::project(const std::vector<ExprApi>& projections) {
 
   resolveProjections(projections, outputNames, exprs, *newOutputMapping);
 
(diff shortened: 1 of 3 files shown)


natashasehgal (approved):
LGTM

PR #987, opened 2026-03-03
feat: Allow generating plans optimized for remote or local query result consumption

arhimondr:
### Execution Modes Considered

**1. In process planning and execution on a single machine**

This model is similar to a model DuckDB employes. Plan is created, executed and results are returned all within the same process. No network communication involved.

**2. Planning in one process and execution by a single different process**

This model usually involves a server accepting a query, generating a plan and sending it for execution to a different process (machine).

This model can be employed by a deployed service with many coordinators accepting queries and many workers processing them. For queries determined to be "small" the service may decide to process the whole query on a single worker and streaming results back to the client. 

**3. Planning in one process and executing by multiple other processes**

This model can be employed for "larger" queries that require large scale, distributed processing by employing multiple physical machines.

### Performance Considerations

To minimize the performance impact of the results transfer back to the client the planner must allow two things:

1. For in process planning and execution avoid any form of serialization all together
2. For remote execution minimize the impact of serialization / deserialization by reducing the number of times it is performed.

The current implementation of adding a `gather` exchange when the number of workers is greater than 1 does not satisfy the second requirements. Due to the Exchange operator implementation adding a remote exchange forces the engine to serialize query results and deserialize them back on the receiver end. 

When the result set is large it can be costly. In case of a large data set it can be highly beneficial to ask the engine to serialize results in a format that a client can understand and stream them back using zero copy mechanisms. The streaming results back can be further optimized by asking the client to pull results directly from a select set of workers to avoid utilizing coordinator CPU and Network bandwidth altogether.

### Proposed Change

This chance introduces a new flag, `remoteOutput` that allows a user of the Axiom planner to chose the behavior desired. 

To generate plan for in process execution mode the user needs to set the `.numWorkers` to `1` and instruct the planner that no remote transfer of query results is necessary by setting the `.remoteOutput` flag to `false`.

To generate plan for single node, remote execution the user needs to set the `.numWorkers` to `1` and instruct the planner that remote transfer of query results is needed by setting the `.remoteOutput` flag to `true`.

To generate plan for distributed execution the user needs to set the `.numWorkers` to `X > 1` and instruct the planner that remote transfer of query results is needed by setting the `.remoteOutput` flag to `true`.

It is also allowed to generated distributed plans with the `.remoteOutput` flag set to `flase`. In this case the planner will add an additional Exchange only fragment if necessary (if the last fragment is not a single node fragment already).

Differential Revision: D94688349

### DIFF
--- a/axiom/optimizer/ToVelox.cpp
+++ b/axiom/optimizer/ToVelox.cpp
@@ -304,27 +304,31 @@ PlanAndStats ToVelox::toVeloxPlan(
   prediction_.clear();
   nodeHistory_.clear();
 
-  if (options_.numWorkers > 1) {
+  if (options_.numWorkers > 1 && !options_.remoteOutput) {
     plan = addGather(plan);
   }
 
-  runner::ExecutableFragment top;
+  runner::ExecutableFragment top = newFragment();
   std::vector<runner::ExecutableFragment> stages;
   top.fragment.planNode = makeFragment(plan, top, stages);
-
-  if (!outputNames.empty()) {
-    top.fragment.planNode =
-        addOutputRenames(top.fragment.planNode, outputNames);
-  }
-
   stages.push_back(std::move(top));
 
-  auto finishWrite = std::move(finishWrite_);
-  VELOX_DCHECK(!finishWrite_);
+  auto& rootPlanNode = stages.back().fragment.planNode;
+  if (!outputNames.empty()) {
+    rootPlanNode = addOutputRenames(rootPlanNode, outputNames);
+  }
 
   for (const auto& stage : stages) {
     velox::core::PlanConsistencyChecker::check(stage.fragment.planNode);
   }
+
+  if (options.remoteOutput) {
+    rootPlanNode = velox::core::PartitionedOutputNode::single(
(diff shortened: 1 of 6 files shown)


axiom/optimizer/ToVelox.cpp
```diff
@@ -300,16 +300,15 @@ PlanAndStats ToVelox::toVeloxPlan(
     RelationOpPtr plan,
     const runner::MultiFragmentPlan::Options& options,
     const std::vector<logical_plan::OutputNode::Entry>& outputNames) {
+  VELOX_CHECK(
+      options.numWorkers == 1 || options.remoteOutput,
+      "Remote output is required for multi-worker plans");
```
mbasmanova: Should we document this requirement in remoteOutput's option? Or, make it so that remoteOutput applies only when numWorkers == 1 and ignored if numWorkers > 1?

arhimondr:
@mbasmanova In most of the test suites we check both, single node and distributed plans. Here is an example: https://fburl.com/code/4jp4hk3q

mbasmanova (approved):

PR #1054, opened 2026-03-13
fix(optimizer): Fix DT columns when finalizing window operation in filter (#1054)

kagamiori:
Summary:

Previously in ToGraph::makeFilterQueryGraph, when the current DT contains a
window plan, it finalize the DT with the filter node's used channels and then
adding the filter to the outer DT. However, the filter node's used channels may
not include all input columns of the filter. This would cause 
Optimization::placeConjuncts later failed to place the filter conjunct.

To fix this problem, this PR makes ToGraph::makeFilterQueryGraph to finalize
the DT with the filter's input node whose used channels are guaranteed to 
include all input columns of the filter.

Reviewed By: peterenescu

Differential Revision: D96413832

### DIFF
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -3440,7 +3440,7 @@ void ToGraph::makeFilterQueryGraph(
   // Window functions compute over the full input; pushing any filter below
   // them changes their semantics.
   if (currentDt_->hasWindow()) {
-    finalizeDt(filter);
+    finalizeDt(input);
   }
 
   addFilter(input, filter.predicate());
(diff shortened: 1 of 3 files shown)


axiom/optimizer/ToGraph.cpp
```diff
@@ -3432,7 +3432,7 @@ void ToGraph::makeFilterQueryGraph(
   // Window functions compute over the full input; pushing any filter below
   // them changes their semantics.
   if (currentDt_->hasWindow()) {
-    finalizeDt(filter);
+    finalizeDt(input);
```
mbasmanova: @kagamiori Wei, would you explain how this change works? The output type of filter and input should be the same, no?

kagamiori: Hi @mbasmanova, no, the output type of a node only contains columns that the downstream of this node needs. For the added unit test in this PR as an example, the output columns of the `filter` node contains only n_name, while the output columns of `input` (i.e., `filter`'s input node) contains both n_name and rn.

If we use finalizeDt(filter) here, the placeConjuncts will not be able to place the filter conjunct of this DT later (because not all input columns of the conjunct present), which will trigger the 'Failed to place a conjunct' failure.

PR #1057, opened 2026-03-14
feat(cli): Add query lifecycle callbacks to Console (#1057)

amitkdutta:
Summary:

Add optional start and completion callbacks to Console so callers can
hook into the query lifecycle for logging, metrics, or other processing.

- startCallback: invoked before parse for every query.
- completionCallback: invoked after execution with query timing, row
  counts, and success/failure status.

Reviewed By: hdikeman, mbasmanova

Differential Revision: D96492425

### DIFF
--- a/axiom/cli/Console.h
+++ b/axiom/cli/Console.h
@@ -15,9 +15,10 @@
  */
 #pragma once
 
+#include <chrono>
 #include <functional>
 #include <memory>
-#include <unordered_map>
+#include <optional>
 #include <utility>
 #include "axiom/cli/SqlQueryRunner.h"
 #include "axiom/sql/presto/SqlStatement.h"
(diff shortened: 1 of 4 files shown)


axiom/cli/Console.h
```diff
@@ -43,16 +43,47 @@ using PermissionCheck =
...
+
+/// Holds query metadata at completion time.
+struct QueryCompletionInfo {
+  std::string queryId;
+  std::string_view query;
+  bool succeeded{true};
+  std::string errorMessage;
+  /// Serialized query plan for downstream logging.
+  std::string planString;
+  uint64_t parseMicros{0};
```
mbasmanova: Would you document all the fields?

I think we need timing for parsing, optimizing and execution. optimizeMicros is missing.

PR #1088, opened 2026-03-18
feat(parser): Add logical_plan::Expr::operator== (#1088)

kagamiori:
Summary:

Add custom equality comparison for all Expr types.

- Add operator== to Expr base class with virtual equalTo() dispatch.
- Implement equalTo() for all derived classes except SubqueryExpr: 
InputReferenceExpr, ConstantExpr, CallExpr, SpecialFormExpr, AggregateExpr, 
WindowExpr, LambdaExpr.
- SubqueryExpr::equalTo() is not implemented and throws in this PR because 
there is no operator== of LocalPlanNode for SubqueryExpr::subquery_. This 
support will be added in https://github.com/facebookincubator/axiom/pull/1092.
- All shared_ptr members are dereferenced for value comparison, not pointer comparison.

Differential Revision: D97048671

### DIFF
--- a/axiom/logical_plan/Expr.h
+++ b/axiom/logical_plan/Expr.h
@@ -144,13 +144,31 @@ class Expr : public velox::ISerializable {
   virtual void accept(const ExprVisitor& visitor, ExprVisitorContext& context)
       const = 0;
 
+  /// Returns true if two expressions are structurally equal.
+  bool operator==(const Expr& other) const;
+
   std::string toString() const;
 
   /// Registers deserializers for all Expr subclasses.
   static void registerSerDe();
 
  protected:
-  /// Serializes common base fields (name, type, inputs).
+  // Returns true if derived-class-specific fields are equal. Called only when
+  // kind, type, and inputs already match.
+  virtual bool equalTo(const Expr& other) const = 0;
+
+  // Compares two ExprPtr values for structural equality. Either or both
+  // pointers may be null. Two nulls are considered equal; a null and a
+  // non-null are not.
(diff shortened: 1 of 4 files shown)


axiom/logical_plan/Expr.h
```diff
@@ -749,9 +776,19 @@ class SubqueryExpr : public Expr {
...
 
  private:
+  bool equalTo(const Expr& other) const override;
+
   const LogicalPlanNodePtr subquery_;
 };
 
 using SubqueryExprPtr = std::shared_ptr<const SubqueryExpr>;
 
+// Compares two nullable ExprPtr values for structural equality.
```
mbasmanova: use /// for public APIs

Why do we need equalNullableExprs? Does equalExprVectors also has "nullable" semantic? Would you expand the docs to clarify.

axiom/logical_plan/tests/ExprTest.cpp
```diff
@@ -62,5 +68,369 @@ TEST_F(ExprTest, looksConstant) {
...
+      WindowExpr::BoundType::kPreceding,
+      literal(10),
+      WindowExpr::BoundType::kFollowing,
+      literal(3)};
+  EXPECT_EQ(frame, frameAnother);
+  EXPECT_NE(frame, frameDifferent);
+}
+
+TEST_F(ExprTest, lambdaExprEquality) {
+  auto signature = ROW({"x"}, {BIGINT()});
```
mbasmanova: simplify: ROW("x", BIGINT())

mbasmanova (commented):
@kagamiori Thank you for iterating. PR description needs updating.

PR #1152, opened 2026-03-25
feat: Convert SplitManager/SplitSource APIs to coroutines (#1152)

hdikeman:
Summary:

Making split-related calls non-blocking using coroutines, since some connector implementations may do significant I/O work to construct splits or enumerate partitions and during these long calls may want to release the thread to do other work

Reviewed By: mbasmanova

Differential Revision: D98218415


mbasmanova (approved):
Thanks.

PR #1184, opened 2026-04-01
docs: Add design doc for bounded character types (CHAR(n), VARCHAR(n))

mbasmanova:
Differential Revision: D99098191

### DIFF
--- a/axiom/sql/presto/docs/BoundedCharacterTypes.md
+++ b/axiom/sql/presto/docs/BoundedCharacterTypes.md
@@ -1,0 +1,408 @@
+# Bounded Character Types — CHAR(n) and VARCHAR(n): Design Proposal
+
+*April 1, 2026*
+
+## Three distinct string types
+
+PrestoSQL has three string types. They look similar but have different
+semantics:
+
+| Type | Length | On CAST | Comparison | typeof() |
+|------|--------|---------|------------|----------|
+| `VARCHAR` | Unbounded | No truncation, no padding | Exact (spaces significant) | `varchar` |
+| `VARCHAR(n)` | Variable, ≤ n | Truncates to n | Exact (spaces significant) | `varchar(n)` |
+| `CHAR(n)` | Fixed, exactly n | Truncates to n, pads with spaces | Pad-space (trailing spaces ignored) | `char(n)` |
+
+These are three different types, not variations of the same type:
+
+- `VARCHAR` and `VARCHAR(n)` differ in cast behavior: `CAST('abcdef' AS
+  VARCHAR(3))` returns `'abc'`, but `CAST('abcdef' AS VARCHAR)` returns
+  `'abcdef'`.
+- `VARCHAR(n)` and `CHAR(n)` differ in padding and comparison:
+  `CAST('abc' AS VARCHAR(5))` returns `'abc'` (3 chars), but
+  `CAST('abc' AS CHAR(5))` returns `'abc  '` (5 chars). `CHAR` ignores
+  trailing spaces in comparisons; `VARCHAR` does not.
+
+String literals have type `VARCHAR(n)` where n is the literal's length:
+`'abc'` is `VARCHAR(3)`, `''` is `VARCHAR(0)`. Unbounded `VARCHAR` (no bound)
+is a separate type — only produced by an explicit `CAST(x AS VARCHAR)`.
+
+In the Presto Java implementation, unbounded `VARCHAR` has length
+`Integer.MAX_VALUE`. `CHAR` has a max length of `65536`.
+
+All semantics verified on production Presto (see Appendix A).
+
+Note: Hive connector do not support bounded character types. These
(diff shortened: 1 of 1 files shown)


axiom/sql/presto/docs/BoundedCharacterTypes.md
```diff
@@ -0,0 +1,408 @@
...
+String literals have type `VARCHAR(n)` where n is the literal's length:
+`'abc'` is `VARCHAR(3)`, `''` is `VARCHAR(0)`. Unbounded `VARCHAR` (no bound)
+is a separate type — only produced by an explicit `CAST(x AS VARCHAR)`.
+
+In the Presto Java implementation, unbounded `VARCHAR` has length
+`Integer.MAX_VALUE`. `CHAR` has a max length of `65536`.
+
+All semantics verified on production Presto (see Appendix A).
+
+Note: Hive connector do not support bounded character types. These
```
KcGo007, line 35: One question : For values read from connectors/files, where do you think `CHAR(n)` padding should happen?

For example, if the logical type is `CHAR(5)` and the physical reader returns `A` without padding, would you prefer:

* padding inside each concrete reader (Parquet / ORC / DWRF),
* padding at a shared scan boundary such as `FileDataSource`,
* or some other layer?

My preference is the shared scan boundary because it has both the logical `CHAR(n)` type and the materialized vector, and avoids duplicating the same CHAR logic across file-format readers.

axiom/sql/presto/docs/BoundedCharacterTypes.md
```diff
@@ -0,0 +1,408 @@
...
+  before the comparison reaches the execution engine.
+
+The common supertype rules for comparisons (see Appendix B for production
+Presto plans):
+
+| Comparison | Common supertype | Rule |
+|---|---|---|
+| `CHAR(5)` = `CHAR(10)` | `CHAR(10)` | CHAR wins; `max(n, m)` |
+| `CHAR(5)` = `VARCHAR(10)` | `CHAR(10)` | CHAR wins; `max(n, m)` |
+| `CHAR(5)` = `VARCHAR` | `CHAR(65536)` | CHAR wins; 65536 = max CHAR length (*) |
```
KcGo007, line 60: The common-supertype rule says:

CHAR(n) vs unbounded VARCHAR → CHAR(65536).

With the padded CHAR representation described in the proposal, a short value coerced to CHAR(65536) could require roughly 64 KB of physical string data per value.

Is physically materializing that padding the intended implementation, or is there an expected optimization for this case?

Thanks

mbasmanova:
@KcGo007 Thanks for the careful read.

Before you spend more time on it: I wrote this in April and did not pursue it.
Bounded character types only turn up on legacy MySQL tables, and I concluded the
functionality wasn't worth the implementation cost — none of this is being
built.

Answering your two questions properly means reloading a fair amount of context,
so before I do that: what's your interest here? If you're running into
`CHAR(n)`/`VARCHAR(n)` in practice, or you're thinking about picking the work
up, say so and I'll dig the details back out.

KcGo007:
> @KcGo007 Thanks for the careful read.
> 
> Before you spend more time on it: I wrote this in April and did not pursue it. Bounded character types only turn up on legacy MySQL tables, and I concluded the functionality wasn't worth the implementation cost — none of this is being built.
> 
> Answering your two questions properly means reloading a fair amount of context, so before I do that: what's your interest here? If you're running into `CHAR(n)`/`VARCHAR(n)` in practice, or you're thinking about picking the work up, say so and I'll dig the details back out.

@mbasmanova Thanks for the context. Yes — I’m actively picking up CHAR(n) support for Prestissimo/Velox because we’re running into it in practice, and I found your design while working through the implementation.

I’m trying to follow the architecture you proposed rather than inventing a separate design, so your input would be very helpful.

KcGo007:
For now, I'm following Masa's design doc for most of the CHAR type implementation: https://github.com/facebookincubator/axiom/pull/1184

axiom/sql/presto/docs/BoundedCharacterTypes.md
mbasmanova: @KcGo007 The Hive connector doesn't support bounded types — only MySQL and
similar do, which is why the doc scopes them that way. So the Parquet/ORC/DWRF
layering question doesn't arise.

Where it does arise, it's the connector's job. A connector that declares a
column `CHAR(n)` returns values of length n, the same way one declaring
`BIGINT` returns 8-byte integers. Nothing above the connector should be fixing
up a value to match the type it was handed.

axiom/sql/presto/docs/BoundedCharacterTypes.md
mbasmanova: @KcGo007 This looks like a real gap. As written the proposal does materialize
the padding, so your ~64KB per value is real — there is no optimization behind
it.

This needs more thought.

PR #1221, opened 2026-04-08
feat(cli): Add .run command to execute SQL from a file (#1226)

peterenescu:
Summary:

Add .run <file> dot-command to the interactive CLI. Reads SQL from the
specified file and executes all statements, similar to --init but
available during an interactive session.

Also migrated all interactive commands to dot-command syntax (.help,
.set, .clear, .flags, .exit/.quit). Dot-commands do not require a
semicolon terminator. Unrecognized dot-commands show an error instead
of being sent to the SQL parser.

Reviewed By: amitkdutta

Differential Revision: D100202176


peterenescu:
https://github.com/facebookincubator/axiom/pull/1227

PR #1224, opened 2026-04-09
fix: Failure message templates for programmatic error categorization (#1224)

HeidiHan0000:
Summary:

Three fixes in axel (OSS) and axiom to ensure failure message templates are correctly captured for programmatic error categorization:

1. **QueryError template-aware overload**: Add `QueryError::create(errorCode, messageTemplate, args...)` that stores the static format string as the failure template while formatting the full message for display. Add `createWithTemplate` private helper.

2. **VeloxException template fallback**: When VeloxException has an empty messageTemplate (from messageless VELOX_CHECK/VELOX_FAIL), synthesize a groupable template from `failingExpression()` instead of falling back to `ex.what()` (which includes the full stack trace and is ungroupable).

3. **std::runtime_error → VELOX_FAIL/VELOX_USER_FAIL**: Convert `throw std::runtime_error` to `VELOX_FAIL`/`VELOX_USER_FAIL` with format args in axel client, axiom SQL parser, axiom hive connector, and axiom AST printer so that VeloxException captures the template string.

Reviewed By: kagamiori

Differential Revision: D99975400

### DIFF
--- a/axiom/connectors/hive/LocalHiveConnectorMetadata.cpp
+++ b/axiom/connectors/hive/LocalHiveConnectorMetadata.cpp
@@ -24,6 +24,7 @@
 #include "axiom/connectors/hive/HiveMetadataConfig.h"
 #include "axiom/connectors/hive/LocalTableMetadata.h"
 #include "axiom/optimizer/JsonUtil.h"
+#include "velox/common/base/Exceptions.h"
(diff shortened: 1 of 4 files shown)


kagamiori (approved):
LGTM. Thanks!

PR #1226, opened 2026-04-09
feat(cli): Add .run command to execute SQL from a file

mbasmanova:
Summary:
Add .run <file> dot-command to the interactive CLI. Reads SQL from the
specified file and executes all statements, similar to --init but
available during an interactive session.

Also migrated all interactive commands to dot-command syntax (.help,
.set, .clear, .flags, .exit/.quit). Dot-commands do not require a
semicolon terminator. Unrecognized dot-commands show an error instead
of being sent to the SQL parser.

Differential Revision: D100202176

### DIFF
--- a/axiom/cli/Console.cpp
+++ b/axiom/cli/Console.cpp
@@ -17,6 +17,7 @@
 #include "axiom/cli/Console.h"
 #include <folly/FileUtil.h>
 #include <unistd.h>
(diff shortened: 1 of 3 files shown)


amitkdutta (approved):
Looks good. Thanks @mbasmanova

amitkdutta (approved):

PR #1233, opened 2026-04-10
feat: Add SerDe support to SystemConnector (#1233)

shanhao-203:
Summary:

CONTEXT: The system connector needs serialization/deserialization support so that table handles, column handles, and splits can be transferred across the distributed query execution boundary.

WHAT:
- Add serialize()/create()/registerSerDe() methods to SystemColumnHandle, SystemTableHandle, and SystemSplit.
- Change SystemTableHandle::layout_ from reference to pointer to support a deserialization constructor without a layout reference.
- Add SerDe round-trip unit tests for all three types.

Reviewed By: arhimondr

Differential Revision: D100267751

### DIFF
--- a/axiom/connectors/system/tests/CMakeLists.txt
+++ b/axiom/connectors/system/tests/CMakeLists.txt
@@ -12,12 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-add_executable(axiom_system_connector_metadata_test SystemConnectorMetadataTest.cpp)
+add_executable(
+  axiom_system_connector_test
+  SystemConnectorMetadataTest.cpp
+  SystemConnectorSerDeTest.cpp
+)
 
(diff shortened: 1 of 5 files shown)


axiom/connectors/system/tests/CMakeLists.txt
```diff
@@ -26,3 +26,14 @@ target_link_libraries(
   GTest::gtest
   GTest::gtest_main
 )
+
+add_executable(axiom_system_connector_serde_test SystemConnectorSerDeTest.cpp)
```
mbasmanova: Add test to an existing target. Avoid making lots of small targets in cmake. See https://github.com/facebookincubator/axiom/blob/main/CONTRIBUTING.md#cmake-guidelines

arhimondr (approved):

PR #1253, opened 2026-04-15
feat: Add system.metadata.session_properties table (#1253)

### LINKED ISSUE #1238: Composable session configuration for SET/SHOW/RESET SESSION
## Problem

Axiom is a composable system built from independent components: parser, optimizer, execution engine, and Velox runtime. Each component has its own configuration, but there is no unified way to manage these configs:

- `SET SESSION` only reaches Velox's `QueryConfig` (a flat string map). Optimizer and execution options are not accessible.
- `SHOW SESSION` only shows Velox config entries — no optimizer or execution settings.
- Optimizer options (`OptimizerOptions`) are a typed C++ struct, only controllable via gflags (`.set` commands in the CLI).
- Execution options (`RunOptions`) are also only controllable via gflags.
- Applications using Axiom as a library have no standard way to discover or modify component configs at runtime.

## Overview

Each component (optimizer, Velox runtime, connectors) declares its own
configuration properties — name, type, default, description — without
knowing how they are exposed to the user.

The application that assembles these components registers each one under
a namespace prefix (e.g., `"optimizer"`, `"runtime"`, `"hive"`). Users
refer to properties by their prefix-qualified name:

```sql
SET SESSION optimizer.sample_joins = false;
SET SESSION hive.orc_compression = 'ZSTD';
SHOW SESSION LIKE 'optimizer%';
RESET SESSION optimizer.sample_joins;
```

Property declarations and validation are shared across sessions.
Per-session state is a thin overlay of overrides on top of defaults.
Before each query, the runner reads the effective values and passes
them to components through their existing config mechanisms
(`OptimizerOptions`, `QueryConfig`, `ConnectorSession`).

### Architecture

```
       Shared (built once)                  Per-session
┌──────────────────────────────┐   ┌──────────────────────┐
│       ConfigRegistry         │   │    SessionConfig     │
│                              │   │                      │
│  "optimizer" ──► Provider    │   │  overrides_:         │
│  "runtime"   ──► Provider    │◄──│    {name: value}     │
│  "hive"      ──► Provider    │   │                      │
└──────────────────────────────┘   └──────────────────────┘
                                             │
                ┌────────────────────────────┤
                │               │            │
                ▼               ▼            ▼
    getAll("optimizer") getAll("runtime") getAll("hive")
                │               │            │
                ▼               ▼            ▼
 OptimizerOptions::from() QueryConfig() ConnectorSession()
```

| Class | Lifetime | Mutable | Responsibility |
|---|---|---|---|
| `ConfigProperty` | — | — | Property descriptor: name, type, default, description |
| `ConfigProvider` | Shared | No | Declares properties, validates values |
| `ConfigRegistry` | Shared | No (after assembly) | Maps prefixes to providers, resolves qualified names |
| `SessionConfig` | Per-session | Yes | Stores overrides, merges with defaults |

---

### ConfigProperty

Each configuration property is described by a lightweight descriptor:

```cpp
enum class ConfigPropertyType {
  kBoolean,
  kInteger,
  kDouble,
  kString,
};

struct ConfigProperty {
  std::string name;                          // e.g., "pushdown_subfields"
  ConfigPropertyType type;
  std::optional<std::string> defaultValue;   // e.g., "false"; nullopt if no default.
  std::string description;                   // Human-readable description.
};
```

### ConfigProvider

A component that has configurable properties implements `ConfigProvider`
to declare them and validate values. A `ConfigProvider` is shared and
stateless — it does not hold per-session values. It knows only its own
property names (e.g., `"pushdown_subfields"`), not how it is exposed to
the user.

```cpp
class ConfigProvider {
 public:
  virtual ~ConfigProvider() = default;

  /// Returns the list of supported configuration properties. Property
  /// names must be unique within a provider.
  virtual std::vector<ConfigProperty> properties() const = 0;

  /// Validates a value for the named property. Called after type
  /// validation has already passed. The provider may apply additional
  /// domain-specific validation (e.g., value ranges, allowed enum
  /// values). Throws if the property name is unknown or the value is
  /// invalid.
  virtual void validate(std::string_view name, std::string_view value) const = 0;
};
```

Connectors can optionally expose session properties by implementing this
interface:

```cpp
/// Returns a ConfigProvider for this connector's session properties,
/// or nullptr if the connector has no configurable properties.
virtual std::shared_ptr<ConfigProvider> configProvider() { return nullptr; }
```

### ConfigRegistry

The system that composes components into an application collects their
`ConfigProvider`s into a `ConfigRegistry`. Each provider is registered
under a namespace prefix chosen by the system (e.g., `"optimizer"`,
`"runtime"`, `"hive"`). This prefix is not known to the component — it
is a deployment decision.

The registry is shared and immutable after assembly. It maps prefixes to
providers and knows the full set of available properties.

```cpp
class ConfigRegistry {
 public:
  /// Registers a provider under a namespace prefix. The prefix is
  /// chosen by the system, not the component. Throws if the prefix
  /// is already registered.
  void add(std::string_view prefix, std::shared_ptr<ConfigProvider> provider);

  /// Looks up the provider and property for a prefix-qualified name.
  /// Splits on the first '.', finds the provider by prefix, and finds
  /// the property by name. Throws if the prefix or property name is
  /// unknown.
  struct ResolvedProperty {
    std::string_view name;
    const ConfigProvider* provider;
    const ConfigProperty* property;
  };
  ResolvedProperty resolve(std::string_view qualifiedName) const;

  /// Returns all properties across all providers with their prefix.
  struct Entry {
    std::string prefix;
    ConfigProperty property;
  };
  std::vector<Entry> all() const;
};
```

### SessionConfig

Per-session mutable state. Holds a reference to the shared
`ConfigRegistry` and a map of property overrides. Reads check the
override map first, then fall back to the provider's default.

```cpp
class SessionConfig {
 public:
  explicit SessionConfig(std::shared_ptr<const ConfigRegistry> registry);

  /// Sets a property by prefix-qualified name (e.g.,
  /// "optimizer.pushdown_subfields"). Performs two-level validation:
  /// 1. Type validation — checks the value against the property's
  ///    declared ConfigPropertyType (e.g., rejects "banana" for a
  ///    boolean property).
  /// 2. Domain validation — delegates to the provider's validate(),
  ///    which may apply additional constraints (e.g., value ranges,
  ///    allowed enum values).
  /// Returns true if the value changed, false if the new value is
  /// the same as the current value. Throws if the prefix is unknown,
  /// the property name is unknown, or either validation fails.
  bool set(std::string_view qualifiedName, std::string_view value);

  /// Resets a property to its default value. Returns true if the value
  /// changed. Throws if the property name is unknown.
  bool reset(std::string_view qualifiedName);

  /// Gets a property value by prefix-qualified name. Returns the
  /// override if set, otherwise the default. Returns nullopt if the
  /// property has no default and has not been set.
  std::optional<std::string> get(std::string_view qualifiedName) const;

  /// Returns all properties with their current values (override or
  /// default).
  struct Entry {
    std::string prefix;
    ConfigProperty property;
    std::optional<std::string> currentValue;
    bool isOverridden;
  };
  std::vector<Entry> all() const;

 private:
  std::shared_ptr<const ConfigRegistry> registry_;
  std::unordered_map<std::string, std::string> overrides_;
};
```

**`set()` flow:**

1. Call `registry_->resolve(qualifiedName)` to split the prefix, find the
   provider, and look up the `ConfigProperty`.
2. Validate the value against `ConfigProperty::type` (e.g., reject
   `"banana"` for `kBoolean`).
3. Call `provider->validate(name, value)` for domain-specific checks.
4. Check `overrides_` — if the value is the same as the current effective
   value, return false.
5. Store in `overrides_` and return true.

**Applying configs to components:**

`SessionConfig` stores overrides as strings. Components need their own
typed config objects. Each component's config struct provides a static
factory that translates from name-value pairs:

```cpp
struct OptimizerOptions {
  // ... fields ...

  /// Constructs options from session property name-value pairs.
  static OptimizerOptions from(
      const std::unordered_map<std::string, std::string>& properties);
};
```

The runner materializes configs before each query:

```cpp
auto optimizerProps = sessionConfig->getAll("optimizer");
auto optimizerOptions = OptimizerOptions::from(optimizerProps);
```

For Velox, `QueryConfig` already constructs from a string map, so no
new factory is needed:

```cpp
auto runtimeProps = sessionConfig->getAll("runtime");
auto queryConfig = std::make_shared<core::QueryConfig>(runtimeProps);
```

**Note on per-query values vs session properties.** Today, `config_` in
`SqlQueryRunner` is a flat map that feeds directly into `QueryConfig`.
It mixes two kinds of values:

- **Session properties** — user-facing, persist across queries. Examples:
  `kSessionTimezone` (defaults to local timezone), `kAdjustTimestampToTimezone`
  (defaults to `"true"`). These map naturally to `runtime.*` session properties.
- **Per-query values** — set programmatically by the runner for each query.
  Currently only `kSessionStartTime` (set to `now()`). Not user-facing, not
  overridable via `SET SESSION`. This is query-scoped runtime state (used
  by `current_timestamp()` to return a consistent value within a query),
  not configuration. In Presto, it lives on the `Session` object, not in
  session properties. Velox stores it in `QueryConfig` as a convenience.

With this design, `SessionConfig` handles session properties only.
Per-query values are merged separately when constructing `QueryConfig`:

```cpp
auto runtimeProps = sessionConfig->getAll("runtime");
// Inject per-query values (not session properties).
runtimeProps["start_time"] = currentTimestamp();
auto queryConfig = std::make_shared<core::QueryConfig>(runtimeProps);
```

For connectors, properties are passed via `ConnectorSession`, which
is extended to carry a property map:

```cpp
class ConnectorSession final {
 public:
  ConnectorSession(
      std::string queryId,
      std::unordered_map<std::string, std::string> properties);

  const std::string& queryId() const;
  const std::unordered_map<std::string, std::string>& properties() const;
};
```

Connector methods (`findTable`, `listPartitions`, etc.) already receive
a `ConnectorSession`. Today it carries only `queryId` and is created via
`Session::toConnectorSession(connectorId)`, where `connectorId` is
already passed but currently unused.

With this design, `Session` holds a reference to `SessionConfig`.
`toConnectorSession()` uses the `connectorId` to fetch
connector-scoped properties and pass them to `ConnectorSession`:

```cpp
connector::ConnectorSessionPtr Session::toConnectorSession(
    std::string_view connectorId) const {
  auto props = sessionConfig_->getAll(std::string(connectorId));
  return std::make_shared<connector::ConnectorSession>(queryId_, props);
}
```

A single query may access multiple catalogs. Each call to
`toConnectorSession()` with a different `connectorId` produces a
`ConnectorSession` with that catalog's properties. Connectors read
their properties from the `ConnectorSession` they already receive.

`SessionConfig::getAll()` returns all properties for a prefix with
overrides merged on top of defaults:

```cpp
/// Returns all property values for a given prefix, with overrides
/// merged on top of defaults.
std::unordered_map<std::string, std::string>
    getAll(std::string_view prefix) const;
```

**`get()` flow:**

1. Call `registry_->resolve(qualifiedName)` to find the property.
2. If `overrides_` contains the qualified name, return the override.
3. Otherwise, return `ConfigProperty::defaultValue` (which may be
   nullopt).

**`reset()` flow:**

1. Call `registry_->resolve(qualifiedName)` to verify the property exists.
2. If `overrides_` contains the qualified name, erase it and return true.
3. Otherwise, return false (nothing to reset).

### Assembly

The CLI or application assembles the registry at startup. Core components
are registered explicitly. Connectors are discovered automatically from
the existing connector registry — any connector that provides a
`ConfigProvider` is registered under its catalog name. A `SessionConfig`
is then created per session referencing the shared registry.

```cpp
// Shared, built once at startup.
auto registry = std::make_shared<ConfigRegistry>();

// Core components.
registry->add("optimizer", std::make_shared<OptimizerConfigProvider>());
registry->add("runtime", std::make_shared<VeloxConfigProvider>());

// Connectors — discovered from the existing connector registry.
for (const auto& [catalogName, connector] : connectorRegistry) {
  if (auto provider = connector->configProvider()) {
    registry->add(catalogName, provider);
  }
}

// Per session.
auto sessionConfig = std::make_shared<SessionConfig>(registry);
```

Adding a new connector with session properties requires no changes to
the assembly code — the connector just implements `configProvider()` and
it is automatically picked up.

### SQL Integration

**SET SESSION:**
```sql
SET SESSION optimizer.pushdown_subfields = true;
SET SESSION runtime.session_timezone = 'UTC';
SET SESSION hive.split_target_bytes = 33554432;
SET SESSION hive.orc_compression = 'ZSTD';
```

The runner splits the qualified name on the first `.`, looks up the
prefix in the registry, validates, and stores the override in the
session's `SessionConfig`.

**SHOW SESSION:**
```
Name                               | Value    | Default  | Type    | Description
optimizer.pushdown_subfields       | false    | false    | boolean | Extract accessed subfields of complex columns in table scan
optimizer.sample_joins             | true     | true     | boolean | Sample joins to determine optimal join order
optimizer.syntactic_join_order     | false    | false    | boolean | Disable cost-based join ordering; use query order
runtime.session_timezone           | UTC      |          | string  | Session timezone
hive.split_target_bytes            | 16777216 | 16777216 | integer | Target split size in bytes
hive.orc_compression               | SNAPPY   | SNAPPY   | string  | ORC compression codec
...
```

**SHOW SESSION LIKE 'optimizer%':**
Filters to only optimizer properties.

**SHOW SESSION LIKE 'hive%':**
Filters to only Hive connector properties.

**RESET SESSION:**
```sql
RESET SESSION optimizer.pushdown_subfields;
```

Removes the override, restoring the property to its default value.

### system.metadata.session_properties

A queryable table exposing all registered properties:

| Column | Type | Description |
|---|---|---|
| `component` | VARCHAR | Namespace prefix (e.g., `optimizer`) |
| `name` | VARCHAR | Property name (e.g., `pushdown_subfields`) |
| `type` | VARCHAR | Property type |
| `default_value` | VARCHAR | Default value |
| `current_value` | VARCHAR | Current session value |
| `description` | VARCHAR | Human-readable description |

```sql
SELECT name, current_value, default_value
FROM system.metadata.session_properties
WHERE component = 'optimizer';
```

mbasmanova:
Summary:

Add a queryable system table that exposes all registered session
properties, making them discoverable via standard SQL:

  SELECT * FROM system.metadata.session_properties
  WHERE component = 'optimizer';

Columns: component, name, type, default_value, current_value, description.

The table reads from SessionConfig at query time, so current_value
reflects any SET SESSION overrides in the active session.

Fixes https://github.com/facebookincubator/axiom/issues/1238

Reviewed By: jagill

Differential Revision: D101006261

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,7 @@ with Velox. These components are:
   - top-level “connectors” directory
   - [Hive](axiom/connectors/hive/README.md) — Local filesystem connector for Parquet, DWRF, and TEXT (including CSV) files.
   - [TPC-H](axiom/connectors/tpch/README.md) — Read-only, in-memory TPC-H benchmark data.
+  - [System](axiom/connectors/system/README.md) — Read-only metadata tables for session properties and active queries.
   - [Test](axiom/connectors/tests/README.md) — In-memory read-write connector for unit testing.
 - [CLI](axiom/cli/README.md) - Interactive SQL command line for executing
   queries against in-memory TPC-H dataset and local Hive data.
@@ -316,7 +317,7 @@ Axiom integrates Velox as a Git submodule, referencing a specific commit of the
 Velox repository. The Velox badge at the top of this README shows the current
 commit and how far behind it is from Velox main.
 
-[See what changed since the current Velox commit.](https://github.com/facebookincubator/velox/compare/24e6ab97ba8015c3b6fae82e8184047cadf521df...main)
+[See what changed since the current Velox commit.](https://github.com/facebookincubator/velox/compare/09c4fb905f0e02b0f334b0e99ff9e8b0c8a5165b...main)
 <!-- pre-commit check-velox-readme validates the SHA above matches the submodule -->
 
 Advance Velox when your changes depend on code in Velox that
--- a/axiom/cli/AxiomSql.cpp
+++ b/axiom/cli/AxiomSql.cpp
@@ -39,7 +39,6 @@ int main(int argc, char** argv) {
     auto defaultSchema = "tiny";
 
     connectors.registerTestConnector();
-    connectors.registerSystemConnector();
 
     if (!FLAGS_data_path.empty()) {
       defaultConnector = connectors.registerLocalHiveConnector(
@@ -60,6 +59,9 @@ int main(int argc, char** argv) {
 
     return std::make_pair(connectorId, schema);
   });
+
+  // Register after initialize() so sessionConfig() is available.
+  connectors.registerSystemConnector(runner.sessionConfig());
 
   axiom::sql::Console console{runner};
   console.initialize();
--- a/axiom/cli/CMakeLists.txt
+++ b/axiom/cli/CMakeLists.txt
@@ -26,6 +26,7 @@ add_library(
 
 target_link_libraries(
   axiom_cli
+  velox_connector_registry
   velox_query_config_provider
   axiom_runner_local_runner
   axiom_optimizer_multifragment_plan
--- a/axiom/cli/Connectors.cpp
+++ b/axiom/cli/Connectors.cpp
@@ -17,6 +17,7 @@
 #include "axiom/cli/Connectors.h"
 
 #include <folly/system/HardwareConcurrency.h>
+#include "axiom/common/SessionConfig.h"
 #include "axiom/connectors/ConnectorMetadata.h"
 #include "axiom/connectors/hive/HiveMetadataConfig.h"
 #include "axiom/connectors/hive/LocalHiveConnectorMetadata.h"
@@ -25,6 +26,7 @@
 #include "axiom/connectors/tests/TestConnector.h"
 #include "axiom/connectors/tpch/TpchConnectorMetadata.h"
 #include "velox/connectors/Connector.h"
+#include "velox/connectors/ConnectorRegistry.h"
 #include "velox/connectors/hive/HiveConnector.h"
 #include "velox/dwio/common/FileSink.h"
 #include "velox/dwio/dwrf/RegisterDwrfReader.h"
@@ -48,6 +50,38 @@ void initializeFileFormats() {
   velox::text::registerTextWriterFactory();
 }
 
+// Adapts SessionConfig to the SessionPropertiesProvider interface for the
+// system connector's metadata.session_properties table.
+class SessionConfigPropertiesProvider
+    : public connector::system::SessionPropertiesProvider {
+ public:
+  explicit SessionConfigPropertiesProvider(const SessionConfig& sessionConfig)
+      : sessionConfig_(sessionConfig) {}
+
+  std::vector<connector::system::SessionPropertyInfo> getSessionProperties()
+      const override {
+    using velox::config::ConfigPropertyTypeName;
+
+    auto entries = sessionConfig_.all();
+    std::vector<connector::system::SessionPropertyInfo> result;
+    result.reserve(entries.size());
+    for (const auto& entry : entries) {
+      result.push_back({
+          entry.prefix,
+          entry.property.name,
+          std::string(ConfigPropertyTypeName::toName(entry.property.type)),
+          entry.property.defaultValue.value_or(""),
+          entry.currentValue.value_or(""),
+          entry.property.description,
+      });
+    }
+    return result;
+  }
+
+ private:
+  const SessionConfig& sessionConfig_;
+};
+
 } // namespace
 
 Connectors::Connectors() {
@@ -58,7 +92,7 @@ Connectors::~Connectors() {
   for (const auto& connectorId : connectorIds_) {
     // Unregister metadata first since it may reference the connector.
     connector::ConnectorMetadata::unregisterMetadata(connectorId);
-    velox::connector::unregisterConnector(connectorId);
+    velox::connector::ConnectorRegistry::global().erase(connectorId);
   }
 }
 
@@ -77,6 +111,13 @@ void Connectors::initialize() {
   ioExecutor_ = getSharedIoExecutor();
 }
 
+void Connectors::registerConnector(
+    const std::shared_ptr<velox::connector::Connector>& connector) {
+  connectorIds_.push_back(connector->connectorId());
+  velox::connector::ConnectorRegistry::global().insert(
+      connector->connectorId(), connector);
+}
+
 std::shared_ptr<velox::connector::Connector> Connectors::registerTpchConnector(
     const std::string& connectorId) {
   auto emptyConfig = std::make_shared<velox::config::ConfigBase>(
@@ -84,8 +125,7 @@ std::shared_ptr<velox::connector::Connector> Connectors::registerTpchConnector(
 
   velox::connector::tpch::TpchConnectorFactory factory;
   auto connector = factory.newConnector(connectorId, emptyConfig);
-  connectorIds_.push_back(connector->connectorId());
-  velox::connector::registerConnector(connector);
+  registerConnector(connector);
 
   auto tpchConnector =
       dynamic_cast<velox::connector::tpch::TpchConnector*>(connector.get());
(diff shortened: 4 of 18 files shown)


jagill (approved):
Thank you!

PR #1258, opened 2026-04-16
fix(tests): Increase waitForCompletion timeout and check return value (#1258)

### LINKED ISSUE #1257: Flaky axiom_optimizer_tests: "Unexpected alive participants on destruction"
`axiom_optimizer_tests` fails intermittently on macOS CI with:

```
SharedArbitrator.cpp:358, Function:shutdown
Unexpected alive participants on destruction: query.q128.136
```

The failure occurs during test teardown when `SharedArbitrator::shutdown` finds a query participant still registered.

### Root cause

`QueryTestBase::waitForCompletion` uses a 50ms timeout (50'000 microseconds) and silently swallows timeout exceptions. On loaded CI runners, 50ms can be insufficient for async task cleanup. When the timeout is silently ignored, the query's memory pool participant remains registered, and `OperatorTestBase::TearDown` → `resetMemory` destroys the arbitrator, which fails on the alive participant.

For comparison, `LocalRunnerTest` uses a 500ms timeout (500'000 microseconds).

### Example failure

https://github.com/facebookincubator/axiom/actions/runs/24513051374/job/71649369546

mbasmanova:
Summary:
- Increase `QueryTestBase::waitForCompletion` timeout from 50ms to 500ms to match `LocalRunnerTest`
- Check the return value of `waitForCompletion` instead of silently swallowing timeout exceptions
- Update `LocalRunnerTest` to consistently check the return value

The 50ms timeout was set during the initial repo sync and was never updated when `waitForCompletion` was rewritten in https://github.com/facebookincubator/axiom/issues/835. On loaded CI runners, 50ms can be insufficient for async task cleanup, causing `SharedArbitrator::shutdown` to fail with "Unexpected alive participants on destruction".

Fixes https://github.com/facebookincubator/axiom/issues/1257

Reviewed By: jagill

Differential Revision: D101180582

Pulled By: mbasmanova

### DIFF
--- a/axiom/optimizer/tests/QueryTestBase.cpp
+++ b/axiom/optimizer/tests/QueryTestBase.cpp
@@ -105,12 +105,13 @@ logical_plan::LogicalPlanNodePtr QueryTestBase::parseSelect(
 }
 
 namespace {
+constexpr int32_t kWaitTimeoutUs = 500'000;
+
 void waitForCompletion(const std::shared_ptr<runner::LocalRunner>& runner) {
   if (runner) {
-    try {
-      runner->waitForCompletion(50000);
-    } catch (const std::exception& /*ignore*/) {
-    }
+    VELOX_CHECK(
+        runner->waitForCompletion(kWaitTimeoutUs),
+        "Timed out waiting for query completion");
   }
 }
 } // namespace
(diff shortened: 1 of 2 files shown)


jagill (approved):
Thanks!

PR #1260, opened 2026-04-16
fix: Remove TableLayout from SystemConnector TableHandle

arhimondr:
Summary: TableLayout is not serializable

Reviewed By: mbasmanova

Differential Revision: D101222712

### DIFF
--- a/axiom/connectors/system/SystemConnectorMetadata.cpp
+++ b/axiom/connectors/system/SystemConnectorMetadata.cpp
@@ -107,7 +107,7 @@ velox::connector::ConnectorTableHandlePtr SystemTableLayout::createTableHandle(
   // System connector does not support filter pushdown.
   rejectedFilters = std::move(filters);
   return std::make_shared<SystemTableHandle>(
-      connectorId(), *this, std::move(columnHandles));
+      connectorId(), table().name(), std::move(columnHandles));
 }
 
 // ===================== SystemTable =====================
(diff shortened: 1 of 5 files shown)


axiom/connectors/system/SystemConnectorMetadata.cpp
```diff
@@ -92,7 +92,7 @@ velox::connector::ConnectorTableHandlePtr SystemTableLayout::createTableHandle(
   // System connector does not support filter pushdown.
   rejectedFilters = std::move(filters);
   return std::make_shared<SystemTableHandle>(
-      connectorId(), *this, std::move(columnHandles));
+      connectorId(), table().name().toString(), std::move(columnHandles));
```
mbasmanova: table().name().toString() - we should not use toString() - this is for logging only

mbasmanova (approved):

axiom/connectors/system/SystemConnectorMetadata.cpp
arhimondr: Good point. Let me pass the whole SchemaTableName instead. There is a chance we may need to introduce more schemas that the default.

mbasmanova: We already have 2 schemas: runtime and metadata - https://github.com/facebookincubator/axiom/blob/main/axiom/connectors/system/README.md#schemas-and-tables

PR #1265, opened 2026-04-18
feat(connectors): Add SplitWeight helper

pradeepvaka:
Summary:
`velox::connector::ConnectorSplit::splitWeight` is a raw `int64_t` with no established convention for what its values mean. Different connectors emit different scales (some leave it at the default 0, others hardcode 1, others count bytes), making it hard for downstream consumers to compare or aggregate weights consistently.

Adds an `axiom::connector::SplitWeight` value-helper that anchors a canonical scale and centralizes the size→weight math:

- `kStandard` (== 100): weight of one full-sized split. Mirrors `com.facebook.presto.spi.SplitWeight#UNIT_VALUE` so values flowing through Presto-compatible producers retain their meaning.
- `kMin` (== 1): smallest legal weight. Tiny splits clamp to this so they remain visible to consumers instead of degenerating to 0.
- `normalize(fraction)`: maps a fractional weight in (0, 1] onto `[kMin, kStandard]`, rounding up.
- `forSize(splitSize, standardSize, minFraction)`: convenience helper for file-based connectors weighting splits proportional to byte length, clamped to a configured minimum fraction.

Note this is a value-helper, not a wrapper around `ConnectorSplit::splitWeight`. The contract is enforced by convention plus runtime validation downstream (e.g.  scheduler warns when a connector still emits `splitWeight=0`); compile-time enforcement would require typing the field upstream in Velox.

No call sites in this diff.

Reviewed By: arhimondr

Differential Revision: D101397838

### DIFF
--- a/axiom/connectors/tests/CMakeLists.txt
+++ b/axiom/connectors/tests/CMakeLists.txt
@@ -38,7 +38,7 @@ target_link_libraries(
   GTest::gtest_main
 )
 
-add_executable(axiom_connectors_tests SchemaResolverTest.cpp)
+add_executable(axiom_connectors_tests SchemaResolverTest.cpp SplitWeightTest.cpp)
 
 add_test(axiom_connectors_tests axiom_connectors_tests)
 
--- a/axiom/connectors/CMakeLists.txt
+++ b/axiom/connectors/CMakeLists.txt
@@ -26,7 +26,8 @@ if(AXIOM_BUILD_TESTING)
   add_subdirectory(tests)
 endif()
 
-add_library(axiom_connectors ConnectorMetadata.cpp SchemaResolver.cpp)
+add_library(axiom_connectors ConnectorMetadata.cpp SchemaResolver.cpp
+                             SplitWeight.cpp)
 
 target_link_libraries(
   axiom_connectors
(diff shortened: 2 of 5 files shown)


amitkdutta (approved):

axiom/connectors/tests/CMakeLists.txt
```diff
@@ -43,3 +43,9 @@ add_executable(axiom_connectors_tests SchemaResolverTest.cpp)
 add_test(axiom_connectors_tests axiom_connectors_tests)
 
 target_link_libraries(axiom_connectors_tests axiom_connectors axiom_test_connector gtest gtest_main)
+
+add_executable(axiom_split_weight_test SplitWeightTest.cpp)
```
mbasmanova: See https://github.com/facebookincubator/axiom/blob/main/CONTRIBUTING.md#cmake-guidelines

> Avoid introducing new CMake targets unless there is a clear reason to do so.

CC: @kgpai @pratikpugalia

pradeepvaka: Thanks for review. Removed new Cmake target and added to existing targets.

mbasmanova (changes requested):
> Adds a small header-only axiom::connector::SplitWeight helper 

This doesn't match the code. Why we need this helper? Should we instead define a clearer contract for existing weight?

pradeepvaka:
@mbasmanova - updated the diff summary to remove the header-only comment and added some more details.   Regarding a stronger contract, I think compile-time enforcement in Velox with typed field is a way. I can take that as follow up.

cc @arhimondr

PR #1306, opened 2026-04-30
fix: Set presto.array_agg.ignore_nulls=true in SqlQueryRunner

natashasehgal:
Summary:
Add `presto.array_agg.ignore_nulls=true` as a default session property in `SqlQueryRunner::initialize()`. Without this, `array_agg` includes null inputs in the result array, producing incorrect results for queries that rely on null-skipping — e.g. `COALESCE(array_agg(if(cond, val, null)), ARRAY[])` returns `[null]` instead of `[]`.

```sql
SELECT array_agg(x) FROM (VALUES null) t(x)
-- Before: [null]
-- After:  NULL
```

Differential Revision: D103221465


amitkdutta (approved):
Thanks @natashasehgal

PR #1311, opened 2026-05-01
feat(pyspark): Translate Spark explode and posexplode generators (#1311)

kKPulla:
Summary:

Adds Spark Connect translation for the explode and posexplode generator
functions over arrays and maps. Honors alias preservation, with
posexplode emitting a 0-based position column to match Spark semantics.

The explode_outer and posexplode_outer variants are not yet supported
and throw a clear NYI error rather than silently returning incorrect
results.

Multiple generators in a single Project are rejected, matching Spark's
"Only one generator allowed per select clause" behavior.

Differential Revision: D100885457

### DIFF
--- a/axiom/pyspark/SparkToAxiom.cpp
+++ b/axiom/pyspark/SparkToAxiom.cpp
@@ -36,6 +36,75 @@ using namespace facebook;
 namespace axiom::collagen {
 namespace {
 
+bool isGeneratorFunction(const std::string& functionName) {
+  return functionName == "explode" || functionName == "posexplode" ||
+      functionName == "explode_outer" || functionName == "posexplode_outer";
+}
+
+bool isOuterGenerator(const std::string& functionName) {
+  return functionName == "explode_outer" || functionName == "posexplode_outer";
+}
+
+/// Extracts the generator function name from an expression (direct or aliased).
+/// Returns empty string if no generator is found.
+std::string findGeneratorFunction(
+    const spark::connect::Expression& expression) {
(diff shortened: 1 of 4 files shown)


axiom/pyspark/SparkToAxiom.h
```diff
@@ -210,6 +210,12 @@ class SparkToAxiom : public SparkPlanVisitor {
       SparkPlanVisitorContext& context,
       std::vector<ExprPtr>& params,
       std::vector<facebook::velox::TypePtr>& paramTypes);
+
+  /// Handles a Project containing a generator function (explode/posexplode)
```
pedroerp: maybe add more details on how this works? I guess this splits the expression into two nodes, one unnest and one project?

axiom/pyspark/SparkToAxiom.cpp
```diff
@@ -203,6 +236,177 @@ void SparkToAxiom::visit(
...
+    projNames.emplace_back(std::move(colName));
+  }
+
+  auto projectNode =
+      std::make_shared<facebook::axiom::logical_plan::ProjectNode>(
+          nextId(),
+          std::move(planNode_),
+          std::move(projNames),
+          std::move(projExprs));
+  setPlanNode(projectNode, context.planId);
```
pedroerp: the logic in the function looks good, but it's rather long. Are there parts of this function that would make sure to factor out to make them reusable?

pedroerp (approved):

PR #1313, opened 2026-05-03
refactor: Remove bucket parameter from the SplitSource interface (#1313)

arhimondr:
Summary:

Not needed now. In the future we may decide to implement per-bucket split
buffering in the engine.

Reviewed By: amitkdutta

Differential Revision: D103426997


amitkdutta (approved):
Thanks @arhimondr

PR #1350, opened 2026-05-12
feat: Support basic read ducklake table

PingLiuPing:
The DuckLake connector lets Axiom query DuckLake tables by reading DuckLake
metadata from the catalog database and executing the live Parquet files through
Velox's Iceberg reader path.


mbasmanova:
@PingLiuPing CI is red.

CC: @amitkdutta @xiaoxmeng @arhimondr

PingLiuPing:
> @PingLiuPing CI is red.
> 
> CC: @amitkdutta @xiaoxmeng @arhimondr

@mbasmanova Thank you for reviewing. It is expected to fail, we need to land https://github.com/facebookincubator/velox/pull/17380 first which is upgrading duckdb to v1.5.2. And this version supports ducklake extension.

mbasmanova (changes requested):
Thank you for adding DuckLake support! A few high-level comments.

README and PR description: "executing the live Parquet files through Velox's Iceberg reader path" — "executing files" is strange wording. Also, DuckLake and Iceberg are different formats. The connector reuses Velox's Iceberg reader infrastructure to scan Parquet files, but that's an implementation detail. Consider something like: "The DuckLake connector lets Axiom query DuckLake tables by reading catalog metadata from a DuckLake database and scanning the underlying Parquet data files."

The connector reuses Velox's Iceberg connector for execution (no DuckLake code in Velox). DuckLake and Iceberg are different formats — the reuse works because both need Parquet field-id-based column binding, which `IcebergColumnHandle` supports but `HiveColumnHandle` does not. The coupling is significant — DuckLake constructs Iceberg column handles, creates Iceberg splits with empty delete file lists and unused partition keys, and routes all reads through the Iceberg execution path. Any evolution of Iceberg's read path (e.g. adding delete file support, changing split semantics) could break DuckLake. This feels like a short-term band-aid. Long-term, field-id-based Parquet reading should probably be supported at the Hive connector level, so DuckLake (and any other format that uses Parquet with field IDs) doesn't need to pretend to be Iceberg. Please document this coupling and its rationale clearly in the README, and file a follow-up issue to factor out field-id support from the Iceberg connector.

The top-level README lists all connectors (line 234-237) but the new DuckLake connector is not included. Please add it alongside the others.

CLI changes:
- The try/catch wrapping the entire `main` is unrelated to DuckLake — it changes error handling for all connectors. Should be a separate PR if needed.
- We should not continue adding ad-hoc flags (`--ducklake_catalog`) to enable new connectors. Instead, add `--etc-dir` to specify a directory path to connector config files, similar to how Presto handles catalog configuration. Each connector gets a `.properties` file in that directory. This scales to any number of connectors without modifying CLI code. Please submit the `--etc-dir` support as a separate PR first, then use it here.

`DuckLakeCatalogClient` interpolates user-provided schema/table names into SQL strings and escapes them with `quoteSqlStringLiteral`. DuckDB supports prepared statements (`connection->Prepare()` + `prepared->Execute(value1, value2)`), which would eliminate SQL injection risk entirely and make `quoteSqlStringLiteral` unnecessary.

Tests use `GTEST_SKIP` when the DuckLake extension isn't available, but there's nothing in the build system ensuring the extension is installed. CI could pass green with zero DuckLake test coverage. Is the extension guaranteed to be available in CI? If not, how do we ensure these tests actually run?

`catalog_->loadTable()` is called twice per table scan — once in `co_listPartitions` and again in `getSplitSource`. This reads all table metadata (schema, paths, data files) from DuckDB twice. Consider caching the result from `co_listPartitions` in `DuckLakePartitionHandle` and passing it through to `getSplitSource`.

PingLiuPing:
@mbasmanova Thank you for the comments.

`The coupling is significant — DuckLake constructs Iceberg column handles, creates Iceberg splits with empty delete file lists and unused partition keys, and routes all reads through the Iceberg execution path. Any evolution of Iceberg's read path (e.g. adding delete file support, changing split semantics) could break DuckLake. This feels like a short-term band-aid. Long-term, field-id-based Parquet reading should probably be supported at the Hive connector level, so DuckLake (and any other format that uses Parquet with field IDs) doesn't need to pretend to be Iceberg.`

I have a PR for this months ago, https://github.com/facebookincubator/velox/pull/14851, will rebase it and refine the design later.

PR #1353, opened 2026-05-13
feat(parser): Add SHOW TABLES support (#1353)

amitkdutta:
Summary:

X-link: https://github.com/facebookincubator/axel/pull/36

Add SHOW TABLES support following the SHOW SCHEMAS pattern.

Behavior:
- SHOW TABLES returns a single column "Table" containing table names only — no schema or catalog prefix. LIKE applies to the table name only.
- Results are sorted alphabetically by the parser for deterministic output, matching SHOW SCHEMAS and Presto Java behavior. The listTableNames() API does not guarantee ordering.
- listTableNames() is pure virtual — every connector must explicitly decide. Connectors with unbounded or dynamically resolved table sets (e.g., metastore-backed catalogs) return an empty list.
- Connectors with finite, locally loaded tables (TPC-H, test, local hive, system, impulse) return their full table list.

Error handling (matches Presto Java):
- SHOW TABLES FROM <unknown_schema> fails with "Schema does not exist" (validated via schemaExists before listing).
- SHOW TABLES FROM <unknown_catalog>.<schema> fails with "ConnectorMetadata not registered".
- SHOW TABLES FROM a.b.c (3+ part qualifier) fails with "Invalid schema reference".
- SHOW TABLES LIKE <pattern> ESCAPE <invalid> fails with "Escape string must be a single character" (empty or multi-char escape).

Supported forms:
- SHOW TABLES — tables in the current schema
- SHOW TABLES FROM <schema> — tables in a specific schema
- SHOW TABLES FROM <catalog>.<schema> — tables in a specific catalog and schema
- SHOW TABLES LIKE <pattern> — filter by LIKE pattern
- SHOW TABLES FROM <schema> LIKE <pattern> — combined

Reviewed By: mbasmanova

Differential Revision: D104926294

### DIFF
--- a/axiom/connectors/ConnectorMetadata.h
+++ b/axiom/connectors/ConnectorMetadata.h
@@ -987,6 +987,13 @@ class ConnectorMetadata {
       const ConnectorSessionPtr& session,
       const std::string& schemaName) = 0;
 
+  /// Returns table names in the given schema. Connectors with unbounded
+  /// table sets (e.g., dynamic resolution) return empty. Results are not
+  /// guaranteed to be in any particular order.
+  virtual std::vector<std::string> listTableNames(
+      const ConnectorSessionPtr& session,
+      const std::string& schemaName) = 0;
+
   /// Creates a schema with the given name and properties. If 'ifNotExists' is
   /// true, succeeds silently when the schema already exists. Otherwise, raises
   /// a user error if the schema already exists.
--- a/axiom/cli/tests/SqlQueryRunnerTest.cpp
+++ b/axiom/cli/tests/SqlQueryRunnerTest.cpp
@@ -110,6 +110,17 @@ class SqlQueryRunnerTest : public ::testing::Test, public test::VectorTestBase {
         makeRowVector({"Schema"}, {makeFlatVector(expected)}));
   }
 
+  void assertTables(
+      const std::vector<std::string>& expected,
+      const std::string& query = "SHOW TABLES") {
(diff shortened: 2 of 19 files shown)


mbasmanova (approved):

axiom/connectors/ConnectorMetadata.h
```diff
@@ -987,6 +987,13 @@ class ConnectorMetadata {
       const ConnectorSessionPtr& session,
       const std::string& schemaName) = 0;
 
+  /// Returns table names in the given schema. Connectors with unbounded
+  /// table sets (e.g., dynamic resolution) return empty. Results are not
```
PingLiuPing, line 991: > /// Returns table names in the given schema. Connectors with unbounded 
> /// table sets (e.g., dynamic resolution) return empty. 

Should we return a error/warning message instead?

PingLiuPing (approved):

axiom/connectors/ConnectorMetadata.h
amitkdutta: Thanks for the question @PingLiuPing . I am fine with both, We can throw saying dyanmic connectors does not support table listing as there can be many many tables. @mbasmanova any concern throwing instead of returning empty result?

mbasmanova: @amitkdutta @PingLiuPing Throwing might be better. Otherwise, it might be confusing.

PR #1359, opened 2026-05-14
feat: Add --etc_dir config to axiom_sql

PingLiuPing:
Before this change, Axiom CLI only supported one configurable external catalog path: --data_path for a local Hive catalog. That means users could not declaratively define arbitrary connector catalogs, and the CLI was limited to the built-in TPCH/Test/System connectors plus the Hive shorthand.

This PR expands CLI catalog configuration to support any supported connector via [--etc_dir/*.properties] files. Users can now define connector catalogs in [.properties] files and register multiple catalogs at startup, **without adding new flags for every connector type**.

**Why this matters**
Makes catalog configuration scalable and declarative
Keeps --data_path as a convenient shorthand for a single local Hive catalog
Allows multiple catalogs to coexist and be selected explicitly via [--catalog] or `USE`
Removes the need to hardcode connector registration behavior into CLI flag handling

**Behavior changes**
--data_path and --etc_dir are mutually exclusive
--data_path still registers a local Hive catalog
--etc_dir loads connector catalog definitions from [.properties] files

Default catalog selection is simple:
use [--catalog] if provided
otherwise, use Hive only when --data_path is set
otherwise, default to TPCH

### DIFF
--- a/axiom/cli/AxiomSql.cpp
+++ b/axiom/cli/AxiomSql.cpp
@@ -14,9 +14,12 @@
  * limitations under the License.
  */
 
+#include <fmt/ranges.h>
 #include <folly/init/Init.h>
 #include <gflags/gflags.h>
-#include <iostream>
+#include <set>
+#include <unordered_map>
+#include "axiom/cli/CatalogProperties.h"
 #include "axiom/cli/Connectors.h"
 #include "axiom/cli/Console.h"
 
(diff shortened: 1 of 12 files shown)


PingLiuPing:
@mbasmanova Could you please take a look? Thank you.

mbasmanova (changes requested):
@PingLiuPing, Thank you!

**`--data_path` and `--etc_dir` should be mutually exclusive.** The current code allows both, and the README documents the interaction. This is confusing — `--data_path` is a shorthand for registering a single hive catalog, which `--etc_dir` can do as well. Please make them mutually exclusive and fail with a clear error if both are specified.

**Test namespace.** The test uses `using namespace facebook::axiom::cli; namespace { ... }`. Other tests in the same directory nest the anonymous namespace inside the product namespace (e.g., `namespace axiom::sql { namespace { ... } }`). Please follow the same pattern.

**`defaultSchemaForConnector` seems unnecessary.** Two of the three call sites pass hardcoded strings (`"tpch"` → `"tiny"`, `"test"` → `"default"`), which could just be inlined. Consider removing the method and inlining the values at the call sites.

**Catalog vs connector terminology.** `CatalogProperties` correctly uses `catalogName` (file stem, e.g., "analytics") and `connectorName` (type, e.g., "hive"). But in `AxiomSql.cpp` the terms get mixed: `connectorIds` stores catalog names, the `registerCatalog` lambda takes `connectorId` but receives `catalogProperties.catalogName`, and the error says "Connector catalog is already registered". I understand Velox's API uses `connectorId` for what Presto calls "catalog name", but the PR's own variables and messages should be consistent. Consider renaming the lambda from `registerCatalog` to `registerConnector` (since it registers a connector), shortening the parameter names to `id`, `name`, `config` (the lambda context makes the meaning clear), and fixing the error message.

PingLiuPing:
@mbasmanova Thank you for reviewing, I've resolved your comments.

mbasmanova (changes requested):
Thank you for the updates!

**PR description.** Focuses on mechanical details but misses the big picture. Lead with: what was possible before (only `--data_path` for local Hive), what's possible now (configure any supported connector via `.properties` files), and why (scalable catalog configuration). Drop implementation details like "shared catalog property parsing" and "config-driven connector registration".

**Default catalog auto-selection.** Auto-selecting a default catalog from `--etc_dir` is surprising. The rule should stay simple: `--catalog` if specified, `hive` if `--data_path` is set, `tpch` otherwise. Catalogs from `--etc_dir` are available but the user must select one explicitly via `--catalog` or `USE`. This also eliminates `markAsExternal` and `externalCatalogIds`.

**`--data_path` handling in `AxiomSql.cpp`.** The old code called `registerLocalHiveConnector(FLAGS_data_path, FLAGS_data_format)`. The new code manually builds the config map — 12 lines of boilerplate. `registerLocalHiveConnector` still exists and delegates to `registerConnector` internally. Please use it.

**`registerConfiguredConnector`** — drop "Configured", just `registerConnector`. Also, the doc comment describes what the caller does ("matching the `connector.name` entry from a Presto-style catalog properties file"), not what the method does. Please rewrite to document this method's behavior.

**`using ConnectorConfig`** — unnecessary alias. Just use `std::unordered_map<std::string, std::string>` directly, as `CatalogProperties` already does. If kept, move to `.cpp` — it's an implementation detail.

**Simplify properties parser.** Drop `!` as comment prefix and `:` as separator — there's no need for these, and `:` can cause surprises with values containing colons.

**`loadCatalogProperties`** — take `std::string_view` instead of `const std::string&`.

**`try_emplace` for file format default.** `VELOX_USER_CHECK` for `kLocalDataPath` (required) next to `try_emplace` for `kLocalFileFormat` (silently defaults to "parquet") is confusing — one fails loudly, the other silently fills in a value. The default isn't documented on `kLocalFileFormat` in `HiveMetadataConfig.h` either. Please document the default at the constant declaration.

**`VELOX_USER_CHECK(externalCatalogIds.size() <= 1, ...)`** — use `VELOX_USER_CHECK_LE`.

**README.**
- The intro ("Interactive SQL command line for...") — the set of supported data sources hasn't changed, please revert.
- "connector catalog" is not a thing, just "catalog".
- "Additional catalogs can be configured... Hive can also be configured through..." — awkward repetition of "configured". Please rephrase.
- `--etc_dir` flag description in both the README table and `DEFINE_string` — please mention it's mutually exclusive with `--data_path`.
- Please document the supported `.properties` file format.

mbasmanova (approved):
Thank you for addressing all the feedback!

PR #1423, opened 2026-05-27
build: Skip Velox status badge workflow in forks

mbasmanova:
Forks inherit the scheduled cron trigger and run the badge workflow daily. If a fork has the `GIST_TOKEN` secret, it overwrites the upstream gist with the fork's stale submodule pointer — causing the badge in the upstream README to show an outdated Velox commit.

Add `if: github.repository == 'facebookincubator/axiom'` so the workflow only runs in the upstream repo.


jagill (approved):
lgtm

PR #1435, opened 2026-05-30
feat: Add connector for querying raw files (#1435)

srsuryadev:
Summary:

Adds the file connector — a read-only connector that lets users query raw files (e.g. Parquet as initially supported).

The connector uses a pluggable FileHandler interface so new file formats can be added without modifying the core.

For now, by default Parquet handler shipped in this diff. Later, it can be easily extended to other file type/file formats. Metadata tables are accessed via a $-suffix on the table name

The schema name in the SQL query selects the handler; Example:
```
  SELECT
    name,
    ROUND(SUM(uncompressed_size) / SUM(compressed_size), 2) AS ratio
  FROM file."parquet"."/data/file.parquet$column_chunks"
  GROUP BY name
  HAVING ratio < 2  -- Less than 2x compression
  ORDER BY ratio;
````

```
 SELECT name, encodings, compression, min, max, null_count
  FROM file.\"parquet\".\"$TEST_FILE\$column_chunks\"
  ORDER BY name
```

```
  SELECT
    name,
    ROUND(SUM(uncompressed_size) / NULLIF(SUM(compressed_size), 0), 2) AS ratio
  FROM file.\"parquet\".\"$TEST_FILE\$column_chunks\"
  GROUP BY name
  ORDER BY ratio DESC
```
More details are on this doc:
https://docs.google.com/document/d/19Ism5Tc8FQbmtYBy3TLL5oWNwwh-Iw48YsXkp2usywA/edit?tab=t.0

Differential Revision: D106889121

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,13 +228,17 @@ with Velox. These components are:
   - top-level “optimizer” directory
 - Query Runner capable of orchestrating multi-stage Velox execution.
   - top-level “runner” directory
-- Connector - an extention of Velox Connector APIs to provide functionality
+- Connector - an extension of Velox Connector APIs to provide functionality
   necessary for query parsing and planning.
   - top-level “connectors” directory
-  - [Hive](axiom/connectors/hive/README.md) — Local filesystem connector for Parquet, DWRF, and TEXT (including CSV) files.
-  - [TPC-H](axiom/connectors/tpch/README.md) — Read-only, in-memory TPC-H benchmark data.
-  - [System](axiom/connectors/system/README.md) — Read-only metadata tables for session properties and active queries.
-  - [Test](axiom/connectors/tests/README.md) — In-memory read-write connector for unit testing.
+
+  | Connector | Description |
+  |-----------|-------------|
+  | [Hive](axiom/connectors/hive/README.md) | Local filesystem connector for Parquet, DWRF, and TEXT (including CSV) files. |
+  | [TPC-H](axiom/connectors/tpch/README.md) | Read-only, in-memory TPC-H benchmark data. |
+  | [System](axiom/connectors/system/README.md) | Read-only metadata tables for session properties and active queries. |
+  | [Test](axiom/connectors/tests/README.md) | In-memory read-write connector for unit testing. |
+  | [File](axiom/connectors/file/README.md) | Read-only connector multi-utility for running sql on raw files |
 - [CLI](axiom/cli/README.md) - Interactive SQL command line for executing
   queries against in-memory TPC-H dataset and local Hive data.
   - top-level “cli” directory
@@ -317,7 +321,7 @@ Axiom integrates Velox as a Git submodule, referencing a specific commit of the
 Velox repository. The Velox badge at the top of this README shows the current
 commit and how far behind it is from Velox main.
 
-[See what changed since the current Velox commit.](https://github.com/facebookincubator/velox/compare/505a21e7c5f8970e219bdac44fad3c9cc2313010...main)
+[See what changed since the current Velox commit.](https://github.com/facebookincubator/velox/compare/457bb8ffd7674d4b2396fc8486451f1acc7ccd2b...main)
 <!-- pre-commit check-velox-readme validates the SHA above matches the submodule -->
 
 Advance Velox when your changes depend on code in Velox that
--- a/axiom/cli/AxiomSql.cpp
+++ b/axiom/cli/AxiomSql.cpp
@@ -116,6 +116,7 @@ int main(int argc, char** argv) {
 
   // Register after initialize() so sessionConfig() is available.
   connectors.registerSystemConnector(runner.sessionConfig());
+  connectors.registerFileConnector();
 
   axiom::sql::Console console{runner};
   console.initialize();
--- a/axiom/cli/CMakeLists.txt
+++ b/axiom/cli/CMakeLists.txt
@@ -37,6 +37,8 @@ target_link_libraries(
   axiom_graphviz_printers
   axiom_hive_connector_metadata
   axiom_tpch_connector_metadata
+  axiom_file_connector
+  axiom_file_handler_parquet
   axiom_system_connector
   axiom_test_connector
   axiom_sql_presto_parser
--- a/axiom/cli/Connectors.cpp
+++ b/axiom/cli/Connectors.cpp
@@ -19,6 +19,9 @@
 #include <folly/system/HardwareConcurrency.h>
 #include "axiom/common/SessionConfig.h"
 #include "axiom/connectors/ConnectorMetadataRegistry.h"
+#include "axiom/connectors/file/FileConnector.h"
+#include "axiom/connectors/file/core/FileConnectorMetadata.h"
+#include "axiom/connectors/file/parquet/ParquetFileHandler.h"
 #include "axiom/connectors/hive/HiveMetadataConfig.h"
 #include "axiom/connectors/hive/LocalHiveConnectorMetadata.h"
 #include "axiom/connectors/system/SystemConnector.h"
@@ -232,6 +235,18 @@ void Connectors::registerSystemConnector(
   connector::ConnectorMetadataRegistry::global().insert(
       connector->connectorId(),
       std::make_shared<connector::system::SystemConnectorMetadata>(
+          connector.get()));
+}
+
+void Connectors::registerFileConnector(const std::string& connectorId) {
+  connector::file::registerParquetHandler();
+
+  auto connector =
+      std::make_shared<connector::file::FileConnector>(connectorId);
+  registerConnector(connector);
+  connector::ConnectorMetadataRegistry::global().insert(
+      connector->connectorId(),
+      std::make_shared<connector::file::FileConnectorMetadata>(
           connector.get()));
 }
 
--- a/axiom/cli/Connectors.h
+++ b/axiom/cli/Connectors.h
@@ -101,6 +101,9 @@ class Connectors {
       const SessionConfig& sessionConfig,
       const std::string& connectorId = kSystemConnectorId);
 
+  /// Registers the file connector for querying raw files via SQL.
+  void registerFileConnector(const std::string& connectorId = "file");
+
  protected:
   /// Initialize file formats and ioExecutor. Must be called before
   /// registering any connectors.
--- a/axiom/connectors/file/CMakeLists.txt
+++ b/axiom/connectors/file/CMakeLists.txt
@@ -12,28 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-add_library(axiom_connectors ConnectorMetadata.cpp ConnectorMetadataRegistry.cpp SchemaResolver.cpp)
+add_subdirectory(core)
+add_subdirectory(parquet)
+
+# Core connector library (format-agnostic).
+add_library(axiom_file_connector FileConnector.cpp FileHandler.cpp)
 
 target_link_libraries(
-  axiom_connectors
-  axiom_common
-  velox_common_base
-  velox_memory
-  velox_connector
-  velox_core
-  Folly::folly
+  axiom_file_connector
+  PUBLIC
+    axiom_file_connector_core
+    axiom_connectors
+    velox_connector
+    velox_memory
+    velox_common_base
+    Folly::folly
 )
-
-add_subdirectory(file)
-add_subdirectory(system)
-
-if(VELOX_ENABLE_HIVE_CONNECTOR)
-  add_subdirectory(hive)
-endif()
-
-if(VELOX_ENABLE_TPCH_CONNECTOR)
-  add_subdirectory(tpch)
-endif()
 
 if(AXIOM_BUILD_TESTING)
   add_subdirectory(tests)
--- a/axiom/connectors/file/FileConnector.cpp
+++ b/axiom/connectors/file/FileConnector.cpp
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "axiom/connectors/file/FileConnector.h"
+
+#include "axiom/connectors/file/FileHandler.h"
+#include "axiom/connectors/file/core/FileTable.h"
+#include "velox/common/base/Exceptions.h"
+
+namespace facebook::axiom::connector::file {
+
+std::unique_ptr<velox::connector::DataSource> FileConnector::createDataSource(
+    const velox::RowTypePtr& outputType,
+    const velox::connector::ConnectorTableHandlePtr& tableHandle,
+    const velox::connector::ColumnHandleMap& columnHandles,
+    velox::connector::ConnectorQueryCtx* connectorQueryCtx) {
+  auto* fileHandle = dynamic_cast<const FileTableHandle*>(tableHandle.get());
+  VELOX_CHECK_NOT_NULL(fileHandle, "Expected FileTableHandle");
+
+  auto& fileHandler = handler(fileHandle->schemaTableName().schema);
+  return fileHandler.create(
+      fileHandle->filePath(),
+      fileHandle->suffix().empty()
+          ? std::nullopt
+          : std::optional<std::string>(fileHandle->suffix()),
+      outputType,
+      fileHandle->fullSchema(),
+      columnHandles,
+      connectorQueryCtx->memoryPool());
+}
+
+} // namespace facebook::axiom::connector::file
--- a/axiom/connectors/file/FileConnector.h
+++ b/axiom/connectors/file/FileConnector.h
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "velox/connectors/Connector.h"
+
+namespace facebook::axiom::connector::file {
+
+/// Read-only Velox connector for querying raw files.
+/// See README.md for architecture and usage details.
+class FileConnector : public velox::connector::Connector {
+ public:
+  explicit FileConnector(const std::string& id) : Connector(id) {}
+
+  /// Creates a DataSource by dispatching to the file-type-specific
+  /// handler identified by the table handle's schema name.
+  std::unique_ptr<velox::connector::DataSource> createDataSource(
+      const velox::RowTypePtr& outputType,
+      const velox::connector::ConnectorTableHandlePtr& tableHandle,
+      const velox::connector::ColumnHandleMap& columnHandles,
+      velox::connector::ConnectorQueryCtx* connectorQueryCtx) override;
+
+  std::unique_ptr<velox::connector::DataSink> createDataSink(
+      velox::RowTypePtr,
+      velox::connector::ConnectorInsertTableHandlePtr,
+      velox::connector::ConnectorQueryCtx*,
+      velox::connector::CommitStrategy) override {
+    VELOX_UNSUPPORTED("FileConnector does not support writes");
+  }
+};
+
+} // namespace facebook::axiom::connector::file
--- a/axiom/connectors/file/FileHandler.cpp
+++ b/axiom/connectors/file/FileHandler.cpp
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "axiom/connectors/file/FileHandler.h"
+
+#include <map>
+
+#include "velox/common/base/Exceptions.h"
+
+namespace facebook::axiom::connector::file {
+
+namespace {
+
+std::map<std::string, FileHandler*>& handlerRegistry() {
+  static std::map<std::string, FileHandler*> registry;
+  return registry;
+}
+
+} // namespace
+
+void registerHandler(const std::string& schemaName, FileHandler& handler) {
+  auto [it, inserted] = handlerRegistry().emplace(schemaName, &handler);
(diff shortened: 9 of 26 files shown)

### CODE REVIEW

**Comment by srsuryadev** [comment]
cc: @mbasmanova and @kagamiori

**Review by mbasmanova — CHANGES_REQUESTED** [review]
Thank you for the contribution!

**PR title.** "Add omni-connector for querying any raw files via SQL (here we use Nimble files as an example)" — Nimble isn't an example, it's the only supported format. "Any raw files" sets expectations the code doesn't meet. Suggest: `feat: Add connector for querying Nimble files by path`.

**Naming.** Same issue in the code — "OmniConnector" implies a universal connector for any file format, but the implementation only supports Nimble (guarded by `#ifdef AXIOM_NIMBLE_DEPS`). Consider naming it `NimbleConnector` — the extensibility to other formats can be introduced when it's actually needed.

**No tests.** 1216 lines of new code with no tests. This needs unit tests covering at least: `parseTableName` (data vs metadata suffix parsing), `OmniConnectorMetadata::findTable` (data table, metadata tables, invalid paths), and end-to-end read tests for data and metadata tables with a small Nimble file.

**Reads entire file into memory.** `NimbleDataSource::buildResults()` reads all batches and concatenates into one `RowVector`. For large files this will OOM. The `DataSource::next()` API supports streaming (return one batch at a time, caller calls `next()` repeatedly) — use it.

**Deprecated API.** `OmniConnectorMetadata::findTable` uses `velox::memory::deprecatedAddDefaultLeafMemoryPool()` — use the pool from the connector context instead.

**Section banner comments.** `// ===================== SerDe =====================` appears throughout. Drop these — the code structure is clear from class/function names.

**PR description.** The description doesn't show what this enables. A concrete example would make the value clear, e.g.:

```sql
-- Query data directly from a Nimble file by path:
axiom> SELECT count(*), min(ts), max(ts) FROM omni."default"."/data/events.nimble";

-- Inspect file structure:
axiom> SELECT * FROM omni."default"."/data/events.nimble$fields";
axiom> SELECT * FROM omni."default"."/data/events.nimble$encodings";
```

This is a useful debugging/inspection tool — show it off!

**Documentation.** The new connector needs a README (see [hive](https://github.com/facebookincubator/axiom/blob/main/axiom/connectors/hive/README.md), [tpch](https://github.com/facebookincubator/axiom/blob/main/axiom/connectors/tpch/README.md), [system](https://github.com/facebookincubator/axiom/blob/main/axiom/connectors/system/README.md) for the pattern). In particular, the `$encodings` and `$fields` synthetic tables need to be documented — their schemas, what each column means, and example queries.

**Design doc.** Key design decisions from the doc should be in the PR description or the connector's documentation, not behind an internal link.

srsuryadev:
Thank you @mbasmanova  for the quick review! I have addressed all the comments, except wanted to keep the design generic by FileHandlers to onboard new formats or file types. I can add for iceberg metada files and parquet as a follow-up so it can help OSS community as well for now.

mbasmanova:
Thank you for the updates!

On the naming: to justify "omni" as a generic connector, we'd need to see it work with at least two file formats — e.g., Nimble and Parquet. With only one format, the abstraction is untested and may not actually generalize. Adding a Parquet handler would validate the `FileHandler` pattern and demonstrate the connector is truly format-agnostic. Otherwise, we're speculating that the abstraction works.

One observation for a follow-up: users currently have no way to discover what `$`-suffix metadata tables are available for a given file. Since this connector has no catalog, `SHOW TABLES` can't enumerate files. Ideally, something like `SHOW TABLES LIKE '/path/to/file.nimble$%'` would return the available metadata variants (`$encodings`, `$fields`). This would require changes to how `SHOW TABLES` interacts with `listTableNames`, so not for this PR — but worth tracking.

mbasmanova:
On naming: "Omni" is too generic — the Hive connector also supports multiple file formats. This is a debugging/inspection tool for querying individual files by path. Consider a name that captures that, e.g., `file` (as in `SELECT * FROM file.nimble."/path/to/data.nimble"`).

On design: how does the connector know what format the file is? Currently it hardcodes Nimble. To support multiple formats, consider using the schema as the format selector — e.g., `file.nimble."/path"` vs `file.parquet."/path"`. This makes the user's intent explicit, avoids magic-byte detection, and gives each format its own namespace for `$`-suffix metadata tables (which may differ per format).

`OmniConnector.h` is a grab-bag — constants, enum, utility function, three data classes (`OmniColumnHandle`, `OmniTableHandle`, `OmniSplit`), and the connector implementation all in one 179-line header. Split by responsibility — e.g., protocol types (handles, split) in one header, connector in another.

`OmniTableKind` is a hardcoded enum — every format is limited to the same set of metadata tables. But different formats have different internal structures (Nimble has stripes/encodings, Parquet has row groups/page indexes). The metadata table kinds should be format-owned, not centrally enumerated. Using the schema-as-format approach (`file.nimble."/path"`) would let each format define its own `$`-suffix tables independently.

The README describes a "file handler interface" with three functions, but there is no actual interface — `NimbleFileHandler.h` is just three free functions, and the connector dispatches to them via `#ifdef`. To support multiple formats, this needs to be a real abstract base class with a registration mechanism, not `#ifdef`-guarded free functions.

Overall: adding a second format (Parquet) in this PR is essential to figure out the right design and APIs. The abstractions needed for multi-format support are missing — there's a hardcoded enum instead of format-owned metadata, free functions instead of an interface, `#ifdef` instead of registration. A second format would force these to be designed properly.

srsuryadev:
@mbasmanova Thank you very much! Addressed comments in this PR, for the below comment:

> One observation for a follow-up: users currently have no way to discover what suffix metadata tables are available for a given file. Since this connector has no catalog, SHOW TABLES can't enumerate files. Ideally, something like "SHOW TABLES LIKE '/path/to/file.nimble$%'" would return the available metadata variants ($encodings, $fields). This would require changes to how SHOW TABLES interacts with listTableNames, so not for this PR — but worth tracking.

For this discussion, I have created an issue here https://github.com/facebookincubator/axiom/issues/1438. Will try to close this soon as well.

This can be extended to different file types and different modalities of data easily, and use SQL to do "deterministic analysis" on any file types such as images, videos, tensor data, in addition to the storage files which is helpful for lots of agentic workloads as well, in addition to the original intention of debugging storage file formats.

mbasmanova (changes requested):
Great progress — the `FileHandler` interface, schema-as-format, Parquet handler, and `"file"` connector ID are all the right direction.

The PR needs significant design work before it's ready. The functionality is there, but the code reads like a first draft — things are named inconsistently, responsibilities are mixed across files, the build doesn't match the architecture, and the tests don't actually run. Each issue below is an example of a broader pattern, not a one-off fix. Please step back and think about the design holistically rather than addressing these point by point:

**Tests don't run.** `OmniConnectorTest.cpp` has no `add_executable` or `add_test` in CMake. CI passes by never compiling the tests. The test file also unconditionally includes `dwio/nimble/common/tests/NimbleFileWriter.h`, which doesn't exist in the OSS build — so even adding it to CMake would fail. Gate Nimble tests behind `#ifdef AXIOM_NIMBLE_DEPS` so Parquet tests compile and run independently.

**Naming doesn't match identity.** The connector registers as `"file"` and users write `file.nimble."/path"`, but everything in the code says "Omni" — `OmniConnector`, `OmniTableHandle`, `OmniSplit`, `axiom/connectors/omni/`. A reader seeing `file.nimble."/path"` in SQL and `OmniConnector` in code has to guess they're the same thing. Rename classes, files, and directories to match the connector's actual identity.

**File contents don't match filenames.** `OmniSplit.h` contains `OmniColumnHandle`, `OmniTableHandle`, `OmniSplit`, `parseName()`, and `kOmniConnectorId`. A reader looking for the table handle definition wouldn't look in a file called "Split". Each file should have one clear responsibility.

**Build couples connector to handlers.** `axiom_omni_connector` CMake target compiles `ParquetFileHandler.cpp` directly into the connector library. The whole point of the `FileHandler` interface is that the connector doesn't know about specific formats — but the build links them together. Handlers should be separate libraries that register themselves.

**Handlers know too much / too little.** Handlers can't list their supported `$`-suffixes (no `listMetadataTables()` API), so there's no way for the connector to support discoverability. But handlers are tied to a specific schema name at registration time (`registerHandler("nimble", handler)`) — they should be registerable under any name the caller chooses.

**`FileHandler::metadata(suffix)`** returns a schema but is named as if it returns data.

**Comments duplicate docs.** `OmniConnector.h` has a ~60-line architectural comment block with ASCII diagrams, query flow, metadata tables, and a "how to add a new handler" guide — all of which is already in the README. Keep it in one place.

**Unnecessary wrappers.** `OmniDataSourceBase::createFlat<T>()` wraps `BaseVector::create<FlatVector<T>>(type, size, pool)` — a one-liner that doesn't need a helper. `kOmniConnectorId = "file"` is hardcoded in `OmniSplit.h` but the connector ID is passed at registration time — the constant shouldn't exist.

**Coding standards.** Please review the [coding style guide](https://github.com/facebookincubator/velox/blob/main/CODING_STYLE.md) and do a pass for basic compliance. For example, `///` is used for private/internal comments throughout — `///` is only for public API documentation in headers; everything else uses `//`. These are the kinds of issues that should be caught before requesting review.

srsuryadev:
Will wait for OSS CIs to complete and updated.

srsuryadev:
cc: @mbasmanova  thank you for lots of suggestions! Made quite some improvements on the PR more for better maintainability and readability. Let me know your final review/approval

mbasmanova (dismissed):
Thank you for the updates!

**Overall:** We're on round 5 of this review, and each round we're still finding fundamental design issues — repeated code that should be handled by the framework, unnecessary complexity (SerDe for a single-node connector), headers full of `using` aliases and free functions, code that's never compiled. These are things that should be thought through before submitting, not discovered one layer at a time through review. Please do a thorough self-review before the next round — read every file as if you're seeing it for the first time and ask: is this needed? is this in the right place? does this follow the project's conventions?

Comments:
- `IFileHandler` — the `I` prefix for interfaces is a Java/C# convention. Velox C++ style doesn't use it. Just `FileHandler`.
- `kFileConnectorId` is still hardcoded in the header — the connector ID should be passed at registration time, not baked into the library.
- `nimble/NimbleDataSources.cpp` and `nimble/NimbleFileHandler.cpp` exist in the PR but are not in any CMake target — they're never compiled or tested. The README mentions `axiom_file_handler_nimble` as a library, but no such target exists in `CMakeLists.txt`. Code that's never compiled is dead code. Either set up the build to compile and test Nimble, or remove the Nimble handler from this PR and add it when the dependency is available.
- `core/`, `nimble/`, and `parquet/` subdirectories don't have their own `CMakeLists.txt` — they're compiled from the parent via path prefixes (`core/FieldsDataSource.cpp`, `parquet/ParquetDataSources.cpp`). In Axiom and Velox, each directory has its own CMake file. Follow the convention.
- `using` aliases in headers (e.g., `using ColumnHandleMap = ...` in `IFileHandler.h`) — don't put `using` in headers. They leak into every file that includes the header. Use fully qualified names. Check all headers.
- **PR title** still says "querying any raw files" — it supports Nimble and Parquet, not "any" files.
- **PR description** says "even images or or any format" — typo, and no image support exists. The SQL examples have mangled escaping. Please rewrite to accurately describe what the PR does.
- `FileConnector::registerSerDe()` / `serialize()` / `deserialize()` on table handles, column handles, and splits — SerDe is for distributed execution, but this connector is inherently single-node (worker nodes would need the file at the same local path). Remove the SerDe code.
- `FileHandler::create(suffix, outputType, columnHandles, pool, filePath)` — `filePath` is the primary input but it's the last parameter. Put it first. `suffix` should be `std::optional<std::string>` (`std::nullopt` for data) rather than empty string.
- `$fields` metadata table — how is this different from `DESC file.nimble."/path"`? Is the extra DataSource implementation justified, or does `DESC` already cover this? Also, `$fields` uses `IFileHandler::resolve()` which every handler implements — the connector core could provide `$fields` automatically for all handlers instead of requiring each handler to manually dispatch to `FieldsDataSource`.
- `ParquetDataSources.h` (and likely `NimbleDataSources.h`) is a collection of free functions and `using` aliases in a header. Per coding style, use a class with static methods — no free functions, no `using` in headers.
- `constexpr const char* kFields = "fields"` is repeated in each handler. But `$fields` doesn't need handler involvement at all — the connector has `IFileHandler::resolve()` for the schema and `FieldsDataSource` for the data source. The connector core should handle `$fields` automatically, including `metadataSchema`, `listMetadataTables`, and routing. Handlers should only implement format-specific metadata tables like `$encodings` and `$row_groups`.

srsuryadev:
PR is updated, and the failing MacOS CI seems to be flaky issue

srsuryadev:
* Addressed the comments/suggestions
* Removed unnecessary code / reduced the surface area of code
* Updated the docs
* Kept only the OSS needed code in the PR
* Simplified the method names or reduced some of them.

cc: @mbasmanova, @kagamiori

mbasmanova:
CC: @PingLiuPing

mbasmanova:
CC: @apurva-meta

mbasmanova (dismissed):
Thank you for the updates! A few comments:

**Dead Nimble references in OSS.** `Connectors.cpp` still has, behind
`#ifdef AXIOM_NIMBLE_DEPS`, an `#include "axiom/connectors/file/fb/nimble/NimbleFileHandler.h"`
and a `registerNimbleHandler()` call. `AXIOM_NIMBLE_DEPS` is never defined in the
OSS build, and neither the header nor `registerNimbleHandler()` exist in this PR
or the repo — so in OSS this is never compiled or tested. This is the
never-compiled Nimble code from the earlier round. Please drop the
`#ifdef AXIOM_NIMBLE_DEPS` include and registration from this PR and add them
together with the Nimble handler when it lands in OSS — and remove the Nimble
mentions from the docs accordingly (see below).

**No end-to-end test.** Every "read" test drives the data source directly —
`handler_.create(...)`, a hand-built `ColumnHandleMap`, `addSplit()`, `next()`.
Nothing runs an actual SQL query against a file, so the planning + execution path
(`findTable` → `createColumnHandle` → `createTableHandle` →
`FileConnector::createDataSource` → `SplitManager`/`FileSplitSource`) is never
exercised together — a bug in any of that wiring would pass CI. Since the whole
point of the connector is to be queried via SQL, please add an e2e test that
writes a small Parquet file and runs real queries through the runner: a full
scan, a projection, and a `$row_groups`/`$encodings` query, asserting results.
Once that exists, drop the direct-data-source tests — they add maintenance
without adding coverage the e2e test doesn't already give you.

**Metadata tables miss the most useful data, and `$encodings` is misnamed.** For
a Parquet inspection tool, the most useful metadata is the column-level
statistics the format already carries in column-chunk metadata — min / max /
null_count / distinct_count — and none are exposed. Please add them. Separately,
`$encodings` doesn't expose encodings: it surfaces compression codec, sizes, and
value count. Either add the actual column-chunk encoding(s) (`PLAIN`, `RLE`,
`DICTIONARY`, …) or rename the table to match its contents (e.g.
`$column_chunks`).

**The connector README needs a careful pass — it's too long and partly
inaccurate.** Much of it documents internals that belong in the code and will
drift: cut the per-class "Core layer (format-agnostic)" tour and the "Query
flow" pseudocode, and the handler-interface description is given twice. Trim to
usage, table-name format, the metadata-table schemas (the appendix is the part
worth keeping), and how to add a handler. While trimming, fix the factual
errors:
- "read-only **SQL** access to raw files" — SQL is just one front-end; the
  connector is queryable through any front-end (e.g. dataframe). Say read-only
  access to the files.
- "isolated behind the **`IFileHandler`** interface" — it's `FileHandler` now.
- "the core layer handles ... and **SerDe**" — SerDe was removed; the core
  serializes nothing.
- Parquet handler "exposes **three** synthetic metadata tables" — it exposes
  two (`$row_groups`, `$encodings`).
- "Query flow": `handler.create("encodings", ...)` — `create()`'s first arg is
  `filePath`; `encodings` is the suffix.
- The mid-path `$` claim is wrong — see the `parseName` item below.
- Broken link: `[Metadata Tables](#metadata-tables)` has no target heading.
- "BUCK/CMake" → CMake only (BUCK is internal).

**Top-level `README.md`.** The connector table lists File first (ahead of Hive)
and describes it as "querying any file type (Nimble, Parquet or any raw file)".
It's a debugging/inspection connector supporting only Parquet in OSS — list it
last, next to Test, and describe what it actually supports.

**Stale `$fields` comment.** `cli/Connectors.h`: the `registerFileConnector`
comment lists "`$fields`, `$encodings`" — `$fields` was removed; only
`$row_groups` and `$encodings` exist.

**`parseName()`.** Two things:
- It's a free function in a header — make it a static method on `FileTable`
  (`FileTable::parseName()`), per the earlier round about free functions in
  headers.
- It splits on the last `$` regardless of position, so a real path containing
  `$` mis-parses: `/data/$backup/file.parquet` → path `/data/`, suffix
  `backup/file.parquet`, which then fails with "Unsupported metadata table:
  backup/file.parquet". The README claims such a path "is not treated as a
  suffix because it is not at the end" — but there's no such check, and it's
  untested. Fix the behavior (and the doc), and add a test for a `$`-containing
  path with no metadata suffix.

**Test file names don't match contents.**
- `FileSplitTest.cpp` contains only `ParseNameTest` (no `FileSplit` tests).
  `parseName()` lives in `FileTable.h` — move these cases to `FileTableTest.cpp`
  or rename the file.
- `FileConnectorMetadataTest.cpp` defines two suites
  (`FileConnectorMetadataApiTest`, `HandlerRegistryTest`); the registry tests
  belong in a file matching their suite.

**Empty fixture.** `ParseNameTest` is an empty fixture — use `TEST()` instead
of `TEST_F()`.

Minor:
- `ParquetDataSource`'s constructor re-validates `columnHandles.contains(name)`,
  duplicating the check `MetadataDataSource` already performs. Consider lifting
  column/handle validation to one place.

PingLiuPing:
Thanks.

A few general questions/comments:

* It looks like the supported file format has changed from Nimble to Parquet, and Parquet is now the only supported format. If so, we should update the PR title and description accordingly.
* This feels more like a debugging/inspection tool. For Parquet, there are already many existing tools, such as parquet-tools, parquet-cli, and pq. These tools can inspect metadata and show the actual data. One potential advantage of this connector could be its ability to read large Parquet files using streaming reads.
* Does this connector support reading multiple Parquet files? For example:
`file."parquet"."/path/to/parquet/"`
* Does this connector support federated queries? If both this and multiple-file reading are supported, that could make this tool stand out from other Parquet inspection tools.
* Some similar systems that support querying Parquet files for reference:
    * DuckDB: https://duckdb.org/docs/current/data/parquet/overview
    * DataFusion CLI: https://datafusion.apache.org/user-guide/cli/datasources.html#parquet

I’ll provide more feedback later.

apurva-meta:
A few issues should be addressed before landing. 

The most significant is a correctness bug in NimbleFileHandler.cpp: projected reads double-apply column projection — the ColumnSelector already narrows the reader's output to the selected columns, yet readBatch() then re-projects using full-schema indices, which can index past the end of the returned vector (out-of-bounds or wrong column) for any subset projection; the Parquet handler avoids this by projecting only once, so Nimble should be made consistent, and the projection test strengthened beyond just checking row count. 

Second, parseName() in FileTable.h splits on the last $, so legitimate file paths containing a $ (e.g. a $backup directory) are mis-parsed and become unqueryable — it should only treat $ as a suffix delimiter after the final path separator or restrict to known suffixes. Beyond those, the bots have already flagged several items worth resolving: deprecated registerConnector/unregisterConnector usage across the test files, non-resilient std::call_once/std::once_flag in ParquetFileHandler.cpp and NimbleFileHandler.cpp (use the folly equivalents), README inaccuracies (IFileHandler vs FileHandler, the contradictory mid-path $ description, and a nonexistent $fields table in the Connectors.h docstring), duplicated openFile() logic in both handlers' resolve() methods, and minor line-length/unnecessary-copy lint nits.

PingLiuPing:
The PR advertises file."default"."/tmp/data.parquet$encodings" for Parquet, but the implementation only registers the "parquet" schema and the tests explicitly assert that "default" is unsupported.

The Parquet $encodings table is not correct for nested Parquet schemas. It names each column chunk by indexing into reader->rowType() , but row-group column chunks are leaf-level while rowType() is top-level, so nested fields will be mislabeled or fall back to synthetic column_N names.

The schema of $encodings only has a single name column which is probably too weak for nested Parquet metadata. What you really want is something like: path or column_path + leaf_name, possibly max_definition_level / max_repetition_level if the goal is real encoding introspection. Without a full path, nested leaves remain ambiguous.

srsuryadev:
@PingLiuPing Thank you for the review, and chiming in, and adding your thoughts!


Some background context and motivation for this feature:

We wanted to build a generic file connector leveraging composable database libraries such as Axiom and Velox—on any storage file type/format/modality—to perform deterministic analytics on them.

For the initial use case, we tried well-known analytical storage file formats such as Parquet for inspection and debugging.

Right now, the Parquet handler that comes along with this PR is not feature-rich; however, we can extend it as we evolve this feature. The main focus of this PR is on the base FileConnector and FileHandler interfaces, which can be extended for other use cases.

Also, some of your comments are really wonderful—we will definitely incorporate them to make this feature stand out. These suggested enhancements are easy to address since the initial design of the FileConnector is well thought out. Hence, we should be able to implement them easily. We have already documented these suggestions and will share the respective PRs soon and the corresponding documentation for tracking below:

https://github.com/facebookincubator/axiom/issues/1478
https://github.com/facebookincubator/axiom/issues/1479


--------------

@apurva-meta 

Addressed some of the comments as part of @mbasmanova's suggestions as well. For comment on Nimble, for this PR, we wanted to have only the code which is OSS supported in Axiom currently.

apurva-meta (approved):

srsuryadev:
cc: @mbasmanova / @PingLiuPing  for final look for the initial version

mbasmanova:
This shouldn't have landed without a final review pass — my last review was
REQUEST_CHANGES, and the PR changed substantially afterward. That pass would have
caught the items below. Two are correctness bugs in newly-added code and should
be fixed in a prompt follow-up diff, not parked as enhancement issues:

1. **`parseName()` mis-parses paths containing `$`.** The current guard rejects a
   suffix containing `/`, so `/data/$backup/file.parquet` is handled, but a `$`
   in the filename still breaks: `/tmp/foo$bar.parquet` → path `/tmp/foo`, suffix
   `bar.parquet`. Treat `$` as a suffix only after the final path separator, or
   restrict to known suffixes. This isn't tracked by #1478 or #1479.

2. **`$column_chunks` mislabels nested schemas** (#1478). Leaf-level column chunks
   are named by indexing into the top-level `reader->rowType()`, so nested fields
   get the wrong name or a synthetic `column_N`. #1478 captures it, but as
   "support nested schemas" — it's a correctness bug in what shipped, not a new
   feature.

Separately, the test fixture is too trivial to exercise the logic under test:

- **`fullScan` / `projection` don't exercise `ORDER BY`.** The file is written
  pre-sorted on both columns (ids `{1, 2, 3}`, names `{alpha, beta, gamma}`), so
  `ORDER BY id` and `ORDER BY name` are no-ops against the expected results — a
  broken `ORDER BY` would pass both. Write the rows unsorted in both columns
  (e.g. `(3, gamma), (1, alpha), (2, beta)`).
- **`rowGroupsMetadata` / `columnChunksMetadata` use a single row group**, so the
  multi-row-group behavior these tables exist for — multiple rows, incrementing
  `row_group_id`, per-group `num_rows` — is never tested; a bug in the
  `numRowGroups` loop would pass. Write the file with several row groups (small
  row-group size) and assert across them.
- **`columnChunksStatistics` never sees a non-zero `null_count`** — the fixture
  has no nulls, so a miscount (or hard-coded 0) would pass. Add a nullable column.
- **No streaming / multi-batch coverage.** `query()` asserts
  `result.results.size() == 1`, so everything returns in one vector; the
  connector's batch-by-batch streaming read is never exercised with data spanning
  multiple batches.

cc: @apurva-meta

mbasmanova:
A couple more follow-ups, on docs and test coverage:

**`README.md` mixes user-facing and developer-facing docs.** The "Adding a New
File Handler" section (subclass `FileHandler`, implement
`resolve()`/`createDataSource()`, register, add a CMake target, write tests) is
contributor/extension guidance — it belongs in the code, as doc comments on the
`FileHandler` interface in `FileHandler.h`, not the user-facing README. Likewise
the `FileTable::parseName()` implementation note under "Table Name Format": keep
the user rule ("a `$` inside a directory name is part of the path"), drop the
function/algorithm detail. Leave the README to Overview, Usage, Table Name
Format, and the metadata-table reference.

**No end-to-end tests for `DESC <table>` or `SHOW SCHEMAS`.**
`listSchemaNames`/`schemaExists` are covered only at the API level
(`FileConnectorMetadataTest.cpp`), not through SQL, and there's no `DESC` test at
all — so the describe and show-schemas paths a user actually hits are unverified.

srsuryadev:
FYI, @mbasmanova - Addressing the follow-up comments here in the PRs: #1504, #1505, #1506, #1508

PR #1467, opened 2026-06-10
misc: Advance Velox in Axiom (#1467)

natashasehgal:
Summary:

Advance the Velox submodule in Axiom

Reviewed By: mbasmanova

Differential Revision: D108158996


.github/workflows/macos.yml
```diff
@@ -63,6 +63,10 @@ jobs:
           install_velox_deps_from_brew
           install_gflags
           install_glog
+          install_fizz
+          install_wangle
+          install_mvfst
+          install_fbthrift
```
mbasmanova: Do we need this? Curious, why.

natashasehgal: Removed. If we need any package updates will do in another PR

mbasmanova (approved):

PR #1473, opened 2026-06-10
fix(ci): Install FBThrift dep chain in macOS workflow

kgpai:
## Problem

velox#16019 (landed in the velox SHA bumped by #1467) switched velox from Apache Thrift to FBThrift. This added a hard `find_package(fizz | wangle | FBThrift CONFIG REQUIRED)` block to `velox/CMakeLists.txt:684` which fires whenever `VELOX_ENABLE_PARQUET=ON` (always, for axiom).

Axiom's macOS workflow installs Velox's brew-only dep subset (`install_velox_deps_from_brew`) but doesn't call the full `install_velox_deps`, so the FBThrift dep chain isn't built. CMake-configure now fails with:

\`\`\`
CMake Error at velox/CMakeLists.txt:684 (find_package):
  Could not find a package configuration file provided by "fizz"
\`\`\`

Failing run: https://github.com/facebookincubator/axiom/actions/runs/27295062570/job/80625462959

## Fix

Add the FBThrift dep-chain installs (`install_folly`, `install_fizz`, `install_wangle`, `install_fbthrift`) to the macOS workflow's Install Dependencies step.

## Validation

CI on this PR is the test. If `install_folly` turns out to need additional transitive deps (boost, fmt, fast_float, ranges_v3) we'll see them and iterate. The dep functions are sourced from `velox/scripts/setup-common.sh` at the pinned velox SHA — verified to exist.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -59,11 +59,55 @@ jobs:
           CMAKE_POLICY_VERSION_MINIMUM: '3.5'
         run: |
           source velox/scripts/setup-macos.sh
+          # Match the SHARED build axiom configures with below. Without
+          # this, install_folly's conditional adds -DGFLAGS_SHARED=FALSE
+          # which statically embeds gflags into libfolly.a. Axiom then
+          # also links libgflags.dylib via libvelox.so, putting two
+          # copies of gflags in each test binary, which gflags detects
+          # at runtime: "ERROR: something wrong with flag 'flagfile'
+          # … linked both statically and dynamically".
+          export VELOX_BUILD_SHARED=ON
+          # Keep the NDEBUG state of every installed dep in lockstep
+          # with axiom's own build. cmake_install_dir doesn't pass
+          # CMAKE_BUILD_TYPE, so without this folly/fizz/wangle/etc.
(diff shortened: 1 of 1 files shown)


.github/workflows/macos.yml
```diff
@@ -64,6 +64,16 @@ jobs:
           install_gflags
           install_glog
           install_double_conversion
+          # FBThrift dep chain — required by velox/CMakeLists.txt's
+          # find_package(fizz | wangle | FBThrift CONFIG REQUIRED) block,
+          # which is hit whenever VELOX_ENABLE_PARQUET is on (always, for
+          # axiom). Introduced by velox#16019 (switch from Apache Thrift
+          # to FBThrift). install_fbthrift's transitive deps require
+          # folly + fizz + wangle to be built first.
+          install_folly
```
kgpai, line 106: This will still fail, but this is first attempt.

natashasehgal: Thanks for looking into this! I'm seeing a failure in current run: https://github.com/facebookincubator/axiom/actions/runs/27313995235/job/80690404743

amitkdutta (approved):

natashasehgal (approved):

natashasehgal (approved):

PR #1498, opened 2026-06-18
feat(runner): Report query progress during execution (#1498)

singcha:
Summary:

Adds query progress reporting in `axiom/runner` and wires it into the CLI: `SqlQueryRunner::RunOptions` gains an `onProgress` callback that fires while a query runs. Each report carries query-level and per-stage counters: rows and bytes read from storage (scan) and received over exchanges (shuffle), split counts (queued/running/completed), operator CPU time, and an `ExecutionState` (planned/running/finished; failure and cancellation are reported via the completion callback, not here). There is no percentage: the runner adds splits during execution, so a split-based percentage can run backward.

The reporting lives next to the runner it observes and depends only on the `Runner` interface, split into:
- `QueryProgressBuilder` -- stateless `toStageTopology()` (plan fragments to per-stage producer edges) and `computeProgress()` (a `Runner::stats()` snapshot to a `QueryProgress`), both static and unit-tested directly on synthetic stats and fragments. Scan vs shuffle is keyed off `velox::exec::OperatorType` constants. Malformed input fails loudly instead of degrading silently: an operator outside those buckets that reports raw input, an unresolved producer task prefix, and a topology/stats size mismatch are each a `VELOX_CHECK` (with the runtime values at the end of the message, per the coding-style rule). Stages are referenced purely by position -- `producers` are indexes into the stage vector, so no redundant per-stage index is stored. Completion is taken only from the runner's finished flag -- not inferred from snapshot split counts, which grow over time -- and a clean finish normalizes the frame so no stage is left showing running/queued splits.
- `ProgressReporter` -- the RAII handle that polls a `Runner` on a supplied `folly::FunctionScheduler` and emits a final report on teardown from the live `Runner::stats()` (still present until `waitForCompletion`). A misbehaving `onProgress` callback cannot disrupt the query -- its exceptions are logged and swallowed. `poll()` builds the snapshot outside that guard so a builder invariant violation surfaces during execution; the destructor builds inside its own catch because it can run while a failed query unwinds the stack and must never throw.

`fragments()` is pure-virtual on `Runner` and always exposes the plan topology, so `computeProgress()` is always given a stage tree positionally 1:1 with `Runner::stats()`. `SqlQueryRunner` does not own a scheduler -- it takes an optional, non-owned `folly::FunctionScheduler*` constructor arg; a query that sets `onProgress` without a supplied scheduler fails loudly (`VELOX_USER_CHECK_NOT_NULL`) instead of the runner spinning up a hidden thread. `startProgressReporter()` returns the reporter behind a `std::unique_ptr` (`ProgressReporter` is non-movable because its poll closure captures `this`).

The CLI renderer that turns these frames into a live bar is the next diff in the stack.

Differential Revision: D108909838

### DIFF
--- a/axiom/cli/SqlQueryRunner.cpp
+++ b/axiom/cli/SqlQueryRunner.cpp
@@ -15,8 +15,14 @@
  */
 
 #include "axiom/cli/SqlQueryRunner.h"
+#include <folly/container/F14Map.h>
 #include <folly/system/HardwareConcurrency.h>
+#include <algorithm>
 #include <cmath>
+#include <functional>
+#include <memory>
+#include <optional>
+#include <vector>
 #include "velox/common/base/VeloxException.h"
 #include "velox/common/process/ProcessBase.h"
 
@@ -37,6 +43,7 @@
 #include "axiom/optimizer/Plan.h"
 #include "axiom/optimizer/RelationOpPrinter.h"
 #include "axiom/optimizer/VeloxHistory.h"
+#include "axiom/runner/ProgressReporter.h"
 #include "axiom/sql/presto/PrestoParser.h"
 #include "axiom/sql/presto/PrestoSqlError.h"
 #include "axiom/sql/presto/ShowStatsBuilder.h"
@@ -90,6 +97,11 @@ class PhaseTimer {
       runtimeStats_->recordTiming(cpuKey_, std::chrono::nanoseconds(cpuNanos));
     }
   }
+
+  PhaseTimer(const PhaseTimer&) = delete;
+  PhaseTimer& operator=(const PhaseTimer&) = delete;
+  PhaseTimer(PhaseTimer&&) = delete;
+  PhaseTimer& operator=(PhaseTimer&&) = delete;
 
  private:
   QueryRuntimeStats* const runtimeStats_;
@@ -1094,6 +1106,25 @@ connector::ConnectorSessionPtr SqlQueryRunner::makeConnectorSession(
       sessionConfig_->effectiveValues(connectorId));
 }
 
+std::unique_ptr<runner::ProgressReporter> SqlQueryRunner::startProgressReporter(
+    runner::Runner& runner,
+    std::string_view queryId,
+    const RunOptions& options) {
+  if (!options.onProgress) {
+    return nullptr;
+  }
+  VELOX_USER_CHECK_NOT_NULL(
+      progressScheduler_,
+      "onProgress requires a scheduler supplied to SqlQueryRunner");
+  progressScheduler_->start();
+  return std::make_unique<runner::ProgressReporter>(
+      runner,
+      *progressScheduler_,
+      std::string(queryId),
+      options.onProgress,
+      options.progressReportIntervalMs);
+}
+
 std::string SqlQueryRunner::runExplainAnalyze(
     const logical_plan::LogicalPlanNodePtr& logicalPlan,
     const RunOptions& options,
@@ -1134,6 +1165,8 @@ std::string SqlQueryRunner::runExplainAnalyze(
         runtimeStats.get(),
         QueryRuntimeStats::kExecuteWallNanos,
         QueryRuntimeStats::kExecuteCpuNanos);
+    auto progress =
+        startProgressReporter(*runner, queryCtx->queryId(), options);
     auto results = fetchResults(*runner);
   }
 
@@ -1374,6 +1407,8 @@ SqlQueryRunner::SqlResult SqlQueryRunner::runLogicalPlan(
         runtimeStats.get(),
         QueryRuntimeStats::kExecuteWallNanos,
         QueryRuntimeStats::kExecuteCpuNanos);
+    auto progress =
+        startProgressReporter(*runner, queryCtx->queryId(), options);
     result.results = fetchResults(*runner);
   }
 
--- a/axiom/cli/SqlQueryRunner.h
+++ b/axiom/cli/SqlQueryRunner.h
@@ -16,15 +16,19 @@
 #pragma once
 
 #include <folly/executors/CPUThreadPoolExecutor.h>
+#include <folly/executors/FunctionScheduler.h>
 #include <chrono>
 #include <exception>
 #include <functional>
+#include <vector>
 #include "axiom/common/ConfigRegistry.h"
 #include "axiom/common/QueryRuntimeStats.h"
 #include "axiom/common/SessionConfig.h"
 #include "axiom/optimizer/DerivedTable.h"
 #include "axiom/optimizer/ToVelox.h"
 #include "axiom/runner/LocalRunner.h"
+#include "axiom/runner/ProgressReporter.h"
+#include "axiom/runner/QueryProgress.h"
 #include "axiom/sql/presto/SqlStatement.h"
 #include "velox/common/file/TokenProvider.h"
 
@@ -153,7 +157,14 @@ class SqlQueryRunner {
   /// Constructs the runner with the given session user identity. The user
   /// must be non-empty; it is exposed by the SQL `CURRENT_USER` function and
   /// used as the default table owner in DDL.
-  explicit SqlQueryRunner(std::string user) : user_{std::move(user)} {
+  ///
+  /// `progressScheduler` (not owned, must outlive this runner) drives progress
+  /// polling for queries that set RunOptions::onProgress. If null, such a query
+  /// fails loudly rather than running without the progress it asked for.
+  explicit SqlQueryRunner(
+      std::string user,
+      folly::FunctionScheduler* progressScheduler = nullptr)
+      : user_{std::move(user)}, progressScheduler_{progressScheduler} {
     VELOX_USER_CHECK(!user_.empty(), "SqlQueryRunner user must be non-empty");
   }
 
@@ -205,6 +216,16 @@ class SqlQueryRunner {
     /// Called before parsing. Receives query metadata (query ID, SQL text,
     /// creation time, and session catalog/schema).
     QueryStartCallback onStart;
+
+    /// Called to report progress while the query runs; see
+    /// progressReportIntervalMs for cadence.
+    facebook::axiom::runner::QueryProgressCallback onProgress;
+
+    /// Approximate interval between progress reports, in milliseconds (plus a
+    /// best-effort final report at completion). Clamped up to
+    /// runner::ProgressReporter::kMinProgressReportIntervalMs.
+    int32_t progressReportIntervalMs{facebook::axiom::runner::ProgressReporter::
+                                         kDefaultProgressReportIntervalMs};
 
     /// Called after execution completes (success or failure). Carries
     /// telemetry only: timing, query ID, error info, row counts. Results
@@ -393,6 +414,16 @@ class SqlQueryRunner {
       const RunOptions& options,
       facebook::axiom::QueryRuntimeStats& runtimeStats);
 
+  // Returns a ProgressReporter polling `runner` (starting the shared scheduler
+  // first) when options.onProgress is set, otherwise nullptr. Held behind a
+  // unique_ptr because ProgressReporter is non-movable; keep it in scope for as
+  // long as progress should be reported.
+  std::unique_ptr<facebook::axiom::runner::ProgressReporter>
+  startProgressReporter(
+      facebook::axiom::runner::Runner& runner,
+      std::string_view queryId,
+      const RunOptions& options);
+
   // Runs a parsed SQL statement, writing optimize/execute timing into 'timing'
   // and the serialized Velox plan into 'planString'.
   SqlResult runUnchecked(
@@ -458,6 +489,10 @@ class SqlQueryRunner {
   std::string defaultSchema_;
   const std::string user_;
   std::atomic<int32_t> queryCounter_{0};
+
+  // Progress-polling scheduler (see constructor). Started idempotently before
+  // each progress-reporting query.
+  folly::FunctionScheduler* const progressScheduler_;
 
   // Noop stats instance for code paths that don't track runtime metrics.
   facebook::axiom::QueryRuntimeStats noopRuntimeStats_;
--- a/axiom/cli/tests/SqlQueryRunnerTest.cpp
+++ b/axiom/cli/tests/SqlQueryRunnerTest.cpp
@@ -16,7 +16,7 @@
 
 #include "axiom/cli/SqlQueryRunner.h"
 #include <folly/dynamic.h>
-// NOLINTNEXTLINE(facebook-unused-include-check)
+#include <folly/executors/FunctionScheduler.h>
 #include <folly/init/Init.h>
 #include <folly/json.h>
 #include <gmock/gmock.h>
@@ -26,6 +26,7 @@
 #include "axiom/cli/QueryIdGenerator.h"
 #include "axiom/connectors/ConnectorMetadataRegistry.h"
 #include "axiom/connectors/tests/TestConnector.h"
+#include "axiom/runner/QueryProgress.h"
 #include "axiom/sql/presto/PrestoSqlError.h"
 #include "axiom/sql/presto/tests/ExpectPrestoSqlError.h"
 #include "velox/common/base/VeloxException.h"
@@ -38,6 +39,8 @@ using namespace facebook::velox;
 
 namespace axiom::sql {
 namespace {
+
+namespace runner = facebook::axiom::runner;
 
 class SqlQueryRunnerTest : public ::testing::Test, public test::VectorTestBase {
  protected:
@@ -65,7 +68,8 @@ class SqlQueryRunnerTest : public ::testing::Test, public test::VectorTestBase {
       const std::string& connectorId = "test",
       std::function<std::string()> queryIdGenerator = {},
       PermissionCheck permissionCheck = {}) {
-    auto runner = std::make_unique<SqlQueryRunner>("test_user");
+    auto runner =
+        std::make_unique<SqlQueryRunner>("test_user", &progressScheduler_);
 
     auto initConnectors = [&]() {
       testConnector_ =
@@ -141,6 +145,9 @@ class SqlQueryRunnerTest : public ::testing::Test, public test::VectorTestBase {
         expectedDefault);
   }
 
+  // Drives progress polling for runners built by makeRunner(); declared before
+  // runner_ so it outlives any reporter a query starts.
+  folly::FunctionScheduler progressScheduler_;
   std::unique_ptr<SqlQueryRunner> runner_;
   std::shared_ptr<facebook::axiom::connector::TestConnector> testConnector_;
 
@@ -949,6 +956,7 @@ TEST_F(SqlQueryRunnerTest, completionCallbackReceivesTiming) {
 TEST_F(SqlQueryRunnerTest, startCallbackFiredBeforeCompletion) {
   std::string startQueryId;
   std::string completionQueryId;
+  std::string progressQueryId;
 
   runner_->run(
       "SELECT 1",
@@ -957,6 +965,10 @@ TEST_F(SqlQueryRunnerTest, startCallbackFiredBeforeCompletion) {
              startQueryId = info.queryId;
              EXPECT_EQ(info.query, "SELECT 1");
            },
+       .onProgress =
+           [&](const runner::QueryProgress& info) {
+             progressQueryId = info.queryId;
+           },
        .onComplete =
            [&](const QueryCompletionInfo& info) {
              completionQueryId = info.startInfo.queryId;
@@ -964,6 +976,9 @@ TEST_F(SqlQueryRunnerTest, startCallbackFiredBeforeCompletion) {
 
   EXPECT_FALSE(startQueryId.empty());
   EXPECT_EQ(startQueryId, completionQueryId);
+  // run() forwards onProgress to the runner; the teardown report carries the
+  // id.
+  EXPECT_EQ(progressQueryId, startQueryId);
 }
 
 TEST_F(SqlQueryRunnerTest, callbacksReceiveQueryMetadata) {
--- a/axiom/runner/CMakeLists.txt
+++ b/axiom/runner/CMakeLists.txt
@@ -16,7 +16,14 @@ if(AXIOM_BUILD_TESTING)
   add_subdirectory(tests)
 endif()
 
-add_library(axiom_runner_local_runner LocalRunner.cpp Runner.cpp)
+add_library(
+  axiom_runner_local_runner
+  LocalRunner.cpp
+  ProgressReporter.cpp
+  QueryProgress.cpp
+  QueryProgressBuilder.cpp
+  Runner.cpp
+)
 
 target_link_libraries(
   axiom_runner_local_runner
--- a/axiom/runner/LocalRunner.h
+++ b/axiom/runner/LocalRunner.h
@@ -117,7 +117,7 @@ class LocalRunner : public Runner,
   /// appears before B in the sequence. It's essentially a way to arrange tasks
   /// or items with dependencies so that all prerequisites are completed before
   /// the dependent tasks.
-  const std::vector<optimizer::ExecutableFragment>& fragments() const {
+  const std::vector<optimizer::ExecutableFragment>& fragments() const override {
     return fragments_;
   }
 
--- a/axiom/runner/ProgressReporter.cpp
+++ b/axiom/runner/ProgressReporter.cpp
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "axiom/runner/ProgressReporter.h"
+
+#include <fmt/core.h>
+#include <folly/executors/FunctionScheduler.h>
+#include <glog/logging.h>
+#include <algorithm>
+#include <atomic>
+
+#include "axiom/runner/QueryProgressBuilder.h"
+#include "axiom/runner/Runner.h"
+
+namespace facebook::axiom::runner {
+namespace {
+
+// Process-wide source of unique scheduler function names. Shared schedulers run
+// one function per registered query, so the names must not collide.
+std::string nextName() {
+  static std::atomic<uint64_t> counter{0};
+  return fmt::format("axiom.progress.{}", counter.fetch_add(1));
+}
+
+} // namespace
+
+ProgressReporter::ProgressReporter(
+    Runner& runner,
+    folly::FunctionScheduler& scheduler,
+    std::string queryId,
+    QueryProgressCallback onProgress,
+    int32_t intervalMs)
+    : runner_(runner),
+      scheduler_(scheduler),
+      queryId_(std::move(queryId)),
+      onProgress_(std::move(onProgress)),
+      startTime_(std::chrono::steady_clock::now()) {
+  if (!onProgress_) {
+    return;
+  }
+  stages_ = QueryProgressBuilder::toStageTopology(runner_.fragments());
+  const auto interval = std::chrono::milliseconds(
+      std::max(intervalMs, kMinProgressReportIntervalMs));
+  auto name = nextName();
+  scheduler_.addFunction([this]() { poll(); }, interval, name);
+  // name_ records that a poll is registered, so the destructor knows whether to
+  // cancel one; set it only after addFunction registers the poll.
+  name_ = std::move(name);
+}
+
+ProgressReporter::~ProgressReporter() {
+  // name_ is set only after the poll is registered, so an empty name means
+  // there is nothing to tear down (no callback, or registration never
+  // happened).
+  if (name_.empty()) {
+    return;
+  }
+  // Blocks until the poller is no longer touching this query, so the final
+  // report below is single-threaded and safe even past kRunning.
+  scheduler_.cancelFunctionAndWait(name_);
+  // This destructor may run while the stack unwinds from a query failure, so it
+  // must not throw: swallow any error from building or delivering the final
+  // report (including a builder VELOX_CHECK on a teardown-time stats snapshot).
+  try {
+    const auto wallMicros =
+        std::chrono::duration_cast<std::chrono::microseconds>(
+            std::chrono::steady_clock::now() - startTime_)
+            .count();
+    const bool finished = runner_.state() == Runner::State::kFinished;
+    onProgress_(
+        QueryProgressBuilder::computeProgress(
+            stages_, runner_.stats(), queryId_, wallMicros, finished));
+  } catch (const std::exception& ex) {
+    LOG(WARNING) << "Final progress report failed: " << ex.what();
+  }
+}
+
+void ProgressReporter::poll() {
+  if (runner_.state() != Runner::State::kRunning) {
+    return;
+  }
+  const auto wallMicros = std::chrono::duration_cast<std::chrono::microseconds>(
+                              std::chrono::steady_clock::now() - startTime_)
+                              .count();
+  reportProgress(
+      QueryProgressBuilder::computeProgress(
+          stages_, runner_.stats(), queryId_, wallMicros, /*finished=*/false));
+}
+
+void ProgressReporter::reportProgress(const QueryProgress& progress) {
+  try {
+    onProgress_(progress);
+  } catch (const std::exception& ex) {
+    LOG(WARNING) << "Progress report failed: " << ex.what();
+  }
+}
+
+} // namespace facebook::axiom::runner
--- a/axiom/runner/ProgressReporter.h
+++ b/axiom/runner/ProgressReporter.h
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <chrono>
+#include <string>
+#include <vector>
+
+#include "axiom/runner/QueryProgress.h"
+#include "axiom/runner/QueryProgressBuilder.h"
+
+namespace folly {
+class FunctionScheduler;
+} // namespace folly
+
+namespace facebook::axiom::runner {
+
+class Runner;
+
+/// RAII periodic reporter of one query's progress.
+///
+/// While alive it polls `runner` on the supplied `scheduler` every `intervalMs`
+/// (clamped up to kMinProgressReportIntervalMs) and invokes `onProgress` from a
+/// scheduler thread; the destructor stops polling -- waiting out any in-flight
+/// poll -- then emits one final report from the calling thread, which by then
+/// is single-threaded with respect to this query. The user callback never
+/// throws into the query: an exception from it is logged and skipped.
+///
+/// @code
+///   folly::FunctionScheduler scheduler;
+///   scheduler.start();
+///   {
+///     ProgressReporter reporter(runner, scheduler, queryId, onProgress, 500);
+///     fetchResults(runner); // reporter polls while the query runs
+///   } // ~reporter stops polling and emits the final report
+/// @endcode
+///
+/// Invariants the caller must uphold:
+/// - `scheduler` is started before construction and outlives this object.
+/// - `runner` outlives this object; the final report reads it during teardown.
+/// - `onProgress` must not deregister this query -- that would deadlock the
+/// poll.
+///
+/// When `onProgress` is empty nothing is registered and the scheduler is
+/// untouched.
+class ProgressReporter {
+ public:
+  /// Default target interval between progress reports, in milliseconds.
+  static constexpr int32_t kDefaultProgressReportIntervalMs = 500;
+
+  /// Minimum interval between progress reports, in milliseconds. Caller-set
+  /// intervals are clamped up to this to bound stats() aggregation overhead.
+  static constexpr int32_t kMinProgressReportIntervalMs = 50;
+
+  ProgressReporter(
+      Runner& runner,
+      folly::FunctionScheduler& scheduler,
+      std::string queryId,
+      QueryProgressCallback onProgress,
+      int32_t intervalMs);
+
+  ~ProgressReporter();
+
+  ProgressReporter(const ProgressReporter&) = delete;
+  ProgressReporter& operator=(const ProgressReporter&) = delete;
+  ProgressReporter(ProgressReporter&&) = delete;
+  ProgressReporter& operator=(ProgressReporter&&) = delete;
+
+ private:
+  // Invokes onProgress_ with a caller-built snapshot, logging and swallowing
+  // any exception the user callback throws. Building the snapshot stays with
+  // the caller, so this guards only the callback. poll() builds outside this
+  // guard so a builder invariant violation surfaces during execution; the
+  // destructor builds inside its own catch since it must not throw while a
+  // failing query unwinds.
+  void reportProgress(const QueryProgress& progress);
+
+  // Builds a snapshot and reports it. Does nothing until the runner is running.
+  void poll();
+
+  Runner& runner_;
+  folly::FunctionScheduler& scheduler_;
+  std::string queryId_;
+  QueryProgressCallback onProgress_;
+  std::chrono::steady_clock::time_point startTime_;
+  // Stage topology, derived once from the plan and reused by every poll and the
+  // final report.
+  std::vector<QueryProgressBuilder::StageTopology> stages_;
+  // Scheduler function handle; empty when nothing is registered.
+  std::string name_;
+};
+
+} // namespace facebook::axiom::runner
--- a/axiom/runner/QueryProgress.cpp
+++ b/axiom/runner/QueryProgress.cpp
@@ -14,26 +14,24 @@
  * limitations under the License.
  */
 
-#include "axiom/runner/Runner.h"
+#include "axiom/runner/QueryProgress.h"
 
 #include <folly/container/F14Map.h>
 
 namespace facebook::axiom::runner {
-
 namespace {
-const auto& stateNames() {
-  static const folly::F14FastMap<Runner::State, std::string_view> kNames = {
-      {Runner::State::kInitialized, "initialized"},
-      {Runner::State::kRunning, "running"},
-      {Runner::State::kCancelled, "cancelled"},
-      {Runner::State::kError, "error"},
-      {Runner::State::kFinished, "finished"},
-  };
 
+const auto& executionStateNames() {
+  static const folly::F14FastMap<ExecutionState, std::string_view> kNames = {
+      {ExecutionState::kPlanned, "PLANNED"},
+      {ExecutionState::kRunning, "RUNNING"},
+      {ExecutionState::kFinished, "FINISHED"},
+  };
   return kNames;
 }
+
 } // namespace
 
-AXIOM_DEFINE_EMBEDDED_ENUM_NAME(Runner, State, stateNames);
+AXIOM_DEFINE_ENUM_NAME(ExecutionState, executionStateNames);
 
 } // namespace facebook::axiom::runner
--- a/axiom/runner/QueryProgress.h
+++ b/axiom/runner/QueryProgress.h
@@ -1,0 +1,107 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <functional>
+#include <ostream>
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "axiom/common/Enums.h"
+
+namespace facebook::axiom::runner {
+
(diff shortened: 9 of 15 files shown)


mbasmanova (changes requested):
**Please add an example to the PR description** — what the CLI output looks like
across a longish query's execution. It makes the rest much easier to review and
shows the intended UX.

**1. `poll()` holds the registry lock across `stats()` and the user callback.**
For every registered query, the poll runs `runner.stats()` aggregation and the
`onProgress` callback while holding `state_`'s write lock. That couples
everything: one expensive `stats()` or one slow callback blocks progress for all
other queries and blocks `remove()` — which runs on the teardown path, so it can
delay `run()` returning. It's also what forces the "callback must not re-enter
the dispatcher or it deadlocks" sharp edge. The clean shape is to snapshot under
the lock and run `stats()` + callbacks outside it; keeping the runner alive
without the lock is the real problem to solve, rather than holding the lock
across user code.

**2. `char state` ('P'/'R'/'F') in the public `ExecutionStats`.** This is the
public callback contract — prefer an `enum class` over a magic char. The char
ties the API to Presto's status-column format and isn't self-documenting.

**3. Scan-less queries never reach 'F' / 100%.** Both query and stage state come
from split counts, so `totalSplits == 0` (e.g. the test's `unnest(sequence)`, or
any query without a table scan) leaves `progressPercentage` as nullopt and state
pinned at 'P' forever, including the final teardown report. That's a common CLI
case — worth handling.

**4. The periodic path is effectively untested.** A fast query won't survive a
500ms poll, so the test only exercises the final teardown report (as it notes).
The interval is a compile-time constant, so real periodic reporting can't be
driven deterministically. Making the interval (or the dispatcher) injectable
would let a test force at least one real poll.

**5. Confirm concurrent-read safety.** The design rests on
`LocalRunner::stats()/fragments()/state()` being safe to read from the poll
thread once `kRunning`. The in-function guard for a transient `fragments()` vs
`stats()` size mismatch suggests races are real and known — worth confirming
against the runner's threading contract.

The lazily-started scheduler never stops — it polls an empty map every
500ms for the life of the process. Fine for a CLI, just noting.

singcha:
Thanks @mbasmanova for review!

> Added test plan from diff to summary which shows sample output and screenrecording(Meta only URL) 

> **1. Added snapshot
> **2. Added enum
> **3. Added number of driver and query finish based final status update
> **4. Added test with 1ms poll interval
> **5. It is safe for concurrent-reads - fragments() returns an immutable const member (fixed at construction). stats() reads stages_ and per-task stats under LocalRunner::mutex_; the only mutation after start() — waitForCompletion() clearing stages_ — is also under mutex_, so reads are serialized and always consistent. The state()==kRunning check ensures we don't read before makeStages() populates stages_ during start() . 
Note - ("Consistent" = memory-safe, not a perfect global snapshot — the per-task split counts can still transiently sum to >100%, which is why the percentage is separately clamped to 100%

The PR for adding to CLI is here - https://github.com/facebookincubator/axiom/pull/1499

mbasmanova (changes requested):
The individual notes below cluster into a few recurring themes — probably worth
addressing at that level rather than line by line:

- **The progress data model needs another pass (the main one).** Several fields
  aren't well-defined or sit in the wrong place: `processedRows`/`processedBytes`
  mean different things at query vs stage level and miss scan-less work (points 4, 6);
  split counts are a moving denominator, so the percentage can run backwards (point 7,
  point 12); CPU/wall live on the query wrapper instead of the shared counters (point 9);
  `progressPercentage`'s optional is redundant with `state` (point 10); and the final
  report can come back degenerate (point 11). Worth settling before the API freezes.
- **Presentation belongs in the renderer, not the runner.** `reportProgress`
  computes display depth, ordering, and stage labels, and several fields/comments
  describe how the CLI draws the grid (points 3, 5). Expose data + topology; let the
  CLI (#1499) lay it out.
- **Docs should state the contract, not the implementation or external systems.**
  Comments lean on Presto Java as a reference (point 1) and expose dispatcher/test
  internals (point 13); a reader shouldn't have to reverse-engineer the guarantee.
- **Comments should carry the non-obvious, not restate the code.** Many just
  repeat a field/enum name or its default (points 2, 5), while the genuinely
  non-obvious behavior (dynamic splits, how the percentage is computed) goes
  undocumented (points 7, 10).

Details:

1. **Don't frame the design as matching Presto Java.** Several comments treat
   Presto's Java client/CLI as a reference to conform to:
   - `QueryState`: "Maps to the Presto CLI status column ('P'/'R'/'F') at render
     time"
   - `ExecutionStats`: "Mirrors the fields Presto's StatementStats and StageStats
     have in common"
   - `reportProgress`: "matching Presto's per-stage StageStats" / "matching
     Presto's StatementStats"

   Presto is fine as inspiration, but it shouldn't be the reference — please
   describe these types and choices on Axiom's own terms. Concretely: the
   `'P'/'R'/'F'` glyphs are the renderer's concern (#1499), and the field set
   should be justified by what the runner actually exposes, not by parity with a
   Presto class. (It also avoids asserting Presto-semantics claims we'd otherwise
   need to verify against Presto source.)

2. **Enum-value docs restate the value name.** "Finished: …", "Running: …",
   "Planned: …" just repeat the enum name before the explanation — drop the
   prefix and keep the description. And for `kPlanned`, drop "queued": the local
   runner doesn't queue, so it names a concept that doesn't exist here (a real
   queued state would be its own value). So:
   - `kPlanned` → "No split or driver has started yet."
   - `kRunning` → "At least one split or driver is executing."
   - `kFinished` → "All splits and drivers have completed."

3. **`reportProgress` does rendering work in the runner.** It computes a per-stage
   display `depth` (a BFS over the fragment tree), sorts stages by depth, and
   re-labels `stageId` sequentially so the output can be indented. That's a
   renderer's job, not the runner's — it couples the progress API to one specific
   tree layout and puts plan-topology logic on the hot poll path. The runner
   should expose raw per-stage stats plus the topology a renderer needs (e.g. each
   stage's producers / fragment indices), and let the CLI (#1499) decide depth,
   order, and indentation. (The `stageId` assigned when the stage is built is also
   immediately overwritten by the relabel — dead store, a symptom of the same
   layering issue.)

4. **Query-total `processedRows`/`processedBytes` count `TableScan` only.** As a
   result a scan-less query (e.g. `SELECT … FROM unnest(...)`) reports 0 processed
   rows/bytes for its whole run, even while it's clearly doing work. Is that
   intended (connector-input semantics), or should the query total reflect
   something a user watching the progress bar would see advancing?

5. **Field comments mostly restate the code (or the rendering).** In
   `ExecutionStats`/`QueryProgressInfo` several comments add nothing the name or
   default already says: `state` ("Defaults to kPlanned …" — restates the
   initializer), `totalSplits`/`queuedSplits`/`runningSplits`/`completedSplits`
   ("Total number of splits.", …), `queryId` ("Unique identifier for this
   query execution."). Keep only what the name doesn't convey — e.g.
   `processedRows` = "raw input from source operators", `wallTimeMicros` = "since
   query start", `progressPercentage`'s range/nullopt meaning all earn their
   comment. Separately, `stageId` ("rendered as the leftmost column"), `depth`
   ("used for indentation"), and `stages` ("rendered as a flat grid below the
   summary line") describe how the CLI draws it — same layering issue as point 3; the
   API shouldn't document the renderer.

6. **`processedRows`/`processedBytes` are under-defined and mean two different
   things.** The doc ("raw input rows read from source operators") has two
   problems. First, "raw" is unexplained jargon — it's Velox's
   `rawInputPositions` (rows the source read before processing, vs output rows),
   which a reader of this header has no way to know. Second, it doesn't match the
   code: the query-total counts `TableScan` only, while the per-stage value sums
   every operator's raw input (scan at leaves, exchange above) — so the *same
   field* carries different semantics in its two uses (and `StageProgress`'s
   comment gives a third wording). Values/unnest are reflected by neither (not a
   `TableScan`, and they generate rather than read, so `rawInputPositions` ≈ 0) —
   the same root cause as point 4. Please define the field precisely and reconcile the
   two meanings. The name doesn't help either: `processed` reads like *output*
   the query has finished, not source input — `sourceRows`/`sourceBytes` (or
   `inputRows`/`inputBytes`) would say what it is.

7. **Document that the split counts are dynamic.** `totalSplits` isn't a fixed
   denominator — the runner adds splits during execution (not all up front), so it
   grows over time, and the substates don't cleanly sum to it
   (`totalSplits >= queuedSplits + runningSplits + completedSplits`, also because
   the counters are aggregated across concurrently-updating tasks). That's the
   non-obvious fact these comments should carry (vs. the restatements in point 5). It
   also means split-based `progressPercentage` can move *backwards* as
   `totalSplits` grows — worth noting, since a progress bar is usually assumed
   monotonic; the current code only guards the overshoot (clamps to 100), not
   regression.

8. **Name the query- and stage-level views as a pair.** `QueryProgressInfo`
   (query level) and `StageProgress` (stage level) are the same thing at two
   scopes but don't read as parallel. "Info" makes sense for the *event* payloads
   (`QueryStartInfo`, `QueryCompletionInfo`) — "Start"/"Completion" are events, so
   "Info" turns them into data — but "Progress" is already a noun, so the suffix is
   redundant there. So `QueryProgress` (drop Info) parallel to `StageProgress`,
   leaving the event types as-is. Also pick one vocabulary — the shared counters
   are `ExecutionStats` ("stats") while the wrappers say "Progress".

9. **CPU/wall time belong in `ExecutionStats`.** `cpuTimeMicros` and
   `wallTimeMicros` sit only on `QueryProgressInfo`, but they're execution
   counters like the rest, so they should live in the shared `ExecutionStats` and
   be reported per stage too (Presto's StageStats carries per-stage CPU/wall), with
   the query value as the aggregate. `reportProgress` already computes per-operator
   CPU in the stage loop and just sums it into the query total, so exposing it per
   stage is nearly free. (`wallTime` is more arguable — query "since start" vs
   per-stage elapsed; if per-stage wall isn't well-defined, keep wall at the query
   level but still move CPU into `ExecutionStats`.)

10. **`progressPercentage`: reconsider the `optional` (and maybe the name).** It's
    `nullopt` only when nothing has started yet (not finished, no splits, no
    drivers) — which `state == kPlanned` already tells you. So the optional is
    redundant with `state`; a plain `double` defaulting to `0.0` is simpler, and a
    renderer wanting an indeterminate cue can read `state`. On the name, `progress`
    is cleaner but loses the scale — if you rename, pin 0–100 vs 0–1 in the doc (or
    make it a 0.0–1.0 fraction). And do describe *how* it's computed — that's the
    non-obvious part worth a comment: `finishedSplits/totalSplits`, falling back to
    `completedDrivers/totalDrivers` when there are no splits, forced to 100 on
    completion, clamped to [0,100], and not necessarily monotonic (see point 7).

11. **When is `stages` empty — and does the final report go degenerate?** The
    comment ("no stage-level detail available") is vague. From the code `stages`
    is empty when `runner.stats()` returns nothing; polls gate on `kRunning`, so
    the real case is the final teardown report — `waitForCompletion()` clears
    `stages_`, and the final report runs after that. If so, the final callback
    reports `kFinished`/100% but with empty `stages` and zeroed counters
    (rows/bytes/cpu), so the totals are lost — it only looks fine because the CLI
    clears the live display and prints results. Is that intended? If the final
    frame should carry totals, snapshot the last good stats before they're cleared.
    And make the comment state the concrete condition.

12. **A non-monotonic "progress %" is questionable — consider an ETA.** Because
    `totalSplits` grows during execution (point 7), `finishedSplits/totalSplits` can
    move *backwards* — and a "percentage complete" that regresses is a
    contradiction in terms and a poor experience. Reconsider the model: when the
    denominator isn't fixed, an ETA / remaining-work estimate degrades more
    gracefully — people accept a revised ETA, but not a progress bar going down.
    Worth deciding whether "progress %" is the right abstraction here at all.

13. **`progressReportIntervalMs`: document the contract, not the implementation.**
    The doc describes how it works (a shared dispatcher ticking at the smallest
    interval across queries) and how tests use it — a reader shouldn't have to
    reverse-engineer the guarantee from internals. State the contract: the callback
    is invoked approximately every N ms while the query runs (best-effort, not a
    hard guarantee), plus once when it finishes. Drop the dispatcher mechanics and
    the test/production framing. (Same principle as points 1/5 — describe the API's
    guarantee, not what's behind it.)

14. **No floor on the poll interval — the "bound overhead" claim isn't enforced.**
    `kProgressReportIntervalMs` (500) is the default, not a max frequency: the
    dispatcher ticks at the *smallest* interval any query requests, and
    `progressReportIntervalMs` is caller-set with no minimum. So the comment's
    rationale ("kept coarse to bound overhead on large plans") doesn't hold — a
    caller (or a test value reaching prod) can set 1ms and drive `stats()`
    aggregation every millisecond, the exact overhead it claims to avoid. Either
    clamp to a sane minimum (a real floor) or drop the overhead claim.

15. **The standalone default is redundant — and only idle-ticks a never-stopping
    scheduler.** Every query carries its own interval (RunOptions defaults to
    `500`), and the cadence is the min across registered queries.
    `kProgressReportIntervalMs` is used only as the fallback when `queries_` is
    empty — i.e., to keep the leaked, always-running scheduler ticking on an empty
    registry. So (a) it's a second, unlinked `500` (RunOptions uses a literal
    `{500}`, not the constant — make it one source of truth), and (b) it only
    matters because the scheduler keeps polling with nothing registered. Parking
    the scheduler when the registry goes empty removes the need for an idle default
    altogether.

singcha:
Thanks @mbasmanova! Let me incorporate review comments. 
qq on links - Those are referring to unrelated PRs, did you mean to add link to code?

singcha:
Thanks @mbasmanova ! 
Updated the PR to address review feedback: rework the progress data model (QueryProgress; inputRows/inputBytes; CPU in ExecutionStats; drop percentage/ETA in favor of raw counts + state), move stage depth/order/indent into the renderer, swap the dispatcher to folly::FunctionScheduler (parks when idle, interval floor), carry totals in the final report. Renames QueryProgress end-to-end through other layers

mbasmanova (changes requested):
1. **Pull progress reporting into its own `.h`/`.cpp`.** It's a self-contained
   subsystem, not a few helpers: `computeProgress` (~150 lines), a process-wide
   threaded `ProgressDispatcher` with its own singleton and lifecycle (~70), and
   `ProgressRegistration` (~55), plus ~80 lines of public types in the header. It
   has its own threading model and little to do with `SqlQueryRunner`'s
   parse/plan/execute job. Give it `axiom/cli/QueryProgress.{h,cpp}` (or
   `ProgressReporter`).

2. **Make `computeProgress` a static method on that class, and design the tests
   around it.** Today the logic has no seam, so the tests can only run real
   queries and watch callbacks — which is why they assert almost nothing
   (existence / non-negativity) and need a 64M-row query just to force one
   periodic poll. A pure static `computeProgress(taskStats, fragments, wallMicros,
   lastGood)` lets a unit test feed synthetic stats/fragments and assert the
   *exact* `QueryProgress`: per-stage state and counters, `producerStageIds` edges
   (including diamond plans and the size-mismatch fallback), query-level sums,
   each state transition, the `lastGood` reuse, scan-vs-shuffle. Then
   `SqlQueryRunnerTest` keeps just a thin "callback is wired and fires" check — no
   giant query. The current tests read as coverage, not designed verification;
   this is the fix.

3. **`QueryState` → `ExecutionState`.** The state lives in `ExecutionStats` and
   describes a query *or* a stage, so `QueryState` is a misnomer. `ExecutionState`
   matches `ExecutionStats` and reads correctly at both levels.

4. **Use the Axiom enum-name convention, not a bare `enum class`.** Declare it
   with `AXIOM_DECLARE_ENUM_NAME(ExecutionState)` in the header and
   `AXIOM_DEFINE_ENUM_NAME(ExecutionState, <names>)` in the `.cpp` (see
   `axiom/common/Enums.h`). That gives the `toName` / `to…` / `tryTo…` +
   `operator<<` the rest of Axiom's enums use.

5. **Report storage reads separately from shuffle.** `inputRows`/`inputBytes`
   currently lump connector reads and exchange input together, which double-counts
   and forces the "throughput signal, not deduplicated" caveat. Splitting them —
   e.g. `scanRows`/`scanBytes` (read from storage = `TableScan`) vs
   `shuffleRows`/`shuffleBytes` (exchange input) — gives two signals a reader can
   actually interpret: how much source data has been read, and how much has moved
   across stages. The data is already there; it's just split by operator type in
   the loop. This also removes the "near zero" ambiguity in the doc — storage
   input is *exactly* 0 for a scan-less query, and shuffle is reported on its own.

6. **Static topology is recomputed on every poll.** `indexByPrefix` and each
   stage's producer edges derive only from `fragments()`, which is immutable for
   the query — but `computeProgress` rebuilds them on every report; only the
   per-stage counters change between polls. Build the stage skeleton (index +
   producer edges) once and refresh only the stats each tick. Avoids O(fragments)
   repeated work on the hot path — the overhead the interval floor exists to
   bound. It also lets the producer→index resolution be done once and correctly:
   today `stageIndex` is the stats index while producers come from the fragment
   index, which coincide only under the unstated assumption that `fragments()` and
   `stats()` are 1:1 in the same order (the guard only checks size, not ordering).

7. **`QueryProgress::stats` comment singles out the wrong field.** "cpuTimeMicros
   here is the sum across stages" — but *every* numeric counter is summed across
   stages (`inputRows`, `inputBytes`, splits, cpu). The only field that isn't a
   plain sum is `state`, which is derived from the aggregate counts. Reword (or
   drop the cpu-specific note).

8. **Struct doc comments aren't consistent.** The related types are each written
   in a different voice — `StageProgress` ("Per-stage execution detail within a
   query.") vs `QueryProgress` ("Snapshot of a running query's progress: …") vs
   `ExecutionStats` ("Execution counters shared by …, so one type backs …"). Pick
   one style — a concise "what it is" line — and apply it uniformly across
   `ExecutionState`/`ExecutionStats`/`StageProgress`/`QueryProgress`.

9. **`stageId` → `stageIndex`.** It's a position in `QueryProgress::stages`, not
   an opaque id — `stageIndex` says that, with comment
   `/// Stage index in QueryProgress::stages list`. For consistency,
   `producerStageIds` (which holds these indexes) should follow →
   `producerStageIndexes`.

10. **`producerStageIds` comment names a specific consumer.** "A renderer rebuilds
    the stage tree from these edges to choose depth, order, and indentation"
    describes who uses it and how — the doc shouldn't name a specific consumer.
    Describe the data instead: the stages form a tree, and these are the producer
    edges from which it can be reconstructed; empty for a connector-only leaf.

11. (nit) **`QueryProgressCallback` doc** says "must not block the query" but
    should also warn that deregistering this query from within its own callback
    would deadlock `cancelFunctionAndWait`.

singcha:
Thanks @mbasmanova! 
Incorporated the review feedback. Can you please review again.

mbasmanova (changes requested):
Most of this is one story — a **reusable runner component** placed, named, and
built as if it were private CLI glue. The main asks:

- **Move it to `axiom/runner` and depend on the `Runner` interface**, not concrete
  `LocalRunner`.
- **Don't install a hidden, leaked, process-wide scheduler thread** — have the
  caller supply the scheduler.
- **Rename to match roles:** `ProgressRegistration` → `ProgressReporter` (it's
  what reports); today's `ProgressReporter` → `QueryProgressBuilder`; and tidy
  redundant/misleading field/method/param names.
- **Fix scan/shuffle classification** — use the `OperatorType::k…` constants and
  assert every `rawInput`-bearing operator is recognized; today it silently
  undercounts.
- **Make the docs state the contract and match the code** — `state` derivability,
  the stats↔stages model, and the final-report guarantee are wrong or unstated.

Plus a doc-grouping cleanup, the double topology build, and two nits. Details:

1. **`ProgressReporter` has nothing SQL-specific — move it next to the Runner.**
   It depends only on `velox::exec::TaskStats`, `optimizer::ExecutableFragment`,
   and `runner::LocalRunner`; it's a progress view over runner state
   (`state()`/`stats()`/`fragments()`). It's in `namespace axiom::sql` /
   `axiom/cli/` only because it was carved out of `SqlQueryRunner`. It belongs in
   `axiom/runner/` under `facebook::axiom::runner`: layering stays clean
   (`LocalRunner` already depends on both optimizer and velox stats), the
   dependency points runner→its-own-progress instead of the CLI owning a runner
   utility, and any runner front-end can reuse it. `SqlQueryRunner::RunOptions`
   just references `runner::QueryProgress`/`QueryProgressCallback`.

   Relatedly, it should depend on the `Runner` interface, not concrete
   `LocalRunner` (e.g. `ProgressRegistration`'s ctor takes `LocalRunner&`).
   `state()` and `stats()` are already pure virtuals on `Runner`; only
   `fragments()` is `LocalRunner`-specific. Lift `fragments()` onto `Runner` (every
   runner executes a fragmented plan) and take `runner::Runner&` in
   `ProgressRegistration` (and the dispatcher), so progress reporting works for any
   runner — not just `LocalRunner`.

2. **`ExecutionStats`: document fields in groups, consistently.** The four
   `*Splits` share one grouped comment, but `scanRows`/`scanBytes` and
   `shuffleRows`/`shuffleBytes` each get a separate line that repeats "Rows…/
   Bytes…". Group each pair like splits — one comment over the scan pair and one
   over the shuffle pair — which drops the repetition and reads more uniformly,
   e.g.:

   ```cpp
   /// Rows/bytes read from storage by table scans.
   int64_t scanRows{0};
   int64_t scanBytes{0};
   /// Rows/bytes received from exchanges (shuffled in from producer stages).
   int64_t shuffleRows{0};
   int64_t shuffleBytes{0};
   ```

3. **Scan/shuffle classification silently drops unrecognized source operators.**
   `computeProgress` counts scan only for `"TableScan"` and shuffle only for
   `"Exchange"`/`"MergeExchange"`; any *other* operator that reports `rawInput`
   falls out of both buckets and the totals undercount with no signal. Two fixes:
   - Use the named constants — `velox::exec::OperatorType::kTableScan` /
     `kExchange` / `kMergeExchange` (`velox/exec/OperatorType.h`, which exists "to
     avoid bare string literals at ... comparison sites") — so a rename can't
     desync the comparison.
   - Add a check that every operator reporting `rawInput`
     (`rawInputPositions`/`rawInputBytes > 0`) is recognized by one of the
     buckets; assert if not. Then a renamed or newly added source type fails
     loudly instead of quietly undercounting.

4. **`QueryProgress::stats` doc: `state` isn't derivable from the exposed
   counts.** The comment says "state is derived from the aggregate counts," but
   `state` is computed from the runner's `finished` flag plus split *and driver*
   completion — and `ExecutionStats` exposes only split counts, not drivers, and
   not the `finished` flag. So a consumer holding a `QueryProgress` can't
   reconstruct `state` from the counters. Reword: `state` reflects overall
   completion and should be read directly, not recomputed.

5. **The stats↔stages consistency model is under-specified, and the comments
   hand-wave it.** Two spots:
   - `stages` doc: "empty on a snapshot taken before any task has stats" reads as
     if `stages` fills in incrementally as tasks start — but `computeProgress`
     treats it as empty-or-complete (the `lastGood` fallback keys on
     `stages.empty()`).
   - `computeProgress` doc: "on a size mismatch (a transient snapshot) the report
     falls back to flat per-stage stats" — but it never says what a "transient
     snapshot" is or what causes a mismatch.

   Given `LocalRunner::stats()` is documented "Corresponds 1:1 to fragments()" and
   `fragments()` is immutable, it's unclear a *non-empty* size mismatch can occur
   at all (the empty case is already handled by `lastGood`). So pin the model
   down: if stats and fragments can't diverge (beyond empty), make the
   size-mismatch branch an explicit invariant (`VELOX_CHECK`) and drop the
   speculative wording; if they can, state the concrete cause and confirm the
   partial-sum case is handled.

6. **`buildStages` runs twice per query.** `ProgressRegistration` calls it in the
   ctor (passed to the dispatcher) and again in the dtor for the final report.
   Cache the result on the registration and reuse it in the dtor.

7. **The `ProgressReporter` docs document implementation, not contract.**
   The class doc's second paragraph walks through internals an API user doesn't need: that the
   topology is "built once with buildStages()", that each poll refreshes only the
   counters, and that the methods are "pure and static so they can be unit
   tested." Trim it to what the class provides; that mechanism belongs in the
   `.cpp` (or nowhere public). And "built once" isn't merely inaccurate (the dtor
   rebuilds — see the double-build point) — "once" is misplaced: it's no more a property of `buildStages`
   than "`plus(a, b)` computes the sum once" is. The function computes the
   topology from the plan; how often it's called is the caller's concern. ("skeleton"
   also clashes with the type name `StageTopology` — use one term.)

   The `buildStages` method doc has the same problems — "Resolves … to stage
   indexes once" (the misplaced "once" again) and "the per-poll path reuses this
   rather than rebuilding it each tick" (caller behavior, not the method's
   contract). Keep the one useful part — that stage `i` ↔ `fragments[i]` ↔
   `stats[i]` — and drop the rest.

8. **Redundant "stage" in `Stage*` field names** (public API). Inside a `Stage*`
   struct, `stageIndex`/`producerStageIndexes` repeat "stage" — `index` and
   `producers` read fine (the producers are stage indexes; the doc says so). Apply
   to both `StageTopology` and `StageProgress`.

9. **`buildStages` → `toStageTopology`** (public API). It's a pure conversion
   (`fragments → StageTopology`), so a `to…` name fits better than "build" and
   lands the one-term ("topology") recommendation above (returns a vector, so
   `toStageTopology` / `toStageTopologies`).

10. **`lastGood` → `latestProgress`** (public API). `lastGood` is vague — "good"
    how? It's the most recent non-empty progress, used as the fallback. Something
    like `latestProgress` (or `lastReported`) reads clearer; rename the
    `computeProgress` parameter (and its doc) and the dtor local. (The `cache_`
    member that feeds it is fine as-is.)

11. **Final-report consistency is conditional and under-documented.** On a clean
    finish the code forces `completedSplits = totalSplits`, `queued = running = 0`
    (adds up). On a non-clean teardown (failed/cancelled, `finished == false`) it
    returns the last mid-flight `lastGood` unchanged — counts that are partial and,
    summed across concurrently-updating tasks, may not satisfy
    `totalSplits == queued + running + completed`. The doc's "marked done on a
    clean finish" only hints at this. State the contract: clean finish → totals +
    finished; otherwise → last observed, possibly partial/inconsistent.

12. **Class names don't match roles** (public API). The class that actually
    reports is `ProgressRegistration` — it registers the periodic poll and emits
    the final report — so it should be `ProgressReporter`. The current
    `ProgressReporter` only computes (`buildStages`/`computeProgress` +
    `StageTopology`) — rename it `QueryProgressBuilder`. The interval constants
    (`kMin/kDefault…IntervalMs`) are about reporting cadence, so they belong on the
    reporter, not the builder.

13. **The progress scheduler is a hidden, leaked, process-wide thread — inject it
    instead.** Reporting silently installs a process-wide `FunctionScheduler`
    singleton, started on first use and never stopped (leaked), that callers of
    this reusable runner can't control, scope, or run on their own executor. Have
    the runner/registration take the scheduler (something that supports periodic
    execution) as a supplied dependency. Document the threading contract too
    (which thread the callback runs on, the cadence, the final report on teardown)
    — none of it is documented now.

14. (nit) **`cache_` type is spelled out raw in the header**
    (`std::shared_ptr<folly::Synchronized<std::optional<QueryProgress>>>`) while the
    `.cpp` has a `ProgressCache` alias — share one alias.

15. (nit) **Drop "snapshot" from `QueryProgress`'s doc.** It says "Snapshot of …"
    while `ExecutionStats`/`StageProgress` don't; all three are point-in-time by
    nature, so the word adds nothing.

singcha:
> Most of this is one story — a **reusable runner component** placed, named, and built as if it were private CLI glue. The main asks:
> 
> * **Move it to `axiom/runner` and depend on the `Runner` interface**, not concrete
>   `LocalRunner`.
> * **Don't install a hidden, leaked, process-wide scheduler thread** — have the
>   caller supply the scheduler.
> * **Rename to match roles:** `ProgressRegistration` → `ProgressReporter` (it's
>   what reports); today's `ProgressReporter` → `QueryProgressBuilder`; and tidy
>   redundant/misleading field/method/param names.
> * **Fix scan/shuffle classification** — use the `OperatorType::k…` constants and
>   assert every `rawInput`-bearing operator is recognized; today it silently
>   undercounts.
> * **Make the docs state the contract and match the code** — `state` derivability,
>   the stats↔stages model, and the final-report guarantee are wrong or unstated.
> 
> Plus a doc-grouping cleanup, the double topology build, and two nits. Details:
> 
> 1. **`ProgressReporter` has nothing SQL-specific — move it next to the Runner.**
>    It depends only on `velox::exec::TaskStats`, `optimizer::ExecutableFragment`,
>    and `runner::LocalRunner`; it's a progress view over runner state
>    (`state()`/`stats()`/`fragments()`). It's in `namespace axiom::sql` /
>    `axiom/cli/` only because it was carved out of `SqlQueryRunner`. It belongs in
>    `axiom/runner/` under `facebook::axiom::runner`: layering stays clean
>    (`LocalRunner` already depends on both optimizer and velox stats), the
>    dependency points runner→its-own-progress instead of the CLI owning a runner
>    utility, and any runner front-end can reuse it. `SqlQueryRunner::RunOptions`
>    just references `runner::QueryProgress`/`QueryProgressCallback`.
>    Relatedly, it should depend on the `Runner` interface, not concrete
>    `LocalRunner` (e.g. `ProgressRegistration`'s ctor takes `LocalRunner&`).
>    `state()` and `stats()` are already pure virtuals on `Runner`; only
>    `fragments()` is `LocalRunner`-specific. Lift `fragments()` onto `Runner` (every
>    runner executes a fragmented plan) and take `runner::Runner&` in
>    `ProgressRegistration` (and the dispatcher), so progress reporting works for any
>    runner — not just `LocalRunner`.
> 2. **`ExecutionStats`: document fields in groups, consistently.** The four
>    `*Splits` share one grouped comment, but `scanRows`/`scanBytes` and
>    `shuffleRows`/`shuffleBytes` each get a separate line that repeats "Rows…/
>    Bytes…". Group each pair like splits — one comment over the scan pair and one
>    over the shuffle pair — which drops the repetition and reads more uniformly,
>    e.g.:
>    ```c++
>    /// Rows/bytes read from storage by table scans.
>    int64_t scanRows{0};
>    int64_t scanBytes{0};
>    /// Rows/bytes received from exchanges (shuffled in from producer stages).
>    int64_t shuffleRows{0};
>    int64_t shuffleBytes{0};
>    ```
> 3. **Scan/shuffle classification silently drops unrecognized source operators.**
>    `computeProgress` counts scan only for `"TableScan"` and shuffle only for
>    `"Exchange"`/`"MergeExchange"`; any _other_ operator that reports `rawInput`
>    falls out of both buckets and the totals undercount with no signal. Two fixes:
>    
>    * Use the named constants — `velox::exec::OperatorType::kTableScan` /
>      `kExchange` / `kMergeExchange` (`velox/exec/OperatorType.h`, which exists "to
>      avoid bare string literals at ... comparison sites") — so a rename can't
>      desync the comparison.
>    * Add a check that every operator reporting `rawInput`
>      (`rawInputPositions`/`rawInputBytes > 0`) is recognized by one of the
>      buckets; assert if not. Then a renamed or newly added source type fails
>      loudly instead of quietly undercounting.
> 4. **`QueryProgress::stats` doc: `state` isn't derivable from the exposed
>    counts.** The comment says "state is derived from the aggregate counts," but
>    `state` is computed from the runner's `finished` flag plus split _and driver_
>    completion — and `ExecutionStats` exposes only split counts, not drivers, and
>    not the `finished` flag. So a consumer holding a `QueryProgress` can't
>    reconstruct `state` from the counters. Reword: `state` reflects overall
>    completion and should be read directly, not recomputed.
> 5. **The stats↔stages consistency model is under-specified, and the comments
>    hand-wave it.** Two spots:
>    
>    * `stages` doc: "empty on a snapshot taken before any task has stats" reads as
>      if `stages` fills in incrementally as tasks start — but `computeProgress`
>      treats it as empty-or-complete (the `lastGood` fallback keys on
>      `stages.empty()`).
>    * `computeProgress` doc: "on a size mismatch (a transient snapshot) the report
>      falls back to flat per-stage stats" — but it never says what a "transient
>      snapshot" is or what causes a mismatch.
>    
>    Given `LocalRunner::stats()` is documented "Corresponds 1:1 to fragments()" and
>    `fragments()` is immutable, it's unclear a _non-empty_ size mismatch can occur
>    at all (the empty case is already handled by `lastGood`). So pin the model
>    down: if stats and fragments can't diverge (beyond empty), make the
>    size-mismatch branch an explicit invariant (`VELOX_CHECK`) and drop the
>    speculative wording; if they can, state the concrete cause and confirm the
>    partial-sum case is handled.
> 6. **`buildStages` runs twice per query.** `ProgressRegistration` calls it in the
>    ctor (passed to the dispatcher) and again in the dtor for the final report.
>    Cache the result on the registration and reuse it in the dtor.
> 7. **The `ProgressReporter` docs document implementation, not contract.**
>    The class doc's second paragraph walks through internals an API user doesn't need: that the
>    topology is "built once with buildStages()", that each poll refreshes only the
>    counters, and that the methods are "pure and static so they can be unit
>    tested." Trim it to what the class provides; that mechanism belongs in the
>    `.cpp` (or nowhere public). And "built once" isn't merely inaccurate (the dtor
>    rebuilds — see the double-build point) — "once" is misplaced: it's no more a property of `buildStages`
>    than "`plus(a, b)` computes the sum once" is. The function computes the
>    topology from the plan; how often it's called is the caller's concern. ("skeleton"
>    also clashes with the type name `StageTopology` — use one term.)
>    The `buildStages` method doc has the same problems — "Resolves … to stage
>    indexes once" (the misplaced "once" again) and "the per-poll path reuses this
>    rather than rebuilding it each tick" (caller behavior, not the method's
>    contract). Keep the one useful part — that stage `i` ↔ `fragments[i]` ↔
>    `stats[i]` — and drop the rest.
> 8. **Redundant "stage" in `Stage*` field names** (public API). Inside a `Stage*`
>    struct, `stageIndex`/`producerStageIndexes` repeat "stage" — `index` and
>    `producers` read fine (the producers are stage indexes; the doc says so). Apply
>    to both `StageTopology` and `StageProgress`.
> 9. **`buildStages` → `toStageTopology`** (public API). It's a pure conversion
>    (`fragments → StageTopology`), so a `to…` name fits better than "build" and
>    lands the one-term ("topology") recommendation above (returns a vector, so
>    `toStageTopology` / `toStageTopologies`).
> 10. **`lastGood` → `latestProgress`** (public API). `lastGood` is vague — "good"
>     how? It's the most recent non-empty progress, used as the fallback. Something
>     like `latestProgress` (or `lastReported`) reads clearer; rename the
>     `computeProgress` parameter (and its doc) and the dtor local. (The `cache_`
>     member that feeds it is fine as-is.)
> 11. **Final-report consistency is conditional and under-documented.** On a clean
>     finish the code forces `completedSplits = totalSplits`, `queued = running = 0`
>     (adds up). On a non-clean teardown (failed/cancelled, `finished == false`) it
>     returns the last mid-flight `lastGood` unchanged — counts that are partial and,
>     summed across concurrently-updating tasks, may not satisfy
>     `totalSplits == queued + running + completed`. The doc's "marked done on a
>     clean finish" only hints at this. State the contract: clean finish → totals +
>     finished; otherwise → last observed, possibly partial/inconsistent.
> 12. **Class names don't match roles** (public API). The class that actually
>     reports is `ProgressRegistration` — it registers the periodic poll and emits
>     the final report — so it should be `ProgressReporter`. The current
>     `ProgressReporter` only computes (`buildStages`/`computeProgress` +
>     `StageTopology`) — rename it `QueryProgressBuilder`. The interval constants
>     (`kMin/kDefault…IntervalMs`) are about reporting cadence, so they belong on the
>     reporter, not the builder.
> 13. **The progress scheduler is a hidden, leaked, process-wide thread — inject it
>     instead.** Reporting silently installs a process-wide `FunctionScheduler`
>     singleton, started on first use and never stopped (leaked), that callers of
>     this reusable runner can't control, scope, or run on their own executor. Have
>     the runner/registration take the scheduler (something that supports periodic
>     execution) as a supplied dependency. Document the threading contract too
>     (which thread the callback runs on, the cadence, the final report on teardown)
>     — none of it is documented now.
> 14. (nit) **`cache_` type is spelled out raw in the header**
>     (`std::shared_ptr<folly::Synchronized<std::optional<QueryProgress>>>`) while the
>     `.cpp` has a `ProgressCache` alias — share one alias.
> 15. (nit) **Drop "snapshot" from `QueryProgress`'s doc.** It says "Snapshot of …"
>     while `ExecutionStats`/`StageProgress` don't; all three are point-in-time by
>     nature, so the word adds nothing.

Thanks @mbasmanova ! Incorporated the review comments.

mbasmanova (changes requested):
Thank you for iterating. This is super close.

1. **`Runner::fragments()` should be pure virtual**, not defaulted to empty. A
   default-empty implementation silently degrades progress (no stage tree) for any
   runner that forgets to override it; an interface method should force each
   `Runner` to implement it explicitly — a runner with no topology returns `{}` on
   purpose.

2. **The unclassified-operator check logs — but a log isn't useful here.** When an
   operator reports `rawInput` but isn't scan/exchange, `QueryProgressBuilder`
   does `LOG_FIRST_N(WARNING, 10)`. A warning buried in worker logs won't be
   noticed, so the drift it's meant to catch — a new source type silently dropped
   from the totals — still slips through. We asked for an assert earlier and still
   think that's right: use `VELOX_CHECK`.

3. **`SqlQueryRunner` owns the scheduler — same hidden-thread issue, one layer
   up.** Injecting the scheduler into `ProgressReporter` is right, but
   `SqlQueryRunner` now holds and `start()`s a `FunctionScheduler progressScheduler_`
   member itself, so a reusable runner's callers still can't control or supply it.
   Make it an optional constructor arg: if supplied, use it; if not, don't report
   progress — and a request that sets `onProgress` fails loudly (it asked for
   progress without supplying a scheduler), rather than `SqlQueryRunner` silently
   owning one.

4. **Drop the redundant test helpers `topology()`, `SplitCounts`, `DriverCounts`.**
   `StageTopology` and `TaskStats` are plain aggregates — designated-init them
   directly (`StageTopology{.index = …, .producers = {…}}`;
   `TaskStats{.numTotalSplits = …, .numTotalDrivers = …, …}`). `SplitCounts`/
   `DriverCounts` just mirror fields `TaskStats` already has, and `topology()`
   wraps a two-field struct — both are indirection. (A helper to build a simple
   aggregate would also hint the prod type is awkward; it isn't.)

5. **The `fragment()` helper builds invalid topology.** It gives every input the
   same `consumerNodeId = "exchange"` when a fragment has multiple producers, but a
   single node can't read from multiple producers — each producer is a distinct
   exchange node. Give each input its own `consumerNodeId` so the test models a
   real plan.

6. **Drop `TEST(QueryProgressBuilderTest, executionStateNames)`.** It only checks
   the macro-generated name strings for `ExecutionState` — not worth a test, and
   `ExecutionState` belongs to `QueryProgress`, not the builder.

7. **`toStageTopology` papers over malformed input with silent fallbacks — make
   them loud failures.** Three tests encode behavior that shouldn't exist:
   - `SkipsUnknownProducer`: an unresolved `producerTaskPrefix` is a malformed
     plan; `VELOX_CHECK` instead of silently skipping the edge.
   - `sizeMismatchDropsProducerEdges`: a stats/topology size mismatch is a bug, not
     something to paper over by dropping edges; fail loudly.
   - `toStageTopologyDiamond`: a diamond topology isn't valid in Velox — drop the
     test, and don't encode support for a shape that can't occur.
   No silent fallbacks when the inputs don't make sense.

singcha:
Thanks @mbasmanova for reviews! Updated to incorporate latest comments

mbasmanova (changes requested):
1. **Drop the "runner with no plan topology returns an empty vector" allowance in
   `Runner::fragments()`.** A runner always has a plan; there's no reason for one
   not to expose its topology. Once that's gone, the "empty / topology
   unavailable" branches it spawned downstream should go too — the `hasTopology`
   bypass and the "stages is either empty …" wording in `computeProgress`, and
   the "when plan topology is unavailable" caveat on `StageProgress::producers`.

2. **The documented teardown fallback can't fire when topology is present.**
   `computeProgress`'s doc says an empty `taskStats` falls back to
   `latestProgress` "so the final report still carries totals" — but with a
   non-empty `stages`, `VELOX_CHECK_EQ(stages.size(), taskStats.size())` throws
   first. `~ProgressReporter` always passes the real `stages_`, so when the runner
   has dropped its stats (exactly the case the `cache_` / `latestProgress`
   machinery exists for) it throws, gets caught, logs "Final progress report
   failed", and emits no final report — totals lost. When stats are present the
   fallback branch is skipped. The only path that reaches it is the unit test
   passing `stages={}`, which the reporter never does. Either let the teardown
   pass empty `stages` (or relax the check) when stats are gone so the fallback
   actually runs, or remove the dead `cache_` + `latestProgress` path. (This is
   independent of point 1 — it's broken whether or not topology is mandatory.)
   Relatedly, the dtor's try-catch should wrap only `onProgress_` — the swallow
   exists for a misbehaving user callback, not for our own `computeProgress`.
   Building the progress outside the catch stops a builder invariant violation
   from being masked as "Final progress report failed". That does make such a
   throw escape the (noexcept) destructor → `terminate`, so the fallback above
   must be fixed first so normal teardown can't throw. Same split applies to
   `poll()`: the rule is swallow only the external callback, never our own logic.
   Factor it into a `reportProgress(progress)` helper that wraps only
   `onProgress_` in the try-catch, and call it from both `poll()` and the dtor
   after building the progress outside the catch.

3. **`StageProgress::index` / `StageTopology::index` is redundant — it's always
   the array position.** `toStageTopology` sets `index = i`; `computeProgress`
   copies it; both headers document `stages[i].index == i` and forbid
   reordering/filtering; and `producers` already references stages by that same
   position. It's derived state stored next to an invariant warning to keep it in
   sync. Drop it and index by position — that also removes the `.index = 0/1`
   boilerplate from every test. If there's a real case where `index != position`,
   document it and add a test that exercises it.

4. **`ProgressReporter`'s non-movability leaks into an awkward out-param API —
   return `std::unique_ptr` instead.** It's non-movable for a good reason: the
   constructor registers a `poll` closure capturing `this` with the scheduler, so
   a move would leave the scheduler calling a moved-from address. But
   `startProgressReporter` works around that by taking a
   `std::optional<ProgressReporter>&` out-param and emplacing into it, with a
   comment apologizing for the contortion. Hold it behind a pointer — a
   `unique_ptr` moves even though the object doesn't — so the factory returns
   naturally (`auto progress = startProgressReporter(...)`) and the out-param
   takes both apologetic comments with it (the header doc's "built in place
   rather than returned" and the inline `.cpp` note).

5. **`progressCallbackFiredDuringExecution`: misnamed, and the id check is weak.**
   The interval is the 500ms default and the query finishes in microseconds, so
   no periodic poll lands — the one report it asserts is the synchronous teardown
   report; the periodic path is untested, so rename it. And
   `EXPECT_FALSE(progressQueryId.empty())` only inspects the last report — capture
   the id from an `onStart` callback and `EXPECT_EQ` every report's `queryId` to
   it inside the callback. (Non-throwing `EXPECT_*` record fine through the
   reporter's catch; just keep throwing checks out of the callback.)

6. **Runtime info belongs at the end of error messages, and the tests should
   assert the full message.** `"Exchange input of stage {} names unknown producer
   task prefix {}"` interleaves the stage index and prefix with the static text —
   CODING_STYLE wants the runtime values after the static description (same for
   the "disagree in size" and "Unclassified source operator" `VELOX_CHECK`s). And
   the throw-tests match only a substring; assert the full message so the wording
   and the runtime values are both covered.

7. (minor) `Runner::stats()` doc has a stray period: "This may be called at any
   time. before waitForCompletion() or abort()." → "at any time before
   waitForCompletion() or abort()."

mbasmanova:
Follow-up on point 5: the periodic path can be covered deterministically, not
just renamed around.

`executeSelectOrInsert` is public and hands back the `LocalRunner`, and
`TestConnector` can build a table from several `RowVector`s — one split per
vector. Together they let a test run a query slowly, one split at a time, and
check progress between batches: build a `ProgressReporter` around the returned
runner with a `kMinProgressReportIntervalMs` interval and a started scheduler,
then drive `next()` one batch at a time with a sleep past the interval between
calls. The poll fires between splits while the runner is `kRunning`, so the
callback observes `completedSplits` / `scanRows` advancing one split at a time —
ending at all splits completed and the final report finished, every report
carrying the query id.

That gives real coverage of the scheduler-driven poll, the split accounting, and
the `cache_` write — none of which has a test today (`QueryProgressBuilderTest`
only covers the pure builder, and `progressCallbackFiredDuringExecution` only
hits teardown).

singcha:
Thanks @mbasmanova! Updated to incorporate review comments.

mbasmanova (changes requested):
Thank you for iterating. Code changes look good. Tests need a bit tightening.

1. **Drop `progressReportedOnTeardown`.** The `run()` path is covered by many
   existing tests; all this one uniquely guards is the single line where `run()`
   forwards `onProgress`, and it proves only one teardown report (`run()` drains
   in a tight loop, so no periodic poll can land). Not worth a standalone test —
   if that one line is worth guarding, set `.onProgress` on an existing `run()`
   callback test instead.

2. **Merge the two paced tests into one.** They overlap — both assert teardown
   accounting over a real runner via `executeSelectOrInsert`, and the join's leaf
   scan is a superset of the 2-stage scan; only `progressPolledDuringTwoStageScan`
   drives the periodic poll. Run the hash join under a `ProgressReporter` +
   scheduler with pacing so one test covers the poll path, the mid-run
   invariants, and the 3-stage teardown accounting. (Its comment's reference to a
   non-existent `progressPolledBetweenSplits` goes away too.)

3. **Assert a progress update on each `next()` iteration — and wait for it, don't
   sleep.** After each pull, block until the callback delivers the next report
   (signal a baton/CV from `onProgress`) with a timeout that fails the test if
   none arrives, then assert the counters advanced. That replaces both the blind
   `sleep_for` and the loose end-of-run `ASSERT_GE(reports.size(), 2)`: stronger
   coverage and no timing flake. It also removes the direct `computeProgress`
   calls the join test uses for its mid-run and teardown snapshots — those just
   re-test the builder (already covered field-by-field by
   `QueryProgressBuilderTest`) and bypass the reporter; take both snapshots from
   the callback and assert on the data each delivered report carries — the
   counters advancing per iteration and the final report's exact stage accounting
   (point 4).

4. **Pin the parallelism the exchange counts depend on.** Most counts are
   data-derived and fine (`scanRows=48`, the scan split counts, output `16`). But
   the exchange fan-out counts assume the current task/partition count — gather
   `totalSplits=4` (one per producer task), broadcast `shuffleRows=64` (dim rows ×
   probe tasks) — and would change under, e.g., single-node/single-thread. The
   test passes `{}` (default options), so set the relevant parallelism explicitly
   to anchor those numbers.

5. **Trim the test comments.** Several of the new tests carry multi-paragraph
   headers — derivation of every magic number, threading rationale, what each
   assertion proves. Keep the non-obvious *why*; the assertions plus a line of
   context carry the rest.

6. **Use `t`, `u`, `v` for test table names** rather than `batched` / `dim` /
   `fact`, and write the inner join as `FROM t, u WHERE t.k = u.k` rather than
   `t JOIN u ON ...`.

singcha:
Thanks @mbasmanova!
Per-split observation isn't reachable at the SqlQueryRunner level (executeSelectOrInsert hardwires the connector split source, which feeds all splits up front, so the driver races to completion before the poll). 

Moved the paced test to the runner level as a new ProgressReporterTest, where LocalRunner takes an injectable split-source factory. A small gated SplitSource (coroutine co_awaits a baton) releases one scan split per signal, so with a single worker/driver the runner genuinely blocks between splits and the scheduler's async poll lands on each stable intermediate. The test now asserts, from the callback, completedSplits/scanRows advancing exactly one split at a time (1→4, 100→400) ending finished — no sleeps, deterministic across runs. The run() forwarding is guarded by adding .onProgress to an existing callback test, and multi-stage/exchange accounting stays covered field-by-field in QueryProgressBuilderTest.

mbasmanova (approved):
1. `waitForCompletedPast`: on a 10s timeout with `reports` still empty,
   `reports.back()` dereferences an empty vector (UB). `EXPECT_TRUE(ok)` records
   the failure but execution falls through into `back()`. Guard it so a hang
   fails cleanly instead of crashing.

2. `velox::ROW({"c0"}, {velox::BIGINT()})` can be
   `velox::ROW("c0", velox::BIGINT())` (single-column overload).

PR #1499, opened 2026-06-18
feat(cli): Show live query progress in the CLI (#1499)

singcha:
Summary:

Subscribes the axiom CLI to the `onProgress` callback and renders a live, in-place status display while a query runs, mirroring the Presto CLI: a summary line, a progress bar, and a per-stage grid. Output goes to stderr.

`Console` owns the decision of whether to show progress, gated on stderr being a terminal and the runner being able to report progress (`SqlQueryRunner::supportsProgress()`, true when a `folly::FunctionScheduler` was supplied). The interactive REPL shows the grid by default; the non-interactive paths (`--query`, piped stdin) opt in via `--show_live_progress`, and `--init` (setup statements) and `--repeat` (perf-measurement runs) never draw it. The grid renders on stderr and is erased before results print, so redirected stdout stays clean, and it is hidden when stderr is redirected (scripted or captured runs).

The renderer itself, `LiveProgressDisplay`, just draws to a stream -- it has no file-descriptor or terminal dependency. `Console` passes the target width to each `update()` (re-queried per frame so the grid tracks a terminal resize), which keeps the renderer testable against a plain stream and keeps the terminal policy in one place. `Console` is constructed with just the runner, so a runner built without a scheduler reports `supportsProgress() == false` and gets no display -- callers that cannot report progress (such as tests) need no extra wiring.

`LiveProgressDisplay` exposes `update(info, width)` to redraw the block and `clear()` to erase it before results print; display knobs live in a small `Options` struct (`showSplitAndCpuDetail`). The stage-depth walk that indents the grid is a private detail of the renderer: the stage graph is a tree (each stage feeds exactly one consumer), so a stage's depth is one past its consumer's.

A representative mid-execution frame:

```
Query 20260622_184329_00000_b3udk, RUNNING, 32 splits
0:17 [32.5M rows, 6.11GB] [1.91M rows/s, 368MB/s] [===================>>>>>>>>>>>>>>>>>>>>>  ] 14/32

     STAGES   ROWS  ROWS/s  BYTES  BYTES/s  QUEUED    RUN   DONE
0.........P      0       0     0B       0B       0      0      4
  1.......P      0       0     0B       0B       0      4      0
    2.....R  3.45M    203K  54.8M    3.22M       0      4      4
      3...R  13.5M    793K   206M    12.1M       0      4      4
      4...F   450K       0   169M       0B       0      0      1
        5.R  10.6M    624K  4.32G     260M       2      4      0
        6.F   4.5M       0  1.37G       0B       0      0      1
```
Final frame after completion (live display cleared, only results remain on stdout):

```
-------------+--------------+----------+-----------+------------------
l_returnflag | l_linestatus |      cnt |       qty |               rev
-------------+--------------+----------+-----------+------------------
A            | F            |  7403889 | 188818373 | 283107483036.1195
N            | F            |   192439 |   4913382 | 7364213967.950009
N            | O            | 14997114 | 382400594 | 573408429286.9312
R            | F            |  7406353 | 188960009 | 283310887148.1993
(4 rows in 1 batches)
```

Reviewed By: mbasmanova

Differential Revision: D108945711

### DIFF
--- a/axiom/cli/AxiomSql.cpp
+++ b/axiom/cli/AxiomSql.cpp
@@ -16,6 +16,7 @@
 
 #include <fmt/ranges.h>
 #include <folly/container/F14Map.h>
+#include <folly/executors/FunctionScheduler.h>
 #include <folly/init/Init.h>
 #include <gflags/gflags.h>
 #include <filesystem>
@@ -41,7 +42,10 @@ int main(int argc, char** argv) {
       facebook::velox::memory::MemoryManager::Options{});
 
   facebook::axiom::Connectors connectors;
-  axiom::sql::SqlQueryRunner runner{axiom::sql::SystemUser::resolve()};
+  // Progress-polling scheduler for the console's live progress bar.
+  folly::FunctionScheduler progressScheduler;
+  axiom::sql::SqlQueryRunner runner{
+      axiom::sql::SystemUser::resolve(), &progressScheduler};
   runner.initialize([&]() {
     VELOX_USER_CHECK(
         FLAGS_data_path.empty() || FLAGS_etc_dir.empty(),
--- a/axiom/cli/CMakeLists.txt
+++ b/axiom/cli/CMakeLists.txt
@@ -17,6 +17,7 @@ add_library(
   CatalogProperties.cpp
   Connectors.cpp
   Console.cpp
+  LiveProgressDisplay.cpp
   QueryIdGenerator.cpp
   SqlQueryRunner.cpp
   SystemUser.cpp
--- a/axiom/cli/Console.cpp
+++ b/axiom/cli/Console.cpp
@@ -17,12 +17,14 @@
 #include "axiom/cli/Console.h"
 #include <fmt/core.h>
 #include <folly/FileUtil.h>
+#include <sys/ioctl.h>
 #include <unistd.h>
 #include <algorithm>
 #include <iostream>
 #include <iterator>
 #include <optional>
 #include <set>
+#include "axiom/cli/LiveProgressDisplay.h"
 #include "axiom/cli/ResultPrinter.h"
 #include "axiom/cli/StdinReader.h"
 #include "axiom/cli/Timing.h"
@@ -66,6 +68,13 @@ DEFINE_bool(
     "Print per-statement timing (parse / optimize / execute). Always on in "
     "interactive mode.");
 
+DEFINE_bool(
+    show_live_progress,
+    false,
+    "Draw the live progress grid for --query and piped-stdin runs (when stderr "
+    "is a terminal). The interactive REPL always shows it; this only opts the "
+    "non-interactive paths in.");
+
 DEFINE_int32(
     repeat,
     1,
@@ -77,6 +86,22 @@ DEFINE_int32(
 using namespace facebook::velox;
 
 namespace {
+// Terminal width assumed when the real width cannot be queried (output is not a
+// tty or the ioctl fails).
+constexpr int kDefaultTerminalWidth = 80;
+
+// Returns the terminal width for `fileDescriptor`, or kDefaultTerminalWidth if
+// it cannot be determined. Re-queried per progress frame so the grid tracks a
+// live terminal resize.
+int terminalWidth(int fileDescriptor) {
+  struct winsize windowSize{};
+  if (ioctl(fileDescriptor, TIOCGWINSZ, &windowSize) == 0 &&
+      windowSize.ws_col > 0) {
+    return windowSize.ws_col;
+  }
+  return kDefaultTerminalWidth;
+}
+
 // Extracts a file path argument from a dot-command string like ".run <file>".
 // Returns the trimmed path, or empty string if none.
 std::string parseDotCommandPath(const std::string& command, size_t prefixLen) {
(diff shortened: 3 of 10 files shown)


mbasmanova (changes requested):
1. **Put the "show progress?" decision in `Console`, and drop `ProgressPrinter`'s
   self-disable.** Today it's split and misplaced: `showProgress` is hardcoded
   `true`, and the only thing stopping progress in batch runs is the printer's own
   `isatty` no-op ("wire unconditionally and let it disable itself"). That couples
   terminal policy into the renderer and masks the real gap — progress is still
   wired for `--query` / piped stdin / `--init`, and only stays hidden when stderr
   *also* isn't a terminal. `Console` already has the context (`interactive =
   isatty(STDIN_FILENO)`); enable progress only for the interactive REPL with a
   terminal stderr, and let `ProgressPrinter` just render. (Bonus: with the
   self-disable gone the tests no longer need a pty — render to a plain stream.)

2. **`stageDepths` models a topology that can't occur.** The stage graph is a
   tree — each producer feeds exactly one consumer, and diamonds aren't valid in
   Velox (we established this in #1498) — so the longest-path BFS, the
   diamond/longest-path handling, and the `diamondTakesLongestPath` test all
   cover a shape that can't happen. A plain "depth = consumer's depth + 1" walk
   from the roots is enough. (The out-of-range-producer guard is worth keeping —
   a transient gap mid-poll is real, unlike a diamond.) It also doesn't need to
   be exposed: its only production caller is `renderStageGrid` in the same `.cpp`
   — it's in the header purely for the test. Move it to the anonymous namespace
   and drop the `StageDepthsTest` cases; the indentation they check is verified
   through the rendered grid (point 4).

3. **`out` and `fileDescriptor` read as unrelated but must be the same stream.**
   The ctor takes a stream and a separate fd the caller has to keep in sync
   (`std::cerr` + `STDERR_FILENO`); nothing in the names ties them. Name them to
   show the pairing — e.g. `out` / `outFd` — or pass them as one unit so they
   can't mismatch.

4. **Test the written output via a provided stream, not a terminal.** The pty
   tests mostly exercise terminal *detection* (isatty), which isn't the
   interesting part. Once the printer just prints (point 1), the natural test
   feeds a crafted `QueryProgress`, renders to a plain `ostringstream`, and
   asserts what's written — the frame structure and the formatter boundaries
   (`"1.5M"`, `"6.11GB"`, the indentation, the bar). And assert it *exactly* —
   strip the ANSI control codes, split into lines, and match line-by-line with
   `ElementsAre(Eq(...))` as in `DerivedTablePrinterTest` — rather than the loose
   `HasSubstr` checks, so column alignment and the bar are actually pinned. That
   also covers the non-trivial formatting helpers (`formatCount` /
   `formatDataSize` / `formatProgressBar` / …), today only spot-checked via two
   substrings, without exposing them or needing a pty.

5. **`ProgressPrinter` undersells what it is.** "Printer" reads like a one-shot
   formatter, but this owns a live, in-place status block that it redraws each
   update and clears on finish. A name that conveys the stateful live display
   would fit better — e.g. `LiveProgressDisplay` / `ProgressDisplay`.

6. **`bool debug` is too cryptic — give the ctor an `Options` struct.** A bare
   `bool` (the call site just passes `FLAGS_debug`) says nothing about what it
   toggles. An `Options` struct with a named flag for the per-split/CPU-time
   detail lines reads self-documenting at the call site and gives future knobs a
   home — or, if it stays a single toggle, just name the bool specifically (e.g.
   `showSplitAndCpuDetail`).

7. `finish()` → `clear()`: it clears the drawn block, and `clear()` pairs
   naturally with `update()` and fits the live-display framing (point 5).

singcha:
Thanks @mbasmanova! Updated the code to reflect review feedback.

mbasmanova (commented):
Thanks for iterating.

1. **Take the terminal width as a parameter instead of `outFd`?** `outFd` exists
   only to `ioctl(TIOCGWINSZ)` for the width — passing the width in drops the fd
   dependency entirely (the display becomes just `out` + a width), removes the
   `out`/`outFd` must-match invariant, and the tests no longer need an fd. One
   tradeoff: width is currently re-queried each `update()`, so it tracks a live
   terminal resize; a fixed ctor width loses that. If resize matters, pass the
   width to `update()` or take a width-provider callback; otherwise fixing it at
   construction is fine.

2. **Gating:** `showProgress = isatty(STDERR_FILENO) && supportsProgress()` isn't
   gated on `interactive`, so `--query` and piped-stdin runs still draw the grid
   whenever stderr is a terminal (only `--init` and `--repeat` are excluded).
   That cleanly handles redirected/captured output — the original concern — but
   it's broader than the interactive REPL. Is showing progress for `--query` /
   piped-stdin (with a terminal stderr) intended? If you want REPL-only, also
   gate on `interactive`.

3. The header `@code` example couples the display to `SqlQueryRunner`
   (`RunOptions`, `runner.run`). `LiveProgressDisplay` only depends on
   `runner::QueryProgress` — write the example in terms of its own API (call
   `update(info)` per snapshot, `clear()` when done), without `SqlQueryRunner`.

4. The `ConsoleTest.cpp` change — `Console console{*CHECK_NOTNULL(runner.get())}`
   plus the `<glog/logging.h>` include — is unrelated to this PR (and the
   `CHECK_NOTNULL` on a just-created `unique_ptr` is gratuitous). Drop it.

5. **Tests — cut boilerplate and tighten the assertions.** Every case repeats the
   same `ostringstream` / `LiveProgressDisplay` / `update` / `renderLines`
   four-liner — extract a `render(info, options = {})` helper returning the lines.
   And assert the full frame with `ElementsAre(...)` throughout: `rendersGridFrame`
   already does, but `formatsSiUnits`, `debugAddsSplitAndCpuLines`, and
   `ordersStagesRootFirstAndIndentsByDepth` use `ASSERT_GE(lines.size(), N)` +
   `lines[i]` index access, which is fragile (a layout shift silently mis-indexes)
   and weaker. (Minor: the repeated static `STAGES … DONE` header literal could be
   a shared constant.)

singcha:
Thanks @mbasmanova! Updated the PR.
* width is now captured per update() call (re-queried each frame), so live terminal resize is preserved
* live progress with --query is now disabled by default and can be enabled with flag --enable_live_progress

mbasmanova (approved):
Thanks for iterating.

1. **Drop the `namespace runner = facebook::axiom::runner;` alias from the
   header.** A namespace alias (like `using`) in a header leaks into every
   translation unit that includes it. Spell out `facebook::axiom::runner::` in
   the declarations; keep the alias in the `.cpp` only.

2. (nit) Consider renaming `--enable_live_progress` to `--show_live_progress`
   for brevity and accuracy.

singcha:
Thanks @mbasmanova! fixed to address latest review comments

PR #1504, opened 2026-06-22
fix: Handle suffix/metadata table separators in the file names (#1504)

srsuryadev:
Summary:

* Handle suffix/metadata table separators ('$') in the file names.
* Adds validation when the metadata table is added as well not include this

Reviewed By: mbasmanova

Differential Revision: D109184438


srsuryadev:
cc: @mbasmanova, @apurva-meta, @PingLiuPing

mbasmanova:
CC: @PingLiuPing

mbasmanova (dismissed):
1. **Shorten the `parseName` header doc** — the current "last '$' … not at the
   start … bare table-name token (no '/' and no '.') … so a '$' inside a directory
   name … or a file name …" is hard to follow. Lead with the rule and let examples
   carry the edge cases:

   ```cpp
   /// Splits "<path>$<suffix>" into {path, suffix}. Takes the part after the last
   /// '$' as the suffix only when that '$' isn't the first character and the part
   /// is a bare token (no '/' or '.'); otherwise the whole name is the path with
   /// an empty suffix. Examples:
   ///   "/data/file.parquet"               -> {"/data/file.parquet", ""}
   ///   "/data/file.parquet$column_chunks" -> {"/data/file.parquet", "column_chunks"}
   ///   "/data/$backup/file.parquet"       -> {"/data/$backup/file.parquet", ""}
   ///   "/tmp/foo$bar.parquet"             -> {"/tmp/foo$bar.parquet", ""}
   ```

2. **Drop the `FileTable.cpp` comment on `parseName`.** Docs belong on the `.h`
   declaration; the `.cpp` comment ("It takes a fully-qualified file table name
   and splits it into {filePath, metadata table}:") duplicates it.

3. **Duplicate validity logic.** The "valid metadata-table token" rule lives in
   two places: `parseName` (`!empty && no '/' && no '.'`) and `addMetadataTable`
   (`!empty && find_first_of("$./") == npos`). Extract one predicate (e.g.
   `isValidMetadataTableName`) used by both so the rule can't drift — this also
   resolves the charset-style mismatch.

4. **Residual ambiguity worth a word:** an extensionless file named `…/foo$bar`
   parses as path `…/foo` + suffix `bar`. The docs cover `$`-in-directory and
   `$`-in-file-with-extension but not this case — confirm it's acceptable or note
   it as a known limitation.

5. **`separatorInFileNameReadsDataTable` — trim, rename, fix the temp file.** The
   parse is already covered by `FileTableTest.parseNameDollarInFileName`, so the
   e2e only needs to confirm a `$`-named file reads through the connector:
   - Drop the full row-vector setup. The `ORDER BY id` only forces order for the
     order-sensitive `assertEqualVectors`; if you keep a result check, use the
     set-based `assertEqualResults` (order-insensitive) and drop the `ORDER BY`.
   - `$weird.parquet` is non-descriptive (what's "weird"? the point is a `$` in
     the name) — name it for the scenario, like the unit test's `foo$bar.parquet`.
   - The cleanup is ad-hoc and not RAII: `getPath() + "$…"` then a manual
     `std::remove(path.c_str())` leaks the file if an assertion above it fails. Use
     a managed temp file (or `SCOPE_EXIT`) so cleanup is automatic.

6. **Test placement.** Two things:
   - `FileHandlerTest.cpp` holds `FileHandlerRegistryTest` — the file name doesn't
     even match its suite (pre-existing) — and the PR adds a second suite,
     `FileHandlerMetadataTableTest` (+ `TestFileHandler`). Distinct suites don't
     share a file, and a file should match its suite: rename the existing one to
     `FileHandlerRegistryTest.cpp` and put the new suite in its own
     `FileHandlerMetadataTableTest.cpp`.
   - `FileHandler` lives in `connectors/file/`, but the test sits under
     `connectors/file/core/tests/` (that's for `file/core/` code like `FileTable`).
     The new file belongs in `connectors/file/tests/`, next to
     `FileConnectorTest.cpp`.

7. **Combine the two metadata-name tests.** `rejectsNamesWithReservedCharacters`
   and `acceptsBareName` are one concern — fold them into a single, shorter-named
   test (e.g. `validatesName`): the reserved-char/empty cases via
   `VELOX_ASSERT_THROW`, the bare name via
   `ASSERT_NO_THROW(handler.registerTable("row_groups"))`.

mbasmanova:
1. `isValidMetadataTableName` is a free function — make it a `static` method on
   `FileTable` next to `parseName`, since it's the same name-scheme contract.
   `FileTable::isValidName` reads fine; the class supplies the "metadata table
   name" context.

2. In `registry`, `EXPECT_NE(std::find(names.begin(), names.end(), "parquet"),
   names.end())` is a roundabout "contains" check. Use
   `EXPECT_THAT(names, ::testing::Contains("parquet"))` (link `GTest::gmock`,
   include `<gmock/gmock.h>`) — clearer, and drops the `<algorithm>` include.

3. `separatorInFileName` only checks `result->size() == 3` on a `SELECT *`. That
   proves the query succeeded (pre-fix it would throw "Unsupported metadata
   table") but not that the `$`-named file's data was read — it could be 3 rows of
   anything. Assert the content; or, for a simpler check, `SELECT count(*) FROM …`
   and assert the returned count is 3.

mbasmanova (approved):

axiom/connectors/file/tests/CMakeLists.txt
```diff
@@ -25,3 +25,16 @@ target_link_libraries(
...
+
+add_test(NAME axiom_file_handler_test COMMAND axiom_file_handler_test)
+
+target_link_libraries(
+  axiom_file_handler_test
+  axiom_file_connector
+  axiom_file_handler_parquet
+  velox_type
+  GTest::gtest
+  GTest::gtest_main
```
PingLiuPing, line 39: Should we link GTest::gmock as well since FileHandlerTest now uses `EXPECT_THAT`?

PR #1505, opened 2026-06-22
test(file-connector): Test DESC and SHOW SCHEMAS (#1505)

srsuryadev:
Summary:

Add test coverage for DESC and SHOW SCHEMAS on the file connector. `listSchemaNames` and `schemaExists` were covered only at the API level, and `DESC` had no coverage at all, so the describe and show-schemas paths a user hits through SQL were unverified.

Differential Revision: D109184101

### DIFF
--- a/axiom/connectors/file/tests/FileConnectorTest.cpp
+++ b/axiom/connectors/file/tests/FileConnectorTest.cpp
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-#include <folly/ScopeGuard.h>
 #include <folly/init/Init.h>
 #include <gtest/gtest.h>
 
@@ -22,9 +21,8 @@
 #include "axiom/cli/SqlQueryRunner.h"
 #include "velox/common/base/tests/GTestUtils.h"
 #include "velox/common/file/FileSystems.h"
+#include "velox/common/testutil/TempDirectoryPath.h"
 #include "velox/dwio/parquet/writer/Writer.h"
-#include "velox/exec/tests/utils/QueryAssertions.h"
-#include "velox/exec/tests/utils/TempFilePath.h"
 #include "velox/vector/tests/utils/VectorTestBase.h"
 
 namespace facebook::axiom::connector::file {
@@ -50,8 +48,8 @@ class FileConnectorTest : public ::testing::Test, public test::VectorTestBase {
       return std::make_pair(std::string(kConnectorId), std::string("parquet"));
     });
 
-    tempFile_ = exec::test::TempFilePath::create();
-    parquetPath_ = tempFile_->getPath() + ".parquet";
+    tempDir_ = velox::common::testutil::TempDirectoryPath::create();
+    parquetPath_ = fmt::format("{}/test.parquet", tempDir_->getPath());
     writeTestParquetFile(parquetPath_);
   }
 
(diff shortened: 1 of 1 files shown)


mbasmanova (dismissed):
1. **`resultRowCount`'s failure check discards the real error.**
   `VELOX_CHECK(!result.message.has_value(), "Query produced no result: {}", sql)`
   — `message.has_value()` is the query *failure* signal (an empty result, the
   case this helper exists for, doesn't set it), so print `*result.message`
   rather than the misleading hardcoded "produced no result". It also repeats
   `query()`'s run-and-check; both could share one `run(sql)` helper.

2. **The file management is overbuilt — use a temp directory.** The
   `extraFiles_` + `extraPaths_` parallel vectors, the per-file `std::remove`s in
   `TearDown`, and the `TempFilePath`-plus-`".parquet"` stub all go away if the
   fixture holds one `velox::common::testutil::TempDirectoryPath`, writes every
   parquet file into it, and lets its destructor delete the directory and its
   contents.

3. **`showSchemasUnknownConnectorFails` tests unrelated code.** "SHOW SCHEMAS
   FROM nonexistent" fails in generic connector-metadata resolution before any
   file-connector code runs — it doesn't exercise this connector. Drop it.

4. **Combine `describePreservesColumnOrder`, `describeReportsScalarTypes`, and
   `describeReportsNestedTypes`.** Three near-identical create-table-DESC-assert
   blocks for not much coverage — one table with out-of-order columns spanning
   scalar and nested types exercises ordering, scalar rendering, and nested
   rendering at once.

5. **`query(fmt::format("DESC {}", ref))` repeats across every DESC test — add a
   `describeTable(name)` helper.** Then add `assertDescribeTable(name, schema)`
   that derives the expected (column, type) result from a `RowType` schema, so
   the tests stop hand-building the column-name and type-string vectors.

6. **Same for the SHOW SCHEMAS tests** — add a helper (e.g.
   `assertShowSchemas(sql, expectedSchemas)`) so they stop repeating the
   `query(...)` + `makeRowVector({"Schema"}, ...)` boilerplate.

7. **`describeNonexistentFileFails` wraps a constant in `fmt::format`.** The only
   argument is a string literal — write the `DESC file."parquet"."/tmp/..."`
   string directly.

8. **Trim the scaffolding comment in `describeUnsupportedFormatFails`.**
   "This is the 'unknown schema' path SHOW SCHEMAS cannot exercise (the registry
   is append-only)" contrasts the test against another's coverage — the kind of
   note CODING_STYLE says to drop. The rest of the test comments earn their lines.

mbasmanova (dismissed):
Thanks for the cleanup — the helpers and the temp-directory handling read much
better.

1. The `describe` test folds three error paths into the happy-path rendering
   test — split them out. `Unsupported file format`, `Unsupported metadata
   table`, and the nonexistent-file case are error behavior, and the file already
   keeps that kind of case as its own test (`unsupportedMetadataTableFails`).
   Combining the three accounting cases was the ask, but folding the error paths
   in too makes a mega-test where an early `assertEqualVectors` failure masks the
   later cases and the test name no longer says what broke. Move them to their
   own test (e.g. `describeErrors`).

2. `run()` returns the full `SqlResult`, but both callers (`query`,
   `resultRowCount`) use only `.results` (after `run()` already checked
   `.message`). Return `std::vector<velox::RowVectorPtr>` instead.

3. `resultRowCount` → `countResultRows` (start with a verb).

4. In `assertDescribeTable`, drop the `columns` vector — use `schema->names()`
   for the column-name column — and build `types` with a range-for over
   `schema->children()` (`for (const auto& child : schema->children()) types.push_back(child->toString());`).

mbasmanova:
Stepping back: much fewer and simpler tests would cover this. The suite is
over-built — exhaustive matrices and many near-duplicate cases — and the comments
are far too verbose (restating the SQL and the asserts), which buries what's
actually being tested. Pare it down to the file-connector behavior that needs
coverage; specifics below.

1. The big scalar/nested type matrix re-verifies shared functionality — the list
   of Velox types and `Type::toString()` rendering — which Velox already tests;
   it isn't file-connector behavior. The connector test only needs to confirm
   DESC resolves the file schema and reports its columns/types (a representative
   column or two via `assertDescribeTable`). Drop the exhaustive matrix and the
   "pin literally to catch format changes" rationale — that's Velox's job.

2. The `describe` comment even contradicts itself — "file-schema order (not
   sorted)" then "Columns are written out of alphabetical order."

3. `assertShowSchemas` takes a `std::vector<std::string>` but every caller passes
   a single schema (`{"parquet"}`). Take a single `const std::string&`.

4. The `makeFlatVector<std::string>(...)` calls over `std::vector<std::string>`
   *variables* don't need the explicit `<std::string>` — `T` deduces from the
   argument. (Only braced-init-list literals need it.)

5. Title and summary. The title — "Additional cases to cover DESC and SHOW
   SCHEMAS through SQL" — is long and doesn't say this is the *file connector*
   specifically (reads like generic DESC/SHOW SCHEMAS coverage). Scope it, e.g.
   `test(file connector): Test DESC and SHOW SCHEMAS`. The summary is redundant
   too — e.g. "Add test coverage for DESC and SHOW SCHEMAS on the file
   connector."

mbasmanova (approved):
Thanks.

PR #1508, opened 2026-06-22
test(file-connector): Extend tests for multi-row-group metadata, null-count statistics, and batch streaming (#1508)

srsuryadev:
Summary:

Extend tests for multi-row-group metadata, null-count  statistics, and batch streaming in file-connector

Differential Revision: D109179258

### DIFF
--- a/axiom/connectors/file/tests/FileConnectorTest.cpp
+++ b/axiom/connectors/file/tests/FileConnectorTest.cpp
@@ -15,6 +15,7 @@
  */
 
 #include <folly/init/Init.h>
+#include <gmock/gmock.h>
 #include <gtest/gtest.h>
 
 #include "axiom/cli/Connectors.h"
@@ -23,6 +24,7 @@
 #include "velox/common/file/FileSystems.h"
 #include "velox/common/testutil/TempDirectoryPath.h"
 #include "velox/dwio/parquet/writer/Writer.h"
+#include "velox/exec/tests/utils/QueryAssertions.h"
 #include "velox/vector/tests/utils/VectorTestBase.h"
 
 namespace facebook::axiom::connector::file {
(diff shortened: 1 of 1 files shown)


mbasmanova:
CC: @PingLiuPing

mbasmanova (dismissed):
The title and summary describe this as a test-only change
(`test: Extend tests…`), but the PR actually contains two production changes:

- `fix(file-connector): Label nested Parquet leaf columns by dotted path` —
  changes `ParquetFileHandler`.
- `feat(file-connector): Query a directory of files as one table` — adds
  directory support across `FileHandler` / `FileConnectorMetadata` /
  `FileDataSources`.

Please lift the fix and the feature into their own PRs. The test commit is fine
as a standalone `test:` PR — it adds missing coverage for existing behavior — so
this PR can stay scoped to that. As written, though, two production changes are
hidden behind a `test:` label.

axiom/connectors/file/core/FileConnectorMetadata.cpp
```diff
@@ -73,8 +79,10 @@ FileConnectorMetadata::SplitManager::getSplitSource(
     QueryRuntimeStats& /*runtimeStats*/) {
   auto* fileHandle = dynamic_cast<const FileTableHandle*>(tableHandle.get());
   VELOX_CHECK_NOT_NULL(fileHandle, "Expected FileTableHandle");
+  auto& fileHandler = handler(fileHandle->schemaTableName().schema);
```
PingLiuPing: I think there is a metadata contract bug here. The query may run and return rows, but the returned metadata loses the identity needed to interpret those rows safely. In a single Parquet file, `row_group_id` is naturally file-local:
file a.parquet
  row_group_id = 0
  row_group_id = 1
But the PR makes a directory path like this behave as one logical table:

```
SELECT * FROM file."parquet"."/dir/";
SELECT * FROM file."parquet"."/dir/$row_groups";
SELECT * FROM file."parquet"."/dir/$column_chunks";
```

The new split source emits one split per file. Then each metadata data source builds metadata independently for each file. That means two files each start their Parquet row group numbering at zero:
a.parquet
  row_group_id = 0

b.parquet
  row_group_id = 0
So the directory metadata result can become:
row_groups
+--------------+----------+
| row_group_id | num_rows |
+--------------+----------+
| 0            | 2        |
| 0            | 3        |
+--------------+----------+
That is ambiguous because there is no file_path, file_index, or globally unique row group id.

The bigger problem is when someone uses two metadata tables together.
column_chunks
+--------------+-----------+------+
| row_group_id | column_id | name |
+--------------+-----------+------+
| 0            | 0         | id   |
| 0            | 0         | id   |
+--------------+-----------+------+
If a user joins row_groups to column_chunks on row_group_id, every row_group_id = 0 chunk can match every row_group_id = 0 row group across files. The result silently mixes metadata from different files.

PingLiuPing (commented):
Thanks.
Agree with @mbasmanova 's comment. PR title and description should be updated.

axiom/connectors/file/core/FileConnectorMetadata.cpp
srsuryadev: Thanks @PingLiuPing! To disambiguate this will add "file_path" as the identifier column, and also will share and move the discussion to separate PR (this PR got compounded with two changes due to OSS tooling).

mbasmanova (dismissed):
1. `ORDER BY` is being added to tests where ordering is irrelevant just to make the assertion deterministic — `projection`'s `ORDER BY name` has nothing to do with projection, and `fullScan`'s `ORDER BY id` likewise. The root cause is `query()` + `assertEqualVectors`, which is order-sensitive and single-vector. For the content checks, `exec::test::assertEqualResults({expected}, results)` compares row multisets across all batches (order-insensitive) — what `separatorInFileName` used before this PR; for a check that genuinely needs order there's `assertResultsOrdered`. Either way the incidental `ORDER BY`s and the hand-rolled `concatenate()` aren't needed (in `streamingMultipleBatches` the file order is just a single-driver scheduling artifact anyway).

2. `streamingMultipleBatches` is too loose to test what it claims. `EXPECT_GT(results.size(), 1)` only checks there are 2+ batches — not 3 of sizes 2/2/1 — so it pins almost nothing about batching. And it can't distinguish the connector honoring the requested output batch size from just returning one row group at a time, since the fixture's row-group size (2) equals `max_output_batch_rows` (2) (both yield 2/2/1). It does confirm the data scan streams (vs the single-batch full-materialization `DataSource`), but to actually pin the behavior: make the output batch size differ from the row-group size and assert the exact batch count and per-batch sizes.

3. The switch from the 3-arg `parquet::Writer(sink, options, schema)` to the 4-arg `(sink, options, rootPool_, schema)` looks unrelated to the row-group work — `options.memoryPool` is already `rootPool_` and both overloads exist, so the 3-arg form keeps the diff focused unless the temp-pool argument is needed.

4. `asRowType(data->type())` can be `data->rowType()` (`RowVector` exposes it directly).

5. Prefer `makeFlatVector<std::string>({...})` over `makeFlatVector<StringView>({...})` for the string-literal columns — avoids the `StringView`-into-literal lifetime footgun and matches the common test idiom.

6. `RunOptions options; options.numWorkers = 1; options.numDrivers = 1;` can be `RunOptions options{.numWorkers = 1, .numDrivers = 1};`.

7. The title repeats "in file-connector" (already in the `test(file-connector)` scope) and is long — worth shortening.

PingLiuPing (approved):
The diff no longer tests ORDERR by. Please change the PR title and description.

srsuryadev:
cc: @PingLiuPing, @mbasmanova

mbasmanova (dismissed):
In `streamingMultipleBatches`, the `batchSizes` vector + `ElementsAre` is overkill — just `EXPECT_EQ(batch->size(), 1)` inside the loop (the 5-row `assertEqualResults` already pins the total).

Several test comments restate what the body and expected output already show — please do a full sweep, not just this one. E.g. `describe`'s "DESC resolves the file path through the handler and reports the file schema as (column, type) rows in file-schema order."

`separatorInFileName` hand-builds `file."parquet"."{}"` inline because `table()` is bound to `parquetPath_`. Give `table()` an optional path argument (default `parquetPath_`) so it reads `table(path)` and the reference format lives in one place, like the rest of the file.

Two leftover redundancies in the test helpers:

- `writeParquetFile` and `writeTestParquetFile` are effectively one operation. `writeParquetFile`'s only caller is `writeTestParquetFile`, so its generality (arbitrary `data`, `rowsInRowGroup`) is unused, and the two near-identical names convey no difference. Collapse into one.

- `run()` and `queryAll()` are the same modulo the options arg — `run(sql)` == `queryAll(sql)` with default options, and `run()` is now only used by `query()`. Route `query()` through `queryAll()` and drop `run()`. And `queryAll`'s comment ("Runs 'sql' with optional run options and returns all of its result batches") just restates the signature — drop it, or keep only the non-obvious part (fails with the query's own error message).

mbasmanova (approved):
All addressed — thanks. LGTM.

PR #1514, opened 2026-06-24
Enforce query timeout in SqlQueryRunner fetchResults (#1514)

singcha:
Summary:

Axiom SqlQueryRunner RunOptions.timeoutMicros is used for post-execution waitForCompletion cleanup, not for aborting running queries. As a result queries with 1s timeout run 72s to completion.

This change adds proper timeout enforcement at Axiom query runner level. Adds generic RunnerException in axiom/runner usable by any runner type (LocalRunner, SqlQueryRunner, etc).

Differential Revision: D109355308

### DIFF
--- a/axiom/cli/SqlQueryRunner.cpp
+++ b/axiom/cli/SqlQueryRunner.cpp
@@ -37,6 +37,7 @@
 #include "axiom/optimizer/Plan.h"
 #include "axiom/optimizer/RelationOpPrinter.h"
 #include "axiom/optimizer/VeloxHistory.h"
+#include "axiom/runner/RunnerException.h"
 #include "axiom/sql/presto/PrestoParser.h"
 #include "axiom/sql/presto/PrestoSqlError.h"
 #include "axiom/sql/presto/ShowStatsBuilder.h"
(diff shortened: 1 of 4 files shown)


mbasmanova (changes requested):
The timeout has to be enforced by the Runner, not by a watchdog above it — and
that means changing the `Runner` interface, not just moving this code.

Make `next()` coroutine-based (`folly::coro::AsyncGenerator<RowVectorPtr>` /
`co_next()`), and state in the interface's written contract that a well-behaved
implementation must support *meaningful* cancellation — cancellation actually
interrupts in-flight work, it isn't best-effort-at-the-next-boundary. With that
contract, the deadline is just `folly::coro::timeout(runner.co_next(), deadline)`:
cancellation rides the standard `CancellationToken` machinery, composes with
every other reason a query gets cancelled, and surfaces the timeout as a
`RunnerException(kQueryTimedOut)` from a single await.

`next()` is where the query blocks, so it's the only place a deadline can be
honored by *waiting less* instead of being forcibly interrupted from outside.
Enforcing it in the CLI forces the opposite: a side thread that `abort()`s the
runner. Almost everything in this diff exists only to compensate for that
inversion —

- the per-query `std::thread` watchdog (create + join on every `run`),
- `abortRunnerOnce` + the shared atomic to serialize the abort across two
  threads,
- the clock-skew reconciliation ("treat `aborted` as a timeout even if my clock
  reads under the limit"), needed only because there are now two timers that
  must agree,
- the "is a null `next()` end-of-stream or an abort?" ambiguity, papered over
  with a third `throwTimeoutIfExceeded()` after the loop.

None of that is intrinsic to time-limiting a query. With a cancelable
`co_next()`, `fetchResults` goes back to a plain loop and the runner *knows* it
timed out rather than inferring it from a race.

This also resolves an inconsistency the PR half-acknowledges: `RunnerException`
is deliberately placed in `axiom/runner/` as reusable by any runner, but the
enforcement stays in `SqlQueryRunner.cpp` and is `LocalRunner`-specific (it
leans on `abort()`). Timeout is a runner concept — put the mechanism where the
exception already lives, so every runner gets it instead of re-implementing the
watchdog.

That `LocalRunner::next()` has no deadline today is the work to do, not a reason
to work around it one layer up.

mbasmanova:
CC: @amitkdutta

PR #1525, opened 2026-06-26
fix: Handle nested leaf columns by dotted path (#1525)

srsuryadev:
Summary:
X-link: https://github.com/facebookincubator/velox/pull/17979

Fixes how column names show up for nested fields (structs, lists, maps) when you inspect a Parquet file's per-column details. Now each nested column is labeled by its real path inside the file.

Example:
```
  SELECT column_id, path, encodings
  FROM file."parquet"."nested.parquet$column_chunks" ORDER BY column_id;

   column_id | path                      | encodings
  -----------+---------------------------+-----------------------
   0         | [id]                      | [RLE, PLAIN]
   3         | [address, city]           | [RLE, PLAIN]
   5         | [tags, list, element]     | [RLE, RLE_DICTIONARY]
   6         | [lookup, key_value, key]  | [RLE, RLE_DICTIONARY]
```

Reviewed By: mbasmanova

Differential Revision: D109186690

### DIFF
--- a/axiom/connectors/file/README.md
+++ b/axiom/connectors/file/README.md
@@ -123,9 +123,9 @@ compression codec, byte sizes, value count, and column-chunk statistics.
 |--------|------|-------------|
 | `row_group_id` | BIGINT | Zero-based row group index. |
 | `column_id` | BIGINT | Zero-based column index within the row group. |
-| `name` | VARCHAR | Column name from the file schema. |
+| `path` | ARRAY<VARCHAR> | Leaf column-chunk physical `path_in_schema` as ordered segments (e.g. `['address', 'city']`, `['tags', 'list', 'element']`, `['lookup', 'key_value', 'key']`). See [Nested column paths](#nested-column-paths). |
 | `compression` | VARCHAR | Compression codec (e.g. `none`, `zstd`, `snappy`, `gzip`). |
-| `encodings` | VARCHAR | Comma-separated Parquet page encodings (e.g. `RLE,PLAIN`, `RLE_DICTIONARY`). |
+| `encodings` | ARRAY<VARCHAR> | Parquet page encodings as a list (e.g. `['RLE', 'PLAIN']`, `['RLE_DICTIONARY']`). |
 | `compressed_size` | BIGINT | Total compressed byte size of the column chunk. |
 | `uncompressed_size` | BIGINT | Total uncompressed byte size of the column chunk. |
 | `num_values` | BIGINT | Number of values (including nulls) in the column chunk. |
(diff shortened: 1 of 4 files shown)


srsuryadev:
cc: @PingLiuPing

axiom/connectors/file/parquet/ParquetFileHandler.cpp
```diff
@@ -238,6 +238,46 @@ ColumnChunkStats readColumnChunkStats(
...
+  return prefix.empty() ? std::string(segment)
+                        : fmt::format("{}.{}", prefix, segment);
+}
+
+// Flattens 'type' into its leaf columns in document order, matching the order
+// of column chunks in a Parquet row group. Nested fields are named by their
+// dotted path: struct fields join with '.', list elements append ".element",
+// and map entries append ".key"/".value" (e.g. "address.city", "tags.element",
+// "lookup.key").
+void collectLeafColumns(
```
PingLiuPing: Wondering have you explored this https://github.com/facebookincubator/velox/blob/61a9053bcac2d54ce0c30e8c2cd17f51da08c76d/velox/dwio/parquet/thrift/parquet.thrift#L728 ? I think this is what you need.

axiom/connectors/file/parquet/ParquetFileHandler.cpp
```diff
@@ -238,6 +238,46 @@ ColumnChunkStats readColumnChunkStats(
...
+    const std::string& prefix,
+    const TypePtr& type,
+    std::vector<LeafColumn>& leaves) {
+  if (type->kind() == TypeKind::ROW) {
+    const auto& rowType = type->asRow();
```
PingLiuPing: Here the rowType is a normalized logical schema, but we need phycial parquet schema.

e.g. The physical paths written by Parquet are tags.list.element and lookup.key_value.key

PingLiuPing (commented):
Thanks for the code and for involving me. Let one comment on the design.

srsuryadev:
@PingLiuPing Updated the PR to use pathSchema

srsuryadev:
cc: @PingLiuPing, @mbasmanova

axiom/connectors/file/README.md
```diff
@@ -123,7 +123,7 @@ compression codec, byte sizes, value count, and column-chunk statistics.
 |--------|------|-------------|
 | `row_group_id` | BIGINT | Zero-based row group index. |
 | `column_id` | BIGINT | Zero-based column index within the row group. |
-| `name` | VARCHAR | Column name from the file schema. |
+| `name` | VARCHAR | Column name from the file schema. Nested fields use their dotted path: struct fields join with `.`, list elements append `.element`, and map entries append `.key`/`.value` (e.g. `address.city`, `tags.element`, `lookup.key`). |
```
PingLiuPing: This should be updated to reflect the new names, e.g. `tags.list.element` and `lookup.key_value.key`.

axiom/connectors/file/tests/FileConnectorTest.cpp
```diff
@@ -15,6 +15,7 @@
  */
```
PingLiuPing, line 15: It seems the diff in this file comes from https://github.com/facebookincubator/axiom/pull/1508/changes.

PingLiuPing (commented):
Left one comment in `README.md`.

Others looks good to me.

You might need to advance velox first once your velox PR landed.

mbasmanova (dismissed):
Reviewing the second commit (nested leaf columns).

- PR description is too thin — one line plus an X-link. Describe the bug (nested columns fell back to `column_N` with no stats), the fix (flatten to leaves, label by `path_in_schema`, decode typed stats), and the resulting naming convention. Include a concrete example — a nested schema and the exact `column_chunks` names it produces — the convention is hard to grasp without one.

- The README and the code disagree on nested paths. The README says logical — "list elements append `.element`, map entries append `.key`/`.value`" (`tags.element`, `lookup.key`) — but the code emits the physical `path_in_schema` with Parquet's synthetic group levels: `tags.list.element`, `lookup.key_value.key`. Reconcile them, and decide whether exposing `list`/`key_value` to users querying this metadata is intended, or whether the logical path (per the README) is what we want. Either way the README needs a worked example (a nested schema and the exact names produced) whose output matches the code.

`collectLeafTypes` can be generic — a type with `size() == 0` is a leaf, otherwise recurse over `childAt(i)`. That matches the switch exactly (ROW children in order, ARRAY element, MAP key then value) without per-kind cases:

```cpp
void collectLeafTypes(const TypePtr& type, std::vector<TypePtr>& leafTypes) {
  if (type->size() == 0) {
    leafTypes.push_back(type);
    return;
  }
  for (auto i = 0; i < type->size(); ++i) {
    collectLeafTypes(type->childAt(i), leafTypes);
  }
}
```

Other points:

- Fixtures: rather than a separate `writeNestedParquetFile` (and a third fixture inlined in `columnChunksDeeplyNestedSchema`), fold complex-type columns into the single standard fixture so the existing `column_chunks` / statistics / `fullScan` tests exercise nesting directly. One fixture with simple + struct + array + map (plus an array-of-struct or map-of-struct for the recursive branch) covers every path. Keep the 2/2/1 row counts so `streamingMultipleBatches` / `rowGroupsMetadata` are unaffected — nested columns add leaf chunks, not rows.

- The new tests use `test::assertEqualVectors` + `ORDER BY column_id`, but commit 1 moved the sibling column-chunk tests to `exec::test::assertEqualResults` (order-independent, no `ORDER BY`). Match that.

- Subtle: the chunk name is read per-chunk from `path_in_schema` (physical), but the type is `leafTypes.at(col)` from the positional flattening of `rowType()`. They line up only if the logical leaf order matches the physical chunk order; the `VELOX_USER_CHECK_EQ` guards the count but not the order, so a reordering would silently pair a chunk with the wrong type's stats.

PingLiuPing (approved):
The second commit looks good.
Please rebase the code once https://github.com/facebookincubator/axiom/pull/1508 landed.

srsuryadev:
@PingLiuPing Thank you for the review updated the PR. (cc: @mbasmanova for final approval)

mbasmanova (dismissed):
Encoding the leaf path into the `name` column as a dot-joined string is misleading and ambiguous:

- It's a path, not a name — `name` is the wrong column for it.
- The `.` separator collides with field names containing dots, and the synthetic `list` / `element` / `key_value` levels are indistinguishable from real struct fields of those names (`tags.list.element` and a struct `tags{list{element}}` are identical strings).

Parquet's `path_in_schema` is a list of strings — expose it as such: an `ARRAY<VARCHAR>` column named `path` (or `path_in_schema`), rather than flattening to a dot-joined `name`. That removes the separator ambiguity and stops labeling a path as a name. (Distinguishing synthetic repeated-group levels from real fields is inherent to `path_in_schema`, out of scope here — but the array form avoids the added collision.)

mbasmanova (dismissed):
path→`ARRAY<VARCHAR>` is the right shape, and the README rationale is clear. Two remaining points:

- `makePathArray` hand-rolls the `ArrayVector` (offsets/sizes/elements). `BaseVector::createFromVariants(ARRAY(VARCHAR()), pathVariants, pool)` — each entry a `Variant::array` of the segments — builds it directly and drops the manual bookkeeping. It's a metadata table, so the Variant overhead doesn't matter.
- `encodings` has the same issue this PR just fixed for `path`: a list (`chunk.encodings()`) flattened to a delimited string (`"RLE,PLAIN"`). For consistency it should be `ARRAY<VARCHAR>` too — comma is safe-ish for encoding names, but it's the same lossy list-as-string, now inconsistent with `path` in the same table.

Also, the PR description has no example output — add one showing a nested-schema query (the `path` array for a struct/array/map leaf), so the behavior change is visible from the description.

mbasmanova (approved):
All addressed — `path` and `encodings` are both `ARRAY<VARCHAR>` built through one `makeStringArray` / `createFromVariants` helper, and the description has the example. LGTM.

PR #1526, opened 2026-06-27
feat(file-connector): Query a directory of files as one table (#1526)

srsuryadev:
Summary:

A path ending in `/` is now read as a directory: the connector scans every data file in it and returns them as one logical table. Previously only a single file path resolved to a table. For example:

    SELECT * FROM file."parquet"."/data/parquet/";

The directory is listed once during planning, sorted, and each remaining file becomes one split. Entries skipped from the listing:

- Names beginning with `_` or `.`.
- Subdirectories, detected from filesystem metadata during listing.
- Entries the handler's format probe rejects — for Parquet, anything without `PAR1` magic at both ends.

The table schema is taken from the first file. Every other file is validated against it during planning: the file schema must match exactly — a missing, extra, renamed, or reordered column, or a column at a different type, fails upfront naming the offending file, before any rows are read. A genuine I/O error while probing an entry propagates rather than the file being silently dropped.

The `$row_groups` and `$column_chunks` metadata tables gain a `file_path` column. Per-file identifiers such as `row_group_id` restart at 0 in each file, so `(file_path, row_group_id)` is what keeps rows from different files distinct.

Reviewed By: mbasmanova

Differential Revision: D109189535

### DIFF
--- a/axiom/connectors/file/FileHandler.cpp
+++ b/axiom/connectors/file/FileHandler.cpp
@@ -16,10 +16,12 @@
 
 #include "axiom/connectors/file/FileHandler.h"
 
+#include <algorithm>
 #include <map>
 
 #include "axiom/connectors/file/core/FileTable.h"
 #include "velox/common/base/Exceptions.h"
+#include "velox/common/file/FileSystems.h"
 
 namespace facebook::axiom::connector::file {
 
@@ -78,6 +80,48 @@ const velox::RowTypePtr& FileHandler::metadataSchema(
   VELOX_USER_CHECK(
       it != metadataTables_.end(), "Unsupported metadata table: {}", suffix);
   return it->second.schema;
+}
+
+std::vector<std::string> FileHandler::listFiles(const std::string& path) const {
+  if (path.empty() || path.back() != '/') {
+    return {path};
+  }
+
+  auto fileSystem = velox::filesystems::getFileSystem(path, /*config=*/nullptr);
+  auto entries = fileSystem->list(path);
+  std::vector<std::string> files;
+  files.reserve(entries.size());
+  for (auto& entry : entries) {
+    const auto slash = entry.find_last_of('/');
+    const std::string_view baseName = std::string_view(entry).substr(
+        slash == std::string::npos ? 0 : slash + 1);
+    if (baseName.empty() || baseName.front() == '_' ||
+        baseName.front() == '.') {
+      continue;
+    }
+    // Skip nested directories, detected from filesystem metadata rather than by
+    // trying to read them, so genuine read errors on real files are not
+    // swallowed as "not a data file".
+    if (fileSystem->isDirectory(entry)) {
+      continue;
+    }
+    // Let the handler's isDataFile() reject entries that are not data files of
+    // its format (e.g. foreign formats). The base accepts every entry.
+    if (!isDataFile(*fileSystem, entry)) {
+      continue;
+    }
+    files.push_back(std::move(entry));
+  }
+  std::sort(files.begin(), files.end());
+  VELOX_USER_CHECK(
+      !files.empty(), "No data files found in directory: {}", path);
+  return files;
+}
+
+bool FileHandler::isDataFile(
+    velox::filesystems::FileSystem& /*fileSystem*/,
+    const std::string& /*filePath*/) const {
+  return true;
 }
 
 std::unique_ptr<velox::connector::DataSource> FileHandler::create(
--- a/axiom/connectors/file/FileHandler.h
+++ b/axiom/connectors/file/FileHandler.h
@@ -24,6 +24,10 @@
 
 #include "velox/connectors/Connector.h"
 
+namespace facebook::velox::filesystems {
+class FileSystem;
+}
+
 namespace facebook::axiom::connector::file {
 
 /// Factory function that creates a metadata DataSource.
@@ -60,6 +64,14 @@ class FileHandler {
   /// Returns the fixed schema for a metadata table by $-suffix.
   const velox::RowTypePtr& metadataSchema(const std::string& suffix) const;
 
+  /// Lists the data files for 'path'. A path that does not end in '/' is a
+  /// single file and yields itself. A path ending in '/' is a directory; its
+  /// entries are returned sorted, so a directory of files can be queried as one
+  /// table. Entries whose base name begins with '.' or '_', nested
+  /// subdirectories, and entries rejected by isDataFile() are skipped, so a
+  /// directory may freely hold sidecar files and nested partitions.
+  std::vector<std::string> listFiles(const std::string& path) const;
+
   /// Creates a DataSource for reading from a file. Dispatches to
   /// createDataSource() for row data or the registered factory for
   /// metadata tables.
@@ -84,6 +96,14 @@ class FileHandler {
       const velox::RowTypePtr& fullSchema,
       const velox::connector::ColumnHandleMap& columnHandles,
       velox::memory::MemoryPool* pool) const = 0;
+
+  // Returns whether 'filePath' is a file of this handler's format, used by
+  // listFiles() to skip entries that are not data files. The base accepts every
+  // entry; formats override this to recognize their own files (e.g. Parquet by
+  // its "PAR1" magic).
+  virtual bool isDataFile(
+      velox::filesystems::FileSystem& fileSystem,
+      const std::string& filePath) const;
 
  private:
   struct MetadataTable {
--- a/axiom/connectors/file/README.md
+++ b/axiom/connectors/file/README.md
@@ -19,6 +19,7 @@ what synthetic metadata tables to expose. See the `FileHandler` interface in
 
 **Reads:**
 - Scan row data from a supported file via its absolute file path.
+- Scan a directory of files (trailing-slash path) as one logical table.
 - Auto-detect schema from the file header.
 - Column projection — only the requested columns are decoded.
 - Streaming reads — row data is returned one batch at a time, not buffered in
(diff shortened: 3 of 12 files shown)


srsuryadev:
cc: @PingLiuPing

axiom/connectors/file/README.md
```diff
@@ -96,7 +114,9 @@ Lists one row per row group with size and row-count statistics.
 
 | Column | Type | Description |
 |--------|------|-------------|
-| `row_group_id` | BIGINT | Zero-based row group index within the file. |
+| `file_path` | VARCHAR | Absolute path of the file the row group belongs to. |
```
mbasmanova: It seems like file name would be sufficient and more convenient to use.

Why do we need file_index - what's the order of files in a directory - sorted alphabetically? of order is not fixed, file indices won't be stable, hence, not very useful.

axiom/connectors/file/README.md
```diff
@@ -45,6 +46,23 @@ FROM file."parquet"."/data/file.parquet"
 WHERE name = 'alpha';
 ```
 
+### Reading a directory of files
+
+A path ending in `/` is treated as a directory. The connector reads every file
+with the handler's extension (e.g. `*.parquet`) as one logical table, taking the
```
mbasmanova: Is it conventional to use .parquet extensions? I don't think so.

axiom/connectors/file/README.md
srsuryadev: Initially, I had the file_index for brevity. Especially when users working with the tables, it can be very handy during merge. 

However, after your comment removed them in the latest version to simplify the code and avoid ambiguity on sorting on some cases.

axiom/connectors/file/README.md
srsuryadev: Updated the code to use PAR1 magic number to pick the parquet files in the directory. https://parquet.apache.org/docs/file-format/ 

cc: @PingLiuPing

axiom/connectors/file/FileHandler.cpp
```diff
@@ -80,6 +82,33 @@ const velox::RowTypePtr& FileHandler::metadataSchema(
   return it->second.schema;
 }
 
+std::vector<std::string> FileHandler::listFiles(const std::string& path) const {
```
PingLiuPing, line 85: `listFiles()` only skips names beginning with . or _. 
What if there is subdirectory?
What if there is an ordinary visible non-Parquet file.
Should we do the PAR1 check here? The current patch check the PAR1 only after the entry has become a split. But we can do it here to exclude them earlier.
Please add tests with both a visible non-Parquet file and a nested directory.

axiom/connectors/file/core/FileConnectorMetadata.cpp
```diff
@@ -23,23 +23,30 @@
...
   auto [filePath, suffix] = FileTable::parseName(tableName.table);
 
-  auto schema = suffix.empty() ? fileHandler.resolve(filePath, pool_.get())
-                               : fileHandler.metadataSchema(suffix);
+  velox::RowTypePtr schema;
+  if (suffix.empty()) {
+    // A directory resolves to the files it contains; take the schema from the
+    // first file and assume the rest share it.
+    schema = fileHandler.resolve(
+        fileHandler.listFiles(filePath).front(), pool_.get());
```
PingLiuPing: Should we cache the listFiles result as immutable list so that it can be reused by getSplitSource?
Is it possible that entries changed between those calls?

axiom/connectors/file/FileHandler.h
```diff
@@ -60,6 +60,13 @@ class FileHandler {
   /// Returns the fixed schema for a metadata table by $-suffix.
   const velox::RowTypePtr& metadataSchema(const std::string& suffix) const;
 
+  /// Lists the data files for 'path'. A path that does not end in '/' is a
+  /// single file and yields itself. A path ending in '/' is a directory; its
+  /// entries are returned sorted, so a directory of files can be queried as one
+  /// table. File extensions are not required (the schema already selects the
+  /// format).
+  std::vector<std::string> listFiles(const std::string& path) const;
```
PingLiuPing, line 73: This looks like a helper method, does it make sense to make it private?

axiom/connectors/file/README.md
PingLiuPing: Should we strictly check that the file ends with magic bytes "PAR1" as well?

axiom/connectors/file/FileHandler.cpp
srsuryadev: Thanks @PingLiuPing, updated the PR to do PAR1 check in the listFiles

axiom/connectors/file/FileHandler.h
srsuryadev: ListFiles is used in FileConnectorMetadata as well!

srsuryadev:
thanks @PingLiuPing for the comments and addressed them

srsuryadev:
cc: @PingLiuPing, @mbasmanova

axiom/connectors/file/core/FileConnectorMetadata.cpp
```diff
@@ -23,26 +23,34 @@
...
-  auto schema = suffix.empty() ? fileHandler.resolve(filePath, pool_.get())
-                               : fileHandler.metadataSchema(suffix);
+  // Resolve the path to its data files once, here during planning, and thread
+  // the list through the table handle to split generation, so the directory is
+  // listed — and each entry's format probed — only once per query.
+  auto filePaths = fileHandler.listFiles(filePath);
+
+  // A directory takes its schema from the first file and assumes the rest share
+  // it; each file is re-checked against this schema as it is read.
+  velox::RowTypePtr schema = suffix.empty()
```
PingLiuPing: We should validate metadata suffixes before probing directories. For an unsupported metadata suffix on a missing or empty directory, this fail while listing/probing the path with error message "No data files found in directory: " instead of "Unsupported metadata table:".

axiom/connectors/file/parquet/ParquetFileHandler.cpp
```diff
@@ -38,8 +39,29 @@ namespace {
...
+// with the 4-byte magic. Used as the directory-listing filter, so any entry
+// that is not a readable Parquet file is skipped rather than fatal. A
+// subdirectory qualifies: on a local filesystem opening it succeeds and only
+// the read fails ("Is a directory"), so both the open and the read are guarded.
+bool isParquetFile(
+    velox::filesystems::FileSystem& fileSystem,
+    const std::string& filePath) {
+  try {
+    auto file = fileSystem.openFileForRead(filePath);
+    return file->size() >= kParquetMagic.size() &&
```
PingLiuPing: This only check file header. Footer is missing.

srsuryadev:
Thanks @PingLiuPing for quick review.
* Addressed the comments by adding validate metadata suffixes before probing.
* Added footer check as well.

cc: @mbasmanova

PingLiuPing (approved):
Thanks.

mbasmanova (dismissed):
This connector opens every file during planning (the `isDataFile` magic probe), which would normally be too slow for a production data path. That's acceptable here because this is an exploration/debugging tool — anyone who needs performance should register a Hive table and use the Hive connector. So the ask below is about behavior, not cost.

Since we already open every file at planning time, validate each against the table schema there and fail with a single upfront error, instead of the read-time per-file check. As written, a later file's mismatch surfaces mid-read, so a query can stream rows from earlier files and then error — confusing, and avoidable. Reading each footer at plan time is cheap at this scale; drop the read-time `validateFileSchema` and the "may emit rows then fail" caveat in the description.

Small: `FileHandler::isDataFile` base returns `true`, but `listFiles`'s comment says it skips subdirectories — the base doesn't (only the Parquet override does, via open/read failure). Reconcile the comment or the default.

mbasmanova (dismissed):
`validateFileSchema` is under-specified and order-dependent: it matches columns by name and checks each with `equivalent` (name-insensitive for complex types, so renamed nested fields pass and then the by-name nested read misreads) while ignoring extra columns (silently dropped — and since the schema comes from the first sorted file, the exposed columns depend on listing order). Just use `RowType`'s `==`: `fileType == schema` compares names and types recursively, so it's order-independent and nested-safe, and collapses the check to one comparison per file.

`isParquetFile` decides control flow by catching exceptions — it opens every entry and any failure returns `false`. That's programming by exception, and it silently drops real data files on genuine errors (permissions, transient remote reads), so the query returns incomplete results with no error. Detect directories from the listing metadata rather than by catching the read failure; then the magic-bytes check needs no catch-all — a non-Parquet file simply fails the magic comparison, and real I/O errors propagate instead of being swallowed.

The PAR1 magic check is also duplicated — `isParquetFile` (start + end) and `openFile` (start). Factor it into one helper so the two can't drift.

mbasmanova (approved):
Substance looks good — schema check uses `fileType == schema`, directories via `isDirectory` (no exception-catching), the PAR1 magic is shared via `hasParquetMagic`, and the metadata test uses `path`.

A few things:

- `validateFileSchema` is now a single `VELOX_USER_CHECK` — it doesn't need to be a separate helper (with its own long comment). Inline it into the `findTable` loop.

- The `listFiles` / `isDataFile` docs over-claim and repeat. `listFiles` already excludes directories and `.`/`_` names, so the `isDataFile` doc shouldn't re-narrate that. Drop the speculative bits too — "cheap header probe" (a remote open isn't necessarily cheap) and "file extensions are not required" (a handler's `isDataFile` may well require one). Keep the docs to the actual contract.

- `file_path` stores the full absolute path in every metadata row, repeating the queried directory prefix on every row. The directory is already known from the query, so store the file name (path relative to the queried directory) instead — shorter values, and `(file_path, row_group_id)` still distinguishes rows since directory entries are unique.

PR #1527, opened 2026-06-29
feat(runner): Make Runner result iteration coroutine-based (#1527)

singcha:
Summary:

Replaces the synchronous `Runner::next()` with a coroutine `Runner::execute()` returning a `folly::coro::AsyncGenerator<velox::RowVectorPtr>` that streams every result batch. This matches the codebase convention where coroutine methods (`co_getSplits`, `co_listPartitions`, `co_close`) are the only form rather than a sync/async pair, and lets callers consume a real stream (`while (auto batch = co_await gen.next())`). It is the foundation for enforcing a query deadline inside the runner (the dependent CLI diff) instead of a CLI watchdog.

Lifecycle. The owner winds a run down by awaiting `Runner::co_close()` exactly once when done with the generator -- whether it drained fully, stopped early, or was unwound by an exception -- before destroying the runner. co_close() stops anything still running and reaps it (split scope cancelled and joined, tasks completed, final stats captured, pools released); because it is awaited it never blocks the awaiting thread, so it is safe on a Velox executor thread. This mirrors the runner's own dependency `SplitSource::co_close()`: the destructor asserts co_close() ran (VELOX_CHECK) rather than doing blocking teardown itself, so no drop ever runs an unbounded blockingWait on an executor thread. Early stop is "stop pulling, then co_await co_close()". External or deadline cancellation still composes through execute()'s awaiting cancellation token (e.g. `folly::coro::timeout` or `co_withCancellation`); there is no separate public abort() and no manual waitForCompletion().

`LocalRunner::execute()` registers one `folly::CancellationCallback` that calls the internal `cancelTasks()` (to stop in-flight work on external/deadline cancellation), then pulls batches through a shared single-await `co_pull()` over `TaskCursor::moveNext(ContinueFuture*)`. Building on the terminal-only consumer signaling from the Velox fix , a resolved wait-future always yields data-or-terminal, so `co_pull()` is a single `co_await` (guarded by a `VELOX_CHECK` against a spurious re-block). `co_close()` cancels a still-running run (benign `kCancelled`), then `co_reap()` co_awaits `cancelAndJoinAsync()` on the split scope plus the task completion/deletion waits, all awaited (not blockingWait) so the executor thread is released while it waits. The split scope is a `CancellableAsyncScope` and each split coroutine checks the injected token every iteration, so teardown is bounded to at most one in-flight `co_getSplits()` rather than draining each source to noMoreSplits; the task waits are hard-bounded by kReapTimeoutMicros via `.within()`. (A hard `.within()` cap on the scope join, like the task waits, is unsafe: the scope must complete its join before destruction, so an abandoned join would abort or leave coroutines touching freed state.) The runner owns the result output pool when the caller passes none, so co_reap() can reset the cursor while batches the caller still holds stay valid until the runner is destroyed. The write path (INSERT/CTAS) is unified on the same cancelable pull via `co_runWrite()`; the produce/drain phase honors cancellation while the commit is a point of no return.

The one sync bridge, `Runner::drain(onBatch, timeoutMicros = 0)`, drives the generator with `folly::coro::blockingWait`, invokes a per-batch callback, and co_closes the runner on the calling (non-executor) thread before returning -- on success, timeout, or failure. Collect-all callers push into their own vector; folding or discarding callers never materialize a result set. When `timeoutMicros > 0`, drain() wraps the loop in `folly::coro::timeout` and fails with `VELOX_USER_FAIL` on the deadline (after reaping). Streaming callers that stop early -- hold the generator, pull one batch at a time, then `blockingWait(co_close())` on their own non-executor thread. Batches live in a pool owned by the runner, so callers that need a batch to outlive the runner copy it out (`ToGraph` constant folding, `PrestoQueryReplayRunner`). The Velox-side `TaskCursor::moveNext` keeps its sync form (Velox core has no coroutines); only the Axiom runner is coroutine-only.

Cancellation correctness: because the `CancellationCallback` may fire on any thread, all `state_`/`error_` transitions and the `stages_` mutation in `makeStages()` are serialized under `mutex_` -- the latter closes a data race between `makeStages()` populating `stages_` and a task-thread `cancelTasks()` reading it -- with `state_` kept `std::atomic` for lock-free `state()` reads. `cancelTasks()` is idempotent and never clobbers a terminal state: a completed query stays `kFinished`, a pre-start cancel makes `start()` surface the cancellation, an early stop or deadline lands the benign `kCancelled` (distinct from `kError` for a genuine execution failure), and a failed write commit records `error_` alongside `kError`.

Differential Revision: D110007170

### DIFF
--- a/axiom/cli/SqlQueryRunner.cpp
+++ b/axiom/cli/SqlQueryRunner.cpp
@@ -16,6 +16,7 @@
 
 #include "axiom/cli/SqlQueryRunner.h"
 #include <folly/container/F14Map.h>
+#include <folly/coro/BlockingWait.h>
 #include <folly/system/HardwareConcurrency.h>
 #include <algorithm>
 #include <cmath>
@@ -268,14 +269,6 @@ connector::ConnectorProperties collectConnectorProperties(
     }
   }
   return result;
-}
-
-std::vector<velox::RowVectorPtr> fetchResults(runner::LocalRunner& runner) {
-  std::vector<velox::RowVectorPtr> results;
-  while (auto rows = runner.next()) {
-    results.push_back(rows);
-  }
-  return results;
 }
 
 int64_t countRows(const std::vector<velox::RowVectorPtr>& results) {
@@ -1156,9 +1149,6 @@ std::string SqlQueryRunner::runExplainAnalyze(
       queryCtx,
       options,
       runtimeStats ? *runtimeStats : noopRuntimeStats_);
-  SCOPE_EXIT {
-    waitForCompletion(runner, options.timeoutMicros);
-  };
 
   {
     PhaseTimer phaseTimer(
@@ -1168,12 +1158,10 @@ std::string SqlQueryRunner::runExplainAnalyze(
         QueryRuntimeStats::kExecuteCpuNanos);
     auto progress =
         startProgressReporter(*runner, queryCtx->queryId(), options);
-    auto results = fetchResults(*runner);
+    // Executed for its runtime stats (printed below); the result batches are
+    // not used, so discard them instead of accumulating the whole result set.
+    runner->drain([](velox::RowVectorPtr) {});
   }
-
-  // Drive the run to completion so operator stats are final before reading
-  // them: draining the output cursor does not guarantee every driver finished.
-  waitForCompletion(runner, options.timeoutMicros);
 
   std::stringstream out;
   out << printPlanWithStats(
@@ -1402,9 +1390,6 @@ SqlQueryRunner::SqlResult SqlQueryRunner::runLogicalPlan(
       queryCtx,
       options,
       runtimeStats ? *runtimeStats : noopRuntimeStats_);
-  SCOPE_EXIT {
-    waitForCompletion(runner, options.timeoutMicros);
-  };
 
   {
     PhaseTimer phaseTimer(
@@ -1414,7 +1399,9 @@ SqlQueryRunner::SqlResult SqlQueryRunner::runLogicalPlan(
         QueryRuntimeStats::kExecuteCpuNanos);
     auto progress =
         startProgressReporter(*runner, queryCtx->queryId(), options);
-    result.results = fetchResults(*runner);
+    runner->drain([&](velox::RowVectorPtr batch) {
+      result.results.push_back(std::move(batch));
+    });
   }
 
   return result;
--- a/axiom/cli/SqlQueryRunner.h
+++ b/axiom/cli/SqlQueryRunner.h
@@ -469,20 +469,6 @@ class SqlQueryRunner {
       std::string_view queryId,
       std::string_view connectorId) const;
 
-  // Wait maxWaitMicros microseconds for the LocalRunner to complete.  If
-  // `maxWaitMicros <= 0` this will check if the LocalRunner is completed and
-  // return immediately.
-  static void waitForCompletion(
-      const std::shared_ptr<facebook::axiom::runner::LocalRunner>& runner,
-      int32_t maxWaitMicros) {
-    if (runner) {
-      try {
-        runner->waitForCompletion(maxWaitMicros);
-      } catch (const std::exception&) {
-      }
-    }
-  }
-
   // Permission check callback invoked before query execution.
   PermissionCheck permissionCheck_;
 
--- a/axiom/optimizer/JoinSample.cpp
+++ b/axiom/optimizer/JoinSample.cpp
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#include <folly/ScopeGuard.h>
+#include <folly/coro/BlockingWait.h>
 #include "axiom/optimizer/Optimization.h"
 #include "axiom/optimizer/Plan.h"
 #include "axiom/runner/LocalRunner.h"
@@ -175,21 +177,30 @@ std::unique_ptr<KeyFreq> runJoinSample(
   auto result = std::make_unique<folly::F14FastMap<uint32_t, uint32_t>>();
 
   int32_t rowCount = 0;
-  while (auto rows = runner.next()) {
-    rowCount += rows->size();
-    auto hashes = rows->childAt(0)->as<velox::SimpleVector<int64_t>>();
+  // Pull batches one at a time so sampling can stop early once it has enough
+  // rows. blockingWait runs on the optimizer planning thread, not a Velox
+  // executor thread.
+  auto generator = runner.execute();
+  // Early stop is "stop pulling, then co_close()". co_close() must run on every
+  // exit -- including when a pull throws -- so the runner is reaped before it
+  // is destroyed (its destructor asserts co_close() ran). co_close() does not
+  // throw, so this is safe during exception unwinding.
+  SCOPE_EXIT {
+    folly::coro::blockingWait(runner.co_close());
+  };
+  while (auto rows = folly::coro::blockingWait(generator.next())) {
+    const velox::RowVectorPtr& batch = *rows;
+    rowCount += batch->size();
+    auto hashes = batch->childAt(0)->as<velox::SimpleVector<int64_t>>();
     for (auto i = 0; i < hashes->size(); ++i) {
       if (!hashes->isNullAt(i)) {
         ++(*result)[static_cast<uint32_t>(hashes->valueAt(i))];
       }
     }
     if (maxRows && rowCount > maxRows) {
-      runner.abort();
       break;
     }
   }
-
-  runner.waitForCompletion(1'000'000);
   return result;
 }
 
--- a/axiom/optimizer/ToGraph.cpp
+++ b/axiom/optimizer/ToGraph.cpp
@@ -15,6 +15,7 @@
  */
 
 #include <fmt/format.h>
+#include <folly/coro/BlockingWait.h>
 #include <velox/common/base/Exceptions.h>
 #include <iostream>
 #include <ranges>
@@ -397,13 +398,14 @@ std::vector<velox::RowVectorPtr> runConstantPlan(
       noopStats);
 
   std::vector<velox::RowVectorPtr> results;
-  while (auto rows = runner->next()) {
-    VELOX_CHECK_GT(rows->size(), 0);
+  runner->drain([&](velox::RowVectorPtr batch) {
+    VELOX_CHECK_GT(batch->size(), 0);
+    // Copy out of the runner's QueryCtx pool: that pool is released when the
+    // runner is closed (drain() reaps before returning).
     results.push_back(
         std::dynamic_pointer_cast<velox::RowVector>(
-            velox::BaseVector::copy(*rows, pool)));
-  }
-  runner->waitForCompletion(1'000'000);
+            velox::BaseVector::copy(*batch, pool)));
+  });
   return results;
 }
 
--- a/axiom/optimizer/tests/AxiomSqlBenchmark.cpp
+++ b/axiom/optimizer/tests/AxiomSqlBenchmark.cpp
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+#include <folly/coro/BlockingWait.h>
 #include <folly/executors/CPUThreadPoolExecutor.h>
 #include <folly/init/Init.h>
 #include <folly/system/HardwareConcurrency.h>
@@ -292,9 +293,8 @@ class VeloxRunner : public velox::QueryBenchmarkBase {
       getrusage(RUSAGE_SELF, &start);
       MicrosecondTimer timer(&micros);
 
-      while (auto rows = runner.next()) {
-        results.push_back(rows);
-      }
+      runner.drain(
+          [&](RowVectorPtr batch) { results.push_back(std::move(batch)); });
 
       struct rusage final;
       getrusage(RUSAGE_SELF, &final);
@@ -549,9 +549,6 @@ class VeloxRunner : public velox::QueryBenchmarkBase {
     auto planAndStats = optimize(statement.plan(), queryCtx);
 
     auto runner = makeRunner(planAndStats, queryCtx);
-    SCOPE_EXIT {
-      waitForCompletion(runner);
-    };
 
     RunStats unused;
     auto results = runInner(*runner, unused);
@@ -717,9 +714,6 @@ class VeloxRunner : public velox::QueryBenchmarkBase {
 
     try {
       auto runner = makeRunner(planAndStats, queryCtx);
-      SCOPE_EXIT {
-        waitForCompletion(runner);
-      };
 
       RunStats runStats;
       auto results = runInner(*runner, runStats);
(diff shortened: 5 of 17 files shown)


mbasmanova (changes requested):
1. The consumer wait-future (added in #17930/#17947) fires spuriously, and that's the root of the wait-loop. `TaskQueue::enqueue(nullptr, ...)` — a producer signaling completion — calls `consumerPromise_.setValue()` on *every* producer finishing, not just the last. With multiple producers the consumer is woken while the queue is still empty and other producers are running; `dequeue` then finds no data, not-all-finished, and hands out *another* future. So "future resolved" doesn't imply the next call yields data — a woke-up-with-nothing-to-consume that shouldn't happen — and it's the only reason the cursor's `moveNext` and `co_pull` need an unbounded loop instead of a single await. Fix the signaling, not the loop: in the producer-done path, only fulfill the consumer promise when it's actually terminal (all producers finished / all drained); data enqueues already wake it. Then a resolved future always yields data-or-terminal on the next call, the cursor's internal loop and `co_pull` collapse to one await, and `co_pull` becomes trivial enough that point 5's Runner generator is a thin pass-through.

2. The `state_` serialization is incomplete. `abort()`, `start()`, and the task `onError` callback in `makeStages()` all mutate `state_`/`error_` under `mutex_` — and `onError` runs on a Velox task thread — but the new `co_next()` / `co_runWrite()` accesses don't take the lock: `state_ = State::kFinished` in `co_next()`, and the `SCOPE_EXIT { state_ = state; }` and `if (state_ == State::kCancelled)` in `co_runWrite()`. Per this class's own header note, producer drivers may still be closing after the output is drained, so `onError` can write `state_` from a task thread concurrently with `co_next()`'s unlocked `kFinished` write — a data race (and the cancel callback's destructor only joins the cancellation-token `abort()`, not the `onError → abort()` path, so there's no happens-before). Take `mutex_` for these accesses too (or make `state_` atomic and drop the mutex for it), so the serialization the PR adds is actually complete.

3. The `SCOPE_EXIT { state_ = state; }` + mutable `state` local + `if (state_ == kCancelled)` read-back in `co_runWrite()` is an awkward way to set the terminal state. One unconditional stamp serves two cases with opposite needs: on a drain-loop failure, `abort()`/`onError` have already set the right `state_`, so it must not be clobbered — hence the read-back; on a commit failure (commit runs outside the `try`), nothing set `state_`, so the default kError is needed. Handle them separately: on drain failure leave `state_` to `abort()`/`onError` and rethrow; on commit failure set kError (or call `abort()`) explicitly. That drops both the SCOPE_EXIT and the read-back. (These are also among the unlocked `state_` accesses in point 2.)

4. `coNextCancellation` only covers the synchronous path. A pre-cancelled token makes `folly::CancellationCallback` invoke the callback inline in its constructor, on the cursor thread — so `abort()` never actually runs cross-thread there. But cross-thread `abort()` (cancellation firing while `co_next()` is `co_await`-ing a future) is the whole reason for the new `mutex_`, and it's untested. Add a case that cancels from another thread mid-await (e.g. `folly::coro::timeout` on a running query, or request cancellation from a separate thread after the pull has started).

5. Design: prefer `folly::coro::AsyncGenerator<velox::RowVectorPtr>` over `co_next()`-per-batch for the Runner's result iteration. A "stream of batches until done" is exactly what an async generator models. The generator wraps the existing (non-coro) Velox cursor internally — it owns the `co_pull` wait-loop and `co_yield`s each batch — so Velox doesn't need to change. This way callers consume a real stream (`while (auto b = co_await gen.next())`) instead of re-looping `co_next()` at every call site, cancellation composes over the generator, and the ~10 per-site `blockingWait` calls collapse to a single drain. `co_next()`-per-call is a synchronous-iterator shape in coroutine syntax; the generator is the idiomatic form and subsumes the boilerplate in the next point.

6. The "blockingWait is safe here…" comment is copy-pasted at ~10 call sites, and that repetition is itself a smell: every synchronous caller now hand-justifies a blocking call. If the safety argument has to be restated everywhere, it should live in one place — a single documented sync drain helper (`drain(runner)` / `collect(runner)`) that owns the `blockingWait` and the rationale, with call sites just calling it. More broadly, every current caller is synchronous; the only thing that needs the coroutine is the planned query-deadline/cancellation feature, which isn't in this PR. Forcing every existing caller through `blockingWait` + a caveat to set up a not-yet-used capability suggests the coroutine-only shape is worth a second look (e.g. keep the helper as the synchronous entry point). At minimum, consolidate so the invariant is stated once rather than ten times.

7. Docs on `co_next()`:
   - The `Runner::co_next()` contract says "must not call this from a Velox executor thread" — that's misplaced and reads backwards. `co_next()` is a coroutine: it can be `co_await`-ed from anywhere, including an executor thread, without blocking (suspending a coroutine doesn't starve the thread — that's the point of coroutinizing it). The no-executor-thread restriction applies only to the synchronous `blockingWait(co_next())` drain, which is what would starve the executor. Move the restriction onto the blocking-drain note and have `co_next()`'s own contract say it may be awaited from any thread.
   - The `LocalRunner::co_next()` override comment's `Cancelable: see Runner::co_next().` is redundant — an override inherits the base contract; drop the redirect and keep only the LocalRunner-specific `First call starts execution.`

singcha:
Thanks @mbasmanova! Opened PR for Velox fix here - https://github.com/facebookincubator/velox/pull/17995
Will update this PR once Velox changes are merged and picked up

singcha:
@mbasmanova - Velox fix has landed. I have updated this PR to incorporate review comments.

mbasmanova (changes requested):
A race, a contract tension, and the `drain` API:

- Data race: `onError` reads `cursor_` without holding `mutex_` (`if (self->cursor_) self->abort();`). Per this class's own invariant, `mutex_` serializes `cursor_` — it's written under the lock in `start()` and reset under it in `waitForCompletion()`. `onError` runs on a task thread, and tasks are created in `makeStages()` before `cursor_` is assigned in `start()`, so a stage task can fire `onError` while `start()` is mid-assignment. `abort()` already re-locks and handles `cursor_ == null`, so drop the guard and call it unconditionally — that removes the race and still propagates the error to sibling tasks when the failure precedes the cursor.

- `co_runWrite`'s error path calls `waitForCompletion(1'000'000)`, a blocking `.wait()`, from inside the coroutine. That's at odds with `execute()`'s documented contract that awaiting it never blocks the thread and is safe on a Velox executor thread. Error path only, but worth reconciling (either narrow the contract or make the cleanup non-blocking).

- The mechanical `next()` → `drain()` swap regresses peak memory at call sites that don't need all batches at once. The old per-batch loop held ~one batch; `drain()` materializes the entire result set first. `SqlTestBase::run` only sums `batch->size()`, and `runSetupStatement` discards batches entirely — both now hold the whole result. `ToGraph::runConstantPlan` is a milder variant: it accumulates into its own `results`, but now holds the `drain()` vector *and* `results` at once — ~2× the result set at peak. All of these should stream one batch at a time via the generator (as `JoinSample` now does), not `drain()`.

- `AxiomSqlBenchmark::runInner` and `pyspark/LocalRunner` both do `for (auto& rows : drain(...)) results.push_back(rows);` — that's just `results = drain(...);` (a move); drop the copy loop.

- `drain`:
  - `fetchResults` and `QueryTestBase::readCursor` are now just `return drain(...);` — trivial forwarders. Drop them and call `drain` at the call sites.
  - Make `drain` a static method on `Runner` rather than a free function.
  - Reframe its doc. It shouldn't say it's for callers "that cannot be coroutines (CLI, FFI, optimizer planning)" — every caller *can* be a coroutine; it's just sometimes inconvenient. Document it as a convenience for non-production / test use.

singcha:
Thanks @mbasmanova ! Fixed the review comments. one exception:
Kept SqlQueryRunner::fetchResults — it has two callers and becomes the timeout entry point (draining under folly::coro::timeout) in next change. so it's a shared helper .

mbasmanova (changes requested):
Thanks, the earlier fixes look good. One larger design point, then a few smaller ones.

The change introduces coroutines into the API but stops halfway. It exposes an `AsyncGenerator`, yet keeps the pre-coroutine lifecycle — a separate imperative `abort()` plus manual `waitForCompletion()`, with the generator not owning its execution. The idiomatic generator model is the opposite: the generator owns teardown, so dropping it cancels and reaps the work (RAII), and cancellation composes through the awaiting scope's token — there is no separate `abort()`.

The PR's own docs surface the gap:
- "Simply dropping a partially-consumed generator leaves `state() == kRunning` until `waitForCompletion()`" — the generator doesn't own teardown.
- Early-stop needs an explicit `abort()`, which records `kCancelled` + "Query cancelled" even for a satisfied stop (`JoinSample`) — success-early and failure collapse to the same terminal state.
- `~LocalRunner()` only joins the split scope; it doesn't reap tasks.

The follow-through: dropping the generator should cancel and reap the work, deadline cancellation should compose through the awaiting scope's token, and `abort()` should become an internal detail rather than a public lever. Early-stop is then just "stop pulling." This is a rethink rather than a mechanical change — worth settling now, before the shape spreads across callers.

Smaller points:

- `drain()`: rather than returning the whole vector, have it take a per-batch callback. Folding/discarding callers (`SqlTestBase::run`, `runSetupStatement`, the benchmark, `ToGraph`) never materialize the result set, and collect-all callers push into their own vector from the callback. That folds the hand-rolled `blockingWait(generator.next())` loops onto one sync primitive and removes the whole-result materialization by construction.

- `fetchResults` is a pure forwarder to `drain()`, and the timeout would go on `drain()` itself, not a per-caller wrapper. Drop it; the two callers should call `drain()` directly.

- `execute()`'s doc says "each batch is allocated in the pool of the QueryCtx." `runConstantPlan` doesn't lean on that — it deep-copies every batch into `pool` via `BaseVector::copy`. Reconcile the two: if the copy is avoidable, drop it (hand the runner a QueryCtx whose pool outlives it, so batches land in `pool` directly); if the copy is required, say so in the doc (the QueryCtx pool's lifetime is tied to the runner, so results must be copied out to survive).

- Two doc nits on `execute()`:
  - "— this is how a query deadline is enforced without a side thread" is redundant with the preceding clause. Drop the tail.
  - "To stop before the generator is exhausted, call abort(); state() only reaches…" — the semicolon joins two independent sentences; use a full stop.

singcha:
Thanks Masha! Updated to address design point and other review comments.

mbasmanova (changes requested):
Thank you for iterating. The design mostly landed — no separate public cancel, cancellation via the scope token, and `drain` is the callback+timeout form. One issue in how teardown is implemented.

`teardown()` runs synchronous blocking work in the generator's destructor: `reap()` does `folly::coro::blockingWait(splitScope_.joinAsync())` — unbounded — plus `.wait()` on the task futures. A destructor can't be async, so RAII teardown *has* to block.

Early-drop is already a live pattern: `JoinSample` drives `execute()` with `blockingWait(generator.next())`, breaks once it has enough rows, and relies on the generator reaping on function return. That works only because it drives *synchronously, on the planning thread* — blocking teardown there is fine. But the API advertises more than that: "safe to await on a Velox executor thread" and "stop early by letting the generator go out of scope." The first coroutine caller that awaits `execute()` on an executor thread and drops early runs `blockingWait` in `~teardownGuard` on that executor thread — and since the split-scope join is unbounded, if that scope's work is on the same executor, it deadlocks. The early-drop pattern is in use today and only the caller's thread keeps it safe, so this is a time bomb, not a doc caveat.

The fix is to make the reap an awaited step rather than a destructor's job — the same pattern `SplitSource` already uses in this repo. `SplitSource::co_close()` (ConnectorSplitManager.h) is a mandatory awaited close whose destructor `VELOX_CHECK`s it ran ("co_close() must be called before destroying SplitSource"), and `LocalRunner` already calls `source->co_close()`. So the runner should mirror the shape of its own dependency:

- Expose `folly::coro::Task<void> co_close()` — the owner's terminal wind-down: it stops anything still running and joins/reaps (splits joined, tasks completed, pools released), and doesn't return until that's done. (The reap half is the `folly::coro::AsyncScope::joinAsync()` the runner already awaits internally, just surfaced as an awaited method.)
- The owner calls `co_await runner.co_close()` once when done — whether fully drained (nothing left, just reaps) or early-stop (winds down the remainder, then reaps). No separate cancel-then-close two-step.
- This is not `abort()` returning: `abort()` was a fire-and-forget external control that returned before work stopped and needed a separate wait; `co_close()` is an awaited, owner-scoped lifecycle finish. External/deadline cancellation still composes through the scope token — that invariant holds.
- The destructor no longer reaps — it asserts `co_close()` already ran, exactly as `SplitSource`'s does. So `co_close()` becomes mandatory before the generator is dropped. Every caller closes; nobody relies on the destructor, so no drop ever blocks (or deadlocks) an executor thread.
- This means the `execute()` doc's "stop early by letting the generator go out of scope" is no longer accurate — early-stop is "stop pulling, then `co_close()`." Update it.
- Synchronous callers (`drain`, `JoinSample`) `blockingWait(co_close())` on their own (non-executor) thread — `JoinSample` then just stops pulling and closes, no `CancellationSource` needed.

Also bound the `splitScope_.joinAsync()` wait like the task waits (`kReapTimeoutMicros`), so no teardown path can hang unboundedly.

mbasmanova:
Separately: a few places still hand-roll `blockingWait(generator.next())` loops that `drain()` now covers. The consume-only ones should use `drain(onBatch)`:

- `ToGraph.cpp:403` — collect → `drain([&](b){ results.push_back(copy(b)); })`
- `SqlTestBase.cpp:205` (`runSetupStatement`) — discards → `drain([](auto){})`
- `SqlTestBase.cpp:332` (`run`) — folds `numRows` → `drain([&](b){ numRows += b->size(); })` (`runAndCollect` at :256 already uses `drain`)
- `PrestoQueryReplayRunner.cpp:41` — collect+transform → `drain` with the transform in the callback
- `ProgressReporterTest.cpp:180` — per-batch asserts + `gate->post()` → same via the callback

singcha:
Thanks Masha! Incorporated the review comments.
Addressed both. co_close() is now the awaited wind-down, and the destructor asserts it ran rather than reaping — no drop runs blocking teardown on an executor thread.
On bounding the split-scope join: I didn't use .within() on it, because that's unsafe for AsyncScope — it requires the join to complete before destruction (joined_ is only set after the  barrier; ~AsyncScope CHECK-fails otherwise), so an abandoned join would abort or leave split coroutines touching freed state. Instead I made splitScope_ a CancellableAsyncScope: each split coroutine checks the injected token every iteration and co_reap() calls cancelAndJoinAsync(). So teardown is bounded to at most one in-flight co_getSplits() rather than draining the source — same "no unbounded teardown" guarantee, without the abort-on-teardown risk. The task completion/deletion waits stay hard-bounded via .within().

mbasmanova (approved):
Verified — `co_close()` is the awaited wind-down and the destructor asserts it ran, so no drop reaps on an executor thread.

And you're right that my `.within()` suggestion was unsafe: `AsyncScope` requires the join to complete before destruction, so a timed-out join would leave it unjoined (terminate) or let split coroutines touch freed state. `CancellableAsyncScope` with a per-iteration token check + `cancelAndJoinAsync()` bounds teardown to one in-flight `co_getSplits()` without abandoning the join — better than what I proposed.

LGTM.

PR #1564, opened 2026-07-10
docs: Distributed (cross-fragment) dynamic filter design (#1564)

mbasmanova:
Summary:

v2's multi-fragment plans lose the probe-side dynamic-filter reduction once the probe is partitioned, because the filter does not cross an exchange. At 4 workers this is a real regression:

    q17's subquery `lineitem` scan runs the full 60M rows instead of the single-node 60M→61k reduction — +286% CPU at sf10.

This adds the design doc for cross-fragment dynamic filtering — the plan of record for closing those regressions — and links it from `TpchV1V2PlanComparison.md`.

The shape: the optimizer places each dynamic filter at plan time (selectivity-gated, one per join key, across exchange boundaries), and the Runner aggregates the build-side filters across tasks and delivers them to the pre-shuffle probe scans. Three new `Task` calls carry the filter Task→Runner→Task; Velox does no aggregation and no task talks to another.

Reviewed By: xiaoxmeng

Differential Revision: D111435685


mbasmanova:
CC: @zhztheplayer

xiaoxmeng (approved):

PR #1589, opened 2026-07-15
feat(cli): Enforce query execution timeout in SqlQueryRunner (#1589)

singcha:
Summary:

Revives `RunOptions::timeoutMicros` as a cooperative execution deadline. The field has been unused since #1527 (coroutine-based result iteration), which replaced the`waitForCompletion(..., timeoutMicros)` cleanup wait with `drain()` and left it an orphan; even before that it bounded only the post-execution cleanup wait, never execution -- so a query with a 1s timeout could run to completion (e.g. 72s).

Both result-collecting call sites in `SqlQueryRunner` now pass `options.timeoutMicros` to `Runner::drain`, where the deadline is enforced via `folly::coro::timeout` over the pull loop, on expiry the query fails with a user error. The limit is cooperative and covers execution only -- not parsing, permission checks, or optimization -- so a non-yielding operator can overrun it. `RunOptions::timeoutMicros` is widened to `int64` to avoid the ~35 min `int32` cap.

Reviewed By: mbasmanova

Differential Revision: D112190981

### DIFF
--- a/axiom/cli/SqlQueryRunner.cpp
+++ b/axiom/cli/SqlQueryRunner.cpp
@@ -15,8 +15,11 @@
  */
 
 #include "axiom/cli/SqlQueryRunner.h"
+#include <folly/CancellationToken.h>
 #include <folly/container/F14Map.h>
 #include <folly/coro/BlockingWait.h>
+#include <folly/coro/Coroutine.h>
+#include <folly/coro/Timeout.h>
 #include <folly/system/HardwareConcurrency.h>
 #include <algorithm>
 #include <cmath>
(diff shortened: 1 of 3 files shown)


mbasmanova (approved):
@singcha The wiring is straightforward and `Runner::drain` already enforces the deadline, so this reads cleanly. A couple of notes:

The summary says the field was "previously only consumed by the post-execution cleanup wait." That's stale — it was, until #1527 ("coroutine-based result iteration") removed every `waitForCompletion(..., timeoutMicros)` call in the switch to `drain()`, leaving it an unused orphan since. This PR revives it, now as an execution deadline. Worth updating the description to something like "reviving a field unused since #1527, now an execution deadline."

PVC2 / `FbAxiomRunner`: the summary says it exposes this as `int64`. Since the field has been dead in axiom since #1527, any PVC2 caller setting it has been a no-op; now it becomes a live execution deadline. That surface isn't in this diff, so worth confirming it's consistent (int64, mapping to this field) and that no caller relied on the old behavior.

Default 0 preserves current behavior (execution is already unbounded — the base calls `drain()` with no timeout), so no behavior change there. The doc comment is honest about the cooperative caveat and the test fires the deadline deterministically — LGTM.

singcha:
Thank you @mbasmanova ! Updated the summary to reflect latest state

PR #1597, opened 2026-07-16
feat(cli): Ctrl+C cancels the query, clears the prompt line (#1597)

singcha:
Summary:

Makes Ctrl+C cancel an in-flight query while keeping the CLI alive, and clears the partially typed line at the prompt. Ctrl+D exits.

Two paths, matching how the terminal delivers Ctrl+C:
- At the prompt, linenoise runs the terminal in raw mode with ISIG cleared, so Ctrl+C is a byte (0x03), not a signal: linenoise returns nullptr with errno EAGAIN. `readCommand` branches on errno -- EAGAIN discards the partial (possibly multi-line) command and re-prompts; ENOENT/EOF (Ctrl+D) ends input.
- During execution the terminal is in cooked mode, so Ctrl+C arrives as SIGINT. A `QueryInterruptHandler` (its own header), installed by `run()` for the whole session, catches SIGINT, nudges a self-pipe (async-signal-safe), and a drain thread trips the running query's `folly::CancellationSource`. A signal handler rather than `sigwait` is used so the process-wide disposition also covers the runner's executor threads.

Cancellation is encapsulated in the runner: `RunOptions` gains a `cancellationToken` that `run()` composes onto its `co_run()` drain, so tripping the token stops the query and surfaces `QueryCancelledError`. `Console` sets that token from the interrupt source and calls `run()` -- no drain reimplemented at the call site -- and reports `QueryCancelledError` plainly as "Query cancelled." rather than as a failure.

Reviewed By: mbasmanova

Differential Revision: D112430321

### DIFF
--- a/axiom/cli/CMakeLists.txt
+++ b/axiom/cli/CMakeLists.txt
@@ -25,6 +25,7 @@ add_library(
   Connectors.cpp
   Console.cpp
   LiveProgressDisplay.cpp
+  QueryInterruptHandler.cpp
   SqlQueryRunner.cpp
   SystemUser.cpp
   ResultPrinter.cpp
--- a/axiom/cli/Console.cpp
+++ b/axiom/cli/Console.cpp
@@ -16,6 +16,7 @@
 
 #include "axiom/cli/Console.h"
 #include <fmt/core.h>
+#include <folly/CancellationToken.h>
 #include <folly/FileUtil.h>
 #include <sys/ioctl.h>
 #include <unistd.h>
@@ -25,6 +26,7 @@
 #include <optional>
 #include <set>
 #include "axiom/cli/LiveProgressDisplay.h"
+#include "axiom/cli/QueryInterruptHandler.h"
 #include "axiom/cli/ResultPrinter.h"
 #include "axiom/cli/StdinReader.h"
 #include "axiom/cli/Timing.h"
@@ -147,6 +149,14 @@ void Console::run() {
 
   VELOX_USER_CHECK_GE(FLAGS_repeat, 1, "--repeat must be at least 1");
 
+  // Install session-wide so Ctrl+C cancels the running query instead of the
+  // CLI; see QueryInterruptHandler.
+  QueryInterruptHandler interrupt;
+  interrupt_ = &interrupt;
+  SCOPE_EXIT {
+    interrupt_ = nullptr;
+  };
+
   if (!FLAGS_init.empty()) {
     std::string sql;
     auto success = folly::readFile(FLAGS_init.c_str(), sql);
@@ -225,6 +235,17 @@ bool Console::runOnce(
           [&](const QueryCompletionInfo& info) { completionInfo = info; },
   };
 
+  // Arm cancellation so Ctrl+C cancels the running query; run() honors the
+  // token.
+  if (interrupt_ != nullptr) {
+    options.cancellationToken = interrupt_->arm();
+  }
+  SCOPE_EXIT {
+    if (interrupt_ != nullptr) {
+      interrupt_->disarm();
+    }
+  };
+
   // When enabled by the caller, draw a live status grid on stderr while the
   // query runs.
   std::optional<cli::LiveProgressDisplay> progress;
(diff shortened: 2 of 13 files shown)


mbasmanova (changes requested):
Thanks @singcha — the async-signal-safety reasoning is careful, and the mechanism checks out: linenoise
sets EAGAIN on Ctrl+C / ENOENT on Ctrl+D (linenoise.c), every run() path goes through runOnce() so all
modes register the source, and SqlQueryRunner copies `VeloxException::errorCode()` into ErrorInfo so the
CLI's typed-code check works.

A few things:

**The SIGINT → cancel path has no test.** `QueryInterruptHandler` (self-pipe, sigaction, drain-thread
teardown) is the most intricate code here and nothing exercises it. It's also defined in Console.cpp and
only forward-declared in Console.h, so a test can't reach it as-is. Lift it to its own header so the
path can be unit-tested (install it, register a `CancellationSource`, `raise(SIGINT)`, assert the source
is cancelled; plus a teardown case restoring the prior disposition), or add a Console-level integration
test.

**Is `externalCancellation` deterministic?** It pre-cancels the token over `nation`, which
`addTpchTables()` leaves empty (the #1596 trap), then runs a trivial `count(*)`. Can the query complete
before the cooperative cancel is observed — the executionTimeout failure mode? If it's guaranteed to
throw, please say why; otherwise use the baton pattern from `drainCancelFromAnotherThread`. The two
LocalRunnerTest cases already look robust. (The empty table also reads as if data matters when it
doesn't.)

**The timeout-message reword looks out of scope.** Renaming "maximum time limit" → "the execution time
limit" in Runner.cpp is the only reason this PR touches executionTimeout, and it edits the same test
#1596 does (which asserts the old text). Dropping the reword removes both the collision and the test
churn, keeping this PR on Ctrl+C. If the "execution-only" clarification is wanted, do it separately.

**Remove internal references.** PVC2 / FbAxiomRunner (and the internal diff link) appear in the PR
description and in code comments — `Runner::drain`'s doc and both new tests ("clients (CLI, PVC2)", "any
client (CLI SIGINT, PVC2, etc.)"). This is a public repo; refer to "other clients" / "a follow-up"
generically.

**Consider splitting.** The generic `Runner::drain` cancellation token — runner layer, reusable,
independently tested — is separable from the CLI signal/stdin wiring, and the follow-up runner work
depends only on it. Two PRs would each be easier to review and let the runner piece land on its own.
Optional, but the size and the clean layer boundary invite it.

singcha:
Thanks @mbasmanova ! 
Splitting it into 2. First PR in stack here - https://github.com/facebookincubator/axiom/pull/1604

singcha:
Hi @mbasmanova 
Can you please review

mbasmanova (approved):
@singcha

1. Drop the `errorInfo.has_value()` guard and its comment. Every exception `run()` throws sets `errorInfo` first (the Velox / Presto / `std::exception` arms at `SqlQueryRunner.cpp:671-696`), and the only errorInfo-less path — `OperationCancelled` → `QueryCancelledError` — is caught separately above, so the guard is dead for `run()`'s exceptions. The one case it does fire — an exception thrown in the `try` outside `run()`, e.g. result printing after a successful query — it mislabels as "Query failed: unknown error". Trust the invariant; a violation should surface, not be masked.

2. Trim the `Console::run()` install comment — it re-lists the execution paths (the README owns that now) and re-explains the per-query-source design (`QueryInterruptHandler` owns that). One terse line: "Install session-wide so Ctrl+C cancels the running query instead of the CLI; see `QueryInterruptHandler`."

3. In the README, drop the "behave like `readline` / `psql`" comparison — `readline` is a library and `psql` another tool, neither known to every SQL-CLI user, and the table right below already describes each key.

4. Have the handler own the `CancellationSource` (`arm()` returning a token, `disarm()` on scope exit) so the call site doesn't juggle a source plus `setSource()` plus `getToken()`.

5. Add a comment on the `~QueryInterruptHandler` teardown race: `SIG_IGN` doesn't stop a handler already running on another thread, so it can `write()` to the write-fd after the destructor closes it (harmless `EBADF`, or a stray byte if the fd number was reused). It can't be closed from the handler side, so document it as accepted.

singcha:
Thanks @mbasmanova !
Addressed the comment in latest iteration.

PR #1604, opened 2026-07-17
refactor(runner): Typed query cancellation, drop co_drain (#1604)

singcha:
Summary:

Surfaces query cancellation as a typed exception at the source instead of
reconstructing the stop cause downstream. Previously a cancel was recorded as a
fabricated `VeloxRuntimeError` shaped like a genuine failure, and a `co_drain()`
helper reconstructed the reason from a callback plus the caller's token.

- `cancelTasks()` no longer fabricates an `error_` for a pure cancel: it records
  `kCancelled` and requests task cancellation, reserving `error_` for genuine
  failures. This also fixes an error-masking race -- a genuine error that
  co-occurs with a cancel is no longer blocked by a fabricated cancel error
  (`onError()`'s `if (error_)` guard now lets the real error win).
- The read path surfaces cooperative cancellation as `folly::OperationCancelled`
  through Velox's `TaskCursor` (changed separately, now landed). `co_pull()` also
  surfaces it directly for a cancel that lands before a cursor exists, and
  rethrows a genuine startup error rather than dereferencing a missing cursor.
- `co_drain()` and `DrainOutcome` are removed. The caller composes external
  cancellation and the deadline with plain structured concurrency (see the CLI
  diff), so the `Runner` virtual surface is unchanged. `drain()` remains a
  synchronous convenience for non-cancelling callers (tests, benchmarks),
  preserving the deadline `VELOX_USER_FAIL` contract; a reap failure no longer
  masks the original stop reason.

Relies on Velox's `TaskCursor` surfacing a requested cancel as
`folly::OperationCancelled`, which landed separately.

Reviewed By: xiaoxmeng

Differential Revision: D112582002

### DIFF
--- a/axiom/runner/LocalRunner.cpp
+++ b/axiom/runner/LocalRunner.cpp
@@ -16,6 +16,7 @@
 
 #include "axiom/runner/LocalRunner.h"
 #include <folly/CancellationToken.h>
+#include <folly/OperationCancelled.h>
 #include <folly/coro/AsyncGenerator.h>
 #include <folly/coro/AsyncScope.h>
 #include <folly/coro/BlockingWait.h>
@@ -313,6 +314,22 @@ folly::coro::Task<velox::RowVectorPtr> LocalRunner::co_pull() {
     start();
   }
 
+  // start() can return without a cursor if the run was cancelled or errored
+  // before starting. There is no cursor to surface the reason, so do it here: a
+  // genuine error as itself, a pure cancel as cooperative cancellation. Once a
+  // cursor exists, moveNext() itself surfaces a cancelled task as
+  // folly::OperationCancelled (translated in TaskCursor) and a genuine error as
+  // itself.
+  if (!cursor_) {
+    {
+      std::lock_guard<std::mutex> lock(mutex_);
+      if (error_) {
+        std::rethrow_exception(error_);
+      }
+    }
+    throw folly::OperationCancelled{};
+  }
+
   velox::ContinueFuture future = velox::ContinueFuture::makeEmpty();
   // The parallel TaskQueue wakes the consumer only when a batch is queued or
   // all producers are finished/drained, so a resolved wait-future implies the
@@ -444,11 +461,14 @@ velox::RowVectorPtr LocalRunner::makeWriteResult(int64_t rows) {
 void LocalRunner::start() {
   {
     std::lock_guard<std::mutex> lock(mutex_);
-    // A cancel (or error) that landed before start() leaves the runner terminal
-    // with no cursor; surface that error rather than starting and tripping the
-    // state precondition below.
+    // A cancel or error that landed before start() leaves the runner terminal
+    // with no cursor. Surface a genuine error here; for a pure cancel, return
+    // without starting so the read path surfaces cooperative cancellation.
     if (error_) {
       std::rethrow_exception(error_);
+    }
+    if (state_ == State::kCancelled) {
+      return;
     }
     VELOX_CHECK_EQ(state_, State::kInitialized);
   }
@@ -467,16 +487,21 @@ void LocalRunner::start() {
 
   {
     std::lock_guard<std::mutex> lock(mutex_);
-    if (!error_) {
+    // Do not adopt the cursor if a cancel or error landed while starting; that
+    // would clobber the terminal state a concurrent cancelTasks()/onError set.
+    if (!error_ && state_ != State::kCancelled) {
       cursor_ = std::move(cursor);
       state_ = State::kRunning;
     }
   }
 
   if (!cursor_) {
-    // The cursor was not set because previous fragments had an error.
+    // A cancel or error landed while starting, so the cursor was not adopted.
+    // Cancel the just-built tasks; co_pull() then surfaces the cause under
+    // 'mutex_' (a genuine error as itself, a pure cancel as cooperative
+    // cancellation). Do not read error_ here: it is written by onError() under
+    // the lock from a stage-task thread, so an unlocked read would race.
     cancelTasks();
-    std::rethrow_exception(error_);
   }
 }
 
@@ -494,6 +519,7 @@ void LocalRunner::cancelTasks() {
   // so serialize the error/state mutation under 'mutex_' against start() (which
   // reads error_ under the same lock) and a second concurrent cancelTasks().
   std::vector<std::shared_ptr<velox::exec::Task>> tasks;
+  std::exception_ptr error;
   {
     std::lock_guard<std::mutex> lock(mutex_);
     // A completed query stays finished: the cancel callback can fire after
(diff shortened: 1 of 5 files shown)


mbasmanova (changes requested):
@singcha Before the mechanics — could you help me understand the high-level goal? Who cancels a running query, from where, and why does that need to live on `Runner::drain`? A short end-to-end picture of the use case would let me review the shape, not just the composition logic.

Two things I'd like to work out with you once that's clear:

**Sync call vs. async API.** Since `drain()` blocks in `blockingWait`, cancelling it takes another thread to trip the token — which makes me wonder whether the async path is the more natural home. `execute()` already returns an `AsyncGenerator`; a caller driving it under its own `co_withCancellation` scope gets cancellation from structured concurrency, with no side-channel token or helper thread. Is there a caller that must cancel through the synchronous bridge and can't use the async generator (or an async `drain()` returning `Task<>`)?

**Signaling cancellation.** `kQueryCancelledErrorCode` — a `VeloxUserError` with a custom string code (via the internal `_VELOX_THROW`) that clients string-match — is a new pattern here, and I'm not sure cancellation should be a user error at all. Would a distinct outcome (a returned status, or a dedicated cancellation type) fit better? The deadline path has the same shape (a benign stop surfaced as a thrown `VELOX_USER_FAIL`), so perhaps both benign stops want to be a result rather than two exception shapes.

Let's settle the shape before iterating on the implementation — what do you think?

singcha:
Thanks @mbasmanova for review! We can finalize the shape first.

Two callers run synchronously and cancel from a different thread than the one running the query and both go through `SqlQueryRunner::run`.
* The interactive `CLI: SqlQueryRunner::run` is blocked in `blockingWait`, and `Ctrl+C` (on a signal-handler thread) trips a `CancellationSource`. The running thread can't cancel itself — it's blocked — so the cancel is inherently cross-thread.
* A user running a query through the Python bindings from a notebook/UI: the query runs on a worker thread with the GIL released, and "stop" on the running cell calls a cancel entry point from another thread that trips the CancellationSource.

**Sync vs async** - you're right, We should change the existing callers. A caller can drive `execute()` under its own `co_withCancellation` scope and get cancellation from structured concurrency; being synchronous at the boundary doesn't stop it — the coroutine is driven under `blockingWait` in C++ either way. And since both callers funnel through the single `SqlQueryRunner::run`, there's no duplication to justify a token on `Runner::drain`.

So, this runner-layer change likely isn't needed. `execute()` already composes scope cancellation and already reaps via `co_close()`, so both the deadline and the external cancel can be composed at the `SqlQueryRunner` layer via `co_withCancellation` over `execute()` — leaving the Runner surface unchanged.

**Signaling**  - We can make the benign stops a returned outcome — `{completed, timedOut, cancelled}` — with genuine execution errors still propagating as themselves. `kQueryCancelledErrorCode` goes away.

singcha:
End-to-end picture — both callers share the same shape

image

The `CancellationSource`/token is the shared handle: Thread B holds the source and trips it; Thread A composed that token into `execute()'s` awaiting scope.

singcha:
Thanks @mbasmanova for offline discussion.
Updated the PR to add co_drain which returns drain outcome — {completed, timedOut, cancelled} — with genuine execution errors still propagating as themselves.

mbasmanova (changes requested):
Thanks for iterating — the outcome enum is cleaner than the string error code. Stepping back on the mechanics, though, I think the shape is still off.

**The disambiguation shouldn't exist — it's a symptom, not a feature.** All of `outerToken` / `scopeToken` / `scopeCancelled` / `state()==kError` exists to reconstruct *why* the run stopped, because that information was thrown away upstream: `cancelTasks()` records cancellation as `error_ = VELOX_FAIL("Query cancelled")` (`LocalRunner.cpp:494`) — a `VeloxRuntimeError` shaped exactly like a genuine failure — which then flows back through the cursor. Fix the signal at the source and the whole layer (and its error-swallowing race — a genuine error co-occurring with a cancel currently resolves to `kTimedOut`/`kCancelled`) disappears.

**Surface cancellation the conventional way.** In folly::coro, cooperative cancellation surfaces as `folly::OperationCancelled`, and `folly::coro::timeout` surfaces `folly::FutureTimeout`. So:
- `cancelTasks()` on the pure-cancel path: set `kCancelled` and cancel the tasks, but don't populate `error_`. Reserve `error_` for genuine failures (the `onError` path already records its error first, so the `if (!error_)` guard leaves it unaffected).
- The read/pull path: when it stops due to cancellation, surface `folly::OperationCancelled` rather than rethrowing a fabricated `VeloxRuntimeError`.

Then the three outcomes are distinguishable by type for free — genuine error propagates as itself, external cancel as `OperationCancelled`, deadline as `FutureTimeout` — with no `scopeCancelled` atomic and no `state()` tie-breaking.

**Which removes the need for `co_drain()` / `DrainOutcome`.** With cancellation typed, the composition is plain structured concurrency the caller already owns: drive `execute()` under `co_withCancellation(token, timeout(..., deadline))`, `co_close()` (shielded) to reap, and `catch` the two benign types. Both callers funnel through `SqlQueryRunner::run`, so that composition lives there once and the `Runner` virtual surface stays unchanged — where we landed last round.

**Please extract these two into a separate PR — they're unrelated to the async surface:**
- the serial `co_pull()` re-block fix, and
- the best-effort reap-stats snapshot.

singcha:
Thanks @mbasmanova!

- Fixed the signal at the source instead of reconstructing it downstream. cancelTasks() no longer fabricates a VeloxRuntimeError("Cancelled"); it records kCancelled and requests task cancellation, reserving error_ for genuine failures. That also removes the error-masking race — a real error co-occurring with a cancel now wins via onError()'s existing `if (error_)` guard.
- Cancellation is now typed: an external cancel surfaces as folly::OperationCancelled and the deadline as folly::FutureTimeout, so the outerToken / scopeToken / scopeCancelled / state()==kError disambiguation is gone entirely.
- co_drain()/DrainOutcome removed. SqlQueryRunner::run composes it directly — co_withCancellation(token, folly::coro::timeout(execute-loop, deadline)) plus a shielded co_close(), catching the two benign types. Both callers funnel through run(), so the composition lives there once and the Runner virtual surface is unchanged.
- Opened a separate Velox change for the root cause : https://github.com/facebookincubator/velox/pull/18222 Velox's Task::terminate() fabricates a VeloxRuntimeError("Cancelled") for a requested cancel (kCanceled) and TaskCursor::moveNext() rethrows it, so a cancelled task was indistinguishable from a genuine failure at the read boundary — and no Axiom-side guard could be race-free, since the throw happens inside moveNext(). The Velox change surfaces a requested cancel as folly::OperationCancelled from both cursor paths, which is what lets the whole disambiguation layer disappear.
- Split the two unrelated LocalRunner fixes (serial co_pull() re-block and the best-effort reap-stats snapshot) into their own PR - https://github.com/facebookincubator/axiom/pull/1625

xiaoxmeng (approved):

PR #1825, opened 2026-09-04
fix(cli): Open a session without a default schema

PingLiuPing:
```
axiom_sql --etc_dir <dir> --catalog <name>` aborted before the prompt:

    Reason: Schema must be specified for connector iceberg
```

Only three connector names carry a built-in default schema, a catalog with no default could not be opened at all. The session already sets a schema at any point with `use <catalog>.<schema>`, and a qualified table name needs none.
With this patch, the missing default now leaves the schema empty and prints what to run instead of refusing to start.
And a session now opens with an empty default schema. Only bare table names need one, so tables are reached either by qualifying them or by setting a schema for the session:

```
    SELECT count(*) FROM metadata.functions
    USE system.metadata
```

The REPL prints a one-line reminder of that after its greeting. Scripted runs stay quiet.

Two changes carry it:

- Connector setup and a new catalog-exists check move inside the `try` that prints `Error: ...`, so a startup problem exits 1 instead of aborting.
- `Console::run()` return status code, so a failed statement reaches `main`'s exit code rather than being swallowed.

These two changes the exit codes:

**Startup, before any session open**

| Condition | Before | After |
|---|---|---|
| `--data_path` + `--etc_dir` | **134 (SIGABRT)** | 1, `Error: … mutually exclusive` |
| `--catalog nosuch` | **134 (SIGABRT)** | 1, `Error: Catalog does not exist: nosuch` |
| `--catalog system` (no default schema) | **134 (SIGABRT)** | starts normally, 0 scripted, prompt + hint interactive |
| `--repeat 0`, `--repeat` with no SQL, unreadable `--init` | 1 | 1 |

**Non-interactive — `--query`, or piped stdin**

| Condition | Before | After |
|---|---|---|
| a statement fails | 0 | **1** |
| multi-statement, middle fails (rest skipped either way) | 0 | **1** |
| `--repeat N`, a run fails | 0 | **1** |
| Ctrl+C during the query | 0 | **130** |
| all statements succeed | 0 | 0 |

**Interactive REPL — nothing changed**

| Condition | Before | After |
|---|---|---|
| a statement typed at the prompt fails | 0 | 0 |
| Ctrl+C during a query (session continues) | 0 | 0 |
| `.exit` / `.quit` / Ctrl+D | 0 | 0 |

**`--init` — runs in both modes**

| Condition | Before | After |
|---|---|---|
| `--init` statement fails, then `--query` | 0, **`--query` still ran** | **1**, `--query` skipped |
| `--init` statement fails, interactive | 0, prompt opened with no warning | 0, prompt opens after an `--init did not finish` notice |
| Ctrl+C during `--init`, then `--query` | 0, **`--query` still run** | **130**, `--query` skipped |

### DIFF
--- a/axiom/cli/AxiomSql.cpp
+++ b/axiom/cli/AxiomSql.cpp
@@ -26,6 +26,7 @@
 #include "axiom/cli/Connectors.h"
 #include "axiom/cli/Console.h"
 #include "axiom/cli/SystemUser.h"
+#include "axiom/connectors/ConnectorMetadataRegistry.h"
 #include "axiom/connectors/tests/TestTableJson.h"
 #include "velox/common/base/Exceptions.h"
 
@@ -51,7 +52,7 @@ int main(int argc, char** argv) {
   folly::FunctionScheduler progressScheduler;
   axiom::sql::SqlQueryRunner runner{
       axiom::sql::SystemUser::resolve(), &progressScheduler, FLAGS_v2};
-  runner.initialize([&]() {
+  auto initializeConnectors = [&]() {
     VELOX_USER_CHECK(
         FLAGS_data_path.empty() || FLAGS_etc_dir.empty(),
         "--data_path and --etc_dir are mutually exclusive. Use --data_path for "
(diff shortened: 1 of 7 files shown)


mbasmanova (changes requested):
Thank you for the fix.

**`--query` reports success when a statement fails.** `Console::runOnce` catches the error, prints `Query failed: ...` and returns false; `runMultiple` stops; `run()` is void, so `main` falls through to `return 0`. This predates the change and holds for every catalog — `--query "select * from nonexistent"` against tpch exits 0 today — but the change routes one more failure into it. A bare table name against a schemaless catalog used to abort during `runner.initialize()` with a non-zero exit; now it is an ordinary failed statement that a script cannot detect. `runOnce` already returns the bool; passing it up through `runMultiple` and `run()` would let `main` exit non-zero.

**The hint prints in non-interactive runs.** `Console::run` gates its greeting on `interactive` and gates live progress on `isatty(STDERR_FILENO)`, with the comment that the non-interactive paths opt in via `--show_live_progress` — the convention being that `--query` and piped stdin emit query output and nothing else. The new message is printed from `main` before any of that is known, so every scripted run against a schemaless catalog carries it. Could it move to where the mode is known and follow the greeting?

mbasmanova (approved):
Thank you — the exit-code table in the description and the scrut cases make this easy to check.

One behaviour I want to make sure is intended: a failed `--init` statement now returns before the interactive branch, so the prompt never appears. For a scripted run that is clearly right. For an interactive one the user has already seen `Query failed: ...`, and `--init` can be expensive — generating data, creating tables — so exiting costs them the whole setup rather than dropping them at a prompt where they can look around and fix it. Would reporting the failure and still opening the REPL be better there?

singcha (approved):

PR #1826, opened 2026-09-04
feat(cli): Show result tables through a pager

PingLiuPing:
With axiom_sql, a result table wider than the terminal wrapped each row across several lines, which made a wide SELECT unreadable.

With this patch, the result tables written to a terminal now go through `--pager`, `less -SFX` by default, so a long row stays on one logical line and scrolls sideways. Redirected output is untouched: a pipe or a file still gets plain text, and `--pager=` turns the pager off.

`printResults` writes to a stream the caller supplies rather than to `std::cout`. The console formats into a buffer and hands it to the pager, then prints that same buffer to stdout when the pager will not start or exits with a failure.


singcha:
Thanks for changes!

- fbcode/axiom/cli/Console.cpp:120 — Give it internal linkage. A non-static free function at file scope leaks an external symbol; per API-design rules free helpers in .cpp belong in an anonymous namespace. Wrap in namespace { ... }.
- fbcode/axiom/cli/Console.cpp:131 — Handle short writes. A single fwrite to a pipe can return short without error (and SIGPIPE/partial consumer close); loop until all bytes are written or an error occurs, otherwise a wide table can silently truncate in the pager.
- fbcode/axiom/cli/Console.cpp:342 — Narrow the fallback. Currently a pager that renders then exits nonzero causes the full table to print to stdout a second time. Fall back to stdout only when the pager never starts or accepts zero bytes; treat nonzero exit after a full write as success.
- fbcode/axiom/cli/Console.cpp:126 — [nit] Rename to pagerPipe (or pagerStream). pager reads as the command string (FLAGS_pager), not the FILE*.
- fbcode/axiom/cli/ResultPrinter.h:32 — Add a unit test pinning the ostream overload (e.g. small table + maxRows truncation + footer into an ostringstream). Behavior change with no test coverage; .sql harness does not apply to CLI formatting, so a C++ expectation is the acceptance test.
- fbcode/axiom/cli/Console.cpp:120 — [nit] Take std::string_view instead of const std::string&. Team style prefers string_view for function parameters, and the body only needs data() + size() for fwrite.

PingLiuPing:
@singcha  Thanks for the review.

> fbcode/axiom/cli/Console.cpp:120 — Give it internal linkage. A non-static free function at file scope leaks an external symbol; per API-design rules free helpers in .cpp belong in an anonymous namespace. Wrap in namespace { ... }.

Thanks, double checked, the code already in anonymous namespace. 

> fbcode/axiom/cli/Console.cpp:131 — Handle short writes. A single fwrite to a pipe can return short without error (and SIGPIPE/partial consumer close); loop until all bytes are written or an error occurs, otherwise a wide table can silently truncate in the pager.
> fbcode/axiom/cli/Console.cpp:342 — Narrow the fallback. Currently a pager that renders then exits nonzero causes the full table to print to stdout a second time. Fall back to stdout only when the pager never starts or accepts zero bytes; treat nonzero exit after a full write as success.
> fbcode/axiom/cli/Console.cpp:126 — [nit] Rename to pagerPipe (or pagerStream). pager reads as the command string (FLAGS_pager), not the FILE*.

Thanks. Refactored the code by adding a new helper class Pager. And now the results will not be printed twice.

> fbcode/axiom/cli/ResultPrinter.h:32 — Add a unit test pinning the ostream overload (e.g. small table + maxRows truncation + footer into an ostringstream). Behavior change with no test coverage; .sql harness does not apply to CLI formatting, so a C++ expectation is the acceptance test.

Thanks, added new test file ResultPrinterTest.cpp.

> fbcode/axiom/cli/Console.cpp:120 — [nit] Take std::string_view instead of const std::string&. Team style prefers string_view for function parameters, and the body only needs data() + size() for fwrite.

Fixed.

PR #1827, opened 2026-09-04
fix(parser): Reject a table version instead of ignoring it

PingLiuPing:
`SELECT ... FROM t FOR VERSION AS OF 8` parsed and then read the current snapshot. The grammar accepts the `FOR ... AS OF`/`BEFORE` clause and the AST carries a slot for it, but `visitTableName` built the table without ever reading `tableVersionExpression`, so the version was dropped and the query silently answered from the wrong snapshot.

The parser now rejects it with "Table version (time travel) is not supported yet". Every form the grammar accepts is rejected the same way, all four version keywords, in both the `AS OF `and `BEFORE` forms. So a versioned reference cannot silently read the current snapshot.

### DIFF
--- a/axiom/sql/presto/ast/AstBuilder.cpp
+++ b/axiom/sql/presto/ast/AstBuilder.cpp
@@ -544,6 +544,17 @@ std::any AstBuilder::visitAliasedRelation(
 std::any AstBuilder::visitTableName(PrestoSqlParser::TableNameContext* ctx) {
   trace("visitTableName");
 
+  // The grammar accepts `FOR ... AS OF`/`BEFORE`, but no layer below the parser
+  // reads a chosen snapshot yet. Reject it rather than build a Table that drops
+  // the clause, which would read the current snapshot as if no version were
+  // given.
+  if (ctx->tableVersionExpression() != nullptr) {
+    AXIOM_PRESTO_SYNTAX_FAIL(
+        getLocation(ctx->tableVersionExpression()),
+        ctx->tableVersionExpression()->getText(),
+        "Table version (time travel) is not supported yet");
+  }
+
   return std::static_pointer_cast<Relation>(std::make_shared<Table>(
       getLocation(ctx), getQualifiedName(ctx->qualifiedName())));
 }
--- a/axiom/sql/presto/tests/PrestoParserTest.cpp
+++ b/axiom/sql/presto/tests/PrestoParserTest.cpp
@@ -1177,6 +1177,24 @@ TEST_F(PrestoParserTest, tablesample) {
   }
 }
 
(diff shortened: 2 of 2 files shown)


mbasmanova (changes requested):
Thank you for picking this up — a version that parses and is then silently dropped is the worst of the options.

**Could this land as parse-and-reject only?** No connector overrides `supportsTableTimeTravel()`, so `FOR VERSION AS OF 8` never reaches `PlanBuilder::tableScan`, in a test or in a deployment. The snapshot on `SchemaTableName`, its place in hashing and identity, the serde field, the `PlanBuilder` parameter and the connector hook are all unexercised.

Unexercised plumbing does not stay right while it waits. The `findView` call that follows `findTable` in `RelationPlanner::processTable` gets the snapshot-bearing name, and `findView` is a `views_.find(tableName)`, so the first connector to opt in would see "table not found" for a versioned view rather than a message about views.

The rejection is the valuable half of this change and stands on its own. When a connector is ready to honor a snapshot, the shape can be settled with a real caller in hand — and Presto's is worth a look first: it keeps `SchemaTableName` a name and passes `ConnectorTableVersion` beside it into `getTableHandle`, with the resolved version living in the table handle. That type also carries a version *operator* for `AS OF` versus `BEFORE` and a typed value rather than an integer, since `FOR TIMESTAMP AS OF` is a timestamp — so an `optional<int64_t>` on the name would have to be replaced rather than extended once either rejected form is implemented.

Also, `rejectsUnsupportedTableVersionForms` names the expectation rather than what it covers. `tableVersion` reads like its neighbours, `tablesample` and `everything`.

mbasmanova (changes requested):
Thank you — and thank you for saying where the plumbing came from. Writing the parser end against a connector you can see makes sense; it just could not be reviewed here without one.

The description still describes the change that was dropped: snapshot id travelling end to end, `SchemaTableName` carrying it as part of table identity, `PlanBuilder::tableScan` accepting it, connectors opting in through `supportsTableTimeTravel()`. None of that is in the diff now, and the description becomes the commit message. Could you rewrite it to what this does — reject every table version form the grammar accepts, so a versioned reference cannot silently read the current snapshot?

`tableVersion` repeats the same call four times with one expected message, and each statement repeats `SELECT * FROM nation FOR`. Looping over just the version clauses would leave only the part that varies.

mbasmanova (approved):
Thank you.

The comment above `tableVersion` repeats the one in `visitTableName` almost word for word. The reasoning belongs at the code; here the test name and the expected message already say what it covers, and two copies will drift.

Also a typo for the commit message: "in both the `AS OF `and `BEFORE` forms" has the space inside the backticks.

PR #1882, opened 2026-09-11
fix(cli): Default to one worker and one driver

natashasehgal:
Summary: The CLI planned every query across four in-process workers with four drivers each, so a query typed at the prompt came back with a distributed plan. Update to default of one worker and one driver. Both flags can be overriden to take other values.

Differential Revision: D119739622


amitkdutta (approved):

PingLiuPing:
@natashasehgal FYI, default to 1 causes huge perf regression on both v1 and v2.

